PR #7, opened 2014-12-09
add tagRegexp option.

succhiello:
add tagRegexp option for using different template tags.


ampedandwired:
Thanks. I'd prefer to keep the tmpl options and the options for this plugin separate though. Perhaps we could add support for passing a reference to an existing tmpl object into the plugin. That way we wouldn't have to add options for every blueimp template feature. What do you think?

mscienski:
Yes please to passing in an existing tmpl object. I ran into a need for this too due to working with Django templates which use the same regex syntax.

PR #8, opened 2015-01-12
Recompile if file has changed

### LINKED ISSUE #5: Template change doesn't trigger recompilation in watch mode
Hi. Thanks for the great plugin.
There is one minor issue though I'm not sure is it the right place to address it.

When I run webpack-dev-server and change some javascript file in dependency graph, webpack notices it and recompiles the bundle. But it doesn't work for html-webpack-plugin templates. Could this be fixed? Thanks.

kurtharriger:
Instruct webpack to watch and recompile when template file has changed.

Fixes #5

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ HtmlWebpackPlugin.prototype.apply = function(compiler) {
       if (!templateFile) {
         templateFile = path.join(__dirname, 'default_index.html');
       }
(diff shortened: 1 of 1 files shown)


ampedandwired:
Cool thanks. I'll have a look at this tonight.

ampedandwired:
Works a treat, thanks.

PR #14, opened 2015-03-11
Support for .css and .manifest files and cache busting

jantimon:
This pull request deprecates the usage of  `{%=o.htmlWebpackPlugin.assets%}` and creates a new representation called `{%=o.htmlWebpackPlugin.files%}`.

Maybe you can come up with a better name than `files`.
I had to rename  `assets` to `files` to support existing (< 1.1.0) templates.

Example structure of the  `{%=o.htmlWebpackPlugin.files%}` object:

```js
{
  "manifest" : "manifest.appcache",
  "css" : [
    "app.css"
  ],
  "js" : [
    "commons.js",
    "index_bundle.js"
  ],
  "chunks" : {
    "commons" : {
      "entry" : "commons.js",
      "css" : [
      ]
    },
    "app" : {
      "entry" : "index_bundle.js",
      "css" : [
         "app.css"
      ]
    }
  }
}
```

It works well together with the https://github.com/lettertwo/appcache-webpack-plugin and https://github.com/webpack/extract-text-webpack-plugin

Added support for cache busting by using the webpack compilation hash.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,93 @@
+{
+    // From JSHint Default Configuration File
+    // See http://jshint.com/docs/ for more details
+
+    "maxerr"        : 50,       // {int} Maximum error before stopping
+
+    // Enforcing
+    "bitwise"       : true,     // true: Prohibit bitwise operators (&, |, ^, etc.)
+    "camelcase"     : false,    // true: Identifiers must be in camelCase
+    "curly"         : true,     // true: Require {} for every new block or scope
+    "eqeqeq"        : true,     // true: Require triple equals (===) for comparison
+    "forin"         : false,    // true: Require filtering for..in loops with obj.hasOwnProperty()
+    "freeze"        : true,     // true: prohibits overwriting prototypes of native objects such as Array, Date etc.
+    "immed"         : false,    // true: Require immediate invocations to be wrapped in parens e.g. `(function () { } ());`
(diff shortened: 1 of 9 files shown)


jantimon:
If you think we could also just jump to 2.0 we could keep the `assets` name and drop the legacy support.
We could also create a 2.0 branch.

jantimon:
@ampedandwired do you think this might be the way to go?
If so I would like to prepare some examples.

jhnns:
That one looks really promising! (Haven't looked at the code though :grinning:)

ampedandwired:
I really like this approach. Not so sure about the name `files` but I can't think of anything better.

Would be good to have a spec covering css (extract text plugin) and manifest (appcache) scenarios if you have the time. If not I can add them post merge, I've done some manual testing on this and it all looks good.

Should the manifest have the publicPath prepended?

Should the insertion of the cache-busting query hash be optional? Generally I would have expected cache-busting to be done by including the hash in the output filename using webpack's `[name]_[hash].js` output convention. Not sure if it's such an issue these days, but apparently the derfault configuration of older versions of Squid actually didn't cache URLs with query params at all.

jantimon:
I'll add some tests for integration with both plugins.

[Squid caches resources with a query string.](http://bizcoder.com/caching-resources-with-query-strings)
But I guess you are right the cache buster should be an option which is off by default.

jantimon:
The cache buster is now optional.
Added jshint configuration.

PR #15, opened 2015-03-13
Feature/append

### LINKED ISSUE #12: Automatically add JS/CSS to HTML files for CDN resources
Tobias seems to think this feature request might be a good fit for your library: https://github.com/webpack/webpack/issues/754#issuecomment-72536024

I have the following concerns:
1. I want to provide an existing HTML file.
2. Ideally, I want to work with well-formed HTML files (not containing any templates). I expect the plugin to append `<script>` and `<link>` tags to the end of `<head>`.

What do you think? Are you willing to add this to this plugin?

jantimon:
This pull requests is build on top of #14 and fixes #12.
Unfortunately it is hard to review as it shows all changes from (#14) as well.
[This commit contains only the changes of this PR](https://github.com/jantimon/html-webpack-plugin/commit/fd47eda875a643a78c362b27cae5f1065fb18a94)

The new code allows to pass a plain html file like [this one](https://github.com/ampedandwired/html-webpack-plugin/blob/fd47eda875a643a78c362b27cae5f1065fb18a94/spec/fixtures/plain.html) and injects all chunks or only the given chunks.

Usage:

```js
new HtmlWebpackPlugin({
       append: true,
       template: path.join(__dirname, 'fixtures/test.html')
     })
```

If append is not set or `false` the HtmlWebpackPlugin will work just like before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ plugins: [
 ]
 ```
 
-The `templateContent` option can also be a function:
+The `templateContent` option can also be a function to use another template language like jade:
 ```javascript
 plugins: [
   new HtmlWebpackPlugin({
@@ -206,3 +206,28 @@ template is rendered. This variable has the following attributes:
 
 - `webpackConfig`: the webpack configuration that was used for this compilation. This
   can be used, for example, to get the `publicPath` (`webpackConfig.output.publicPath`).
+
+
+Filtering chunks
+----------------
+
+To include only certain chunks you can limit the chunks beeing used:
+
+```javascript
+plugins: [
+  new HtmlWebpackPlugin({
+    chunks: ['app']
+  })
+]
+```
+
+It is also possible to exclude certain chunks by setting the `excludeChunks` option:
+
+```javascript
+plugins: [
+  new HtmlWebpackPlugin({
+    excludeChunks: ['dev-helper']
+  })
+]
+```
+
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ template is rendered. This variable has the following attributes:
 Filtering chunks
 ----------------
 
-To include only certain chunks you can limit the chunks beeing used:
+To include only certain chunks you can limit the chunks being used:
 
 ```javascript
 plugins: [
--- a/index.js
+++ b/index.js
@@ -149,11 +149,11 @@ HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function(html, templateParams
   });
   // Turn script files into script tags
   scripts = scripts.map(function(scriptPath) {
-    return '<script src="' + scriptPath + '?' + templateParams.hash + '"></script>';
+    return '<script src="' + scriptPath + templateParams.htmlWebpackPlugin.querystring + '"></script>';
   });
   // Turn css files into link tags
   styles = styles.map(function(stylePath) {
-    return '<link href="' + stylePath + '?' + templateParams.hash + '" rel="stylesheet">';
+    return '<link href="' + stylePath + templateParams.htmlWebpackPlugin.querystring + '" rel="stylesheet">';
   });
   // Append scripts to body element
   html = html.replace(/(<\/body>)/i, function (match) {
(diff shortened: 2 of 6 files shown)


ampedandwired:
This is really cool, I like it, and we should probably encourage this approach as the default. Although templates still useful for advanced/odd requirements I think.

I think perhaps `inject` is better wording than `append`? "Append" implies adding to the end of the file.

I noticed that if you specify `append: true` but use the default template you end up with duplicated script and css includes (once by the template and once by the injection). Would be good to convert the default template to use injection. As a side effect you would also gain the benefit of being able to specify the chunks to include when using the default template too, which is useful functionality.

jhnns:
I'm not sure about this feature. The intent is good – but [replacing stuff in HTML without parsing it is very error-prone](http://stackoverflow.com/a/1732454). And I guess there are more cases where you have "odd requirements".

jantimon:
@jhnns well I know about this however we search case insitive for `<html>` and `</body>` - I guess that's an acceptable requirement for an index.html file don't you think?
And we are not turning off the template capabilities - you will still be able to use the template variables just like before.

jantimon:
@ampedandwired I renamed the option to `inject` and added a fix to prevent the double injection

jantimon:
Rebased down to 4 commits

jantimon:
What is needed to merge this request?

ampedandwired:
As @jhnns points out, would be way less error prone to parse the HTML properly. However given that this feature is opt-in and we're only looking for and replacing specific tags I'm OK with it as a first cut. We can clearly specify the parsing requirements in the readme.

One thing that did occur to me though. I like the idea of being able to limit the chunks that are injected into the HTML. I think that this functionality should probably also be available when you're using the template approach as well though

So the API would be something like this:

```javascript
plugins: [new HtmlWebpackPlugin({
  inject: true,
  chunks: ['util', 'app'],
  templateContent: fs.readFileSync(path.join(__dirname, 'fixtures/plain.html'), 'utf8')
})]
```

The "chunks" attribute could have the effect of scoping the contents of the `files` param that's available for the templates.

What do you think?

ampedandwired:
Also, I made the decision to include the query hash directly in the URL within the `files` object, rather than append it within the template. It seemed a little cleaner and consistent with the "publicPath" behaviour. It does, however, mean a small update to this PR, as the queryString no longer needs to be explicitly appended.

jantimon:
I really don't agree with @jhnns as well we don't parse html like in the link he provided.
Instead we are searching for positions inside a html document.
I believe that using a parser is slower and will cause more problems for two reasons.
First it is likely to restructures the document and requires a valid index.html file to work.

jantimon:
I'll try to change the default querystring behaviour tonight.  
Allowing to pick which chunks should be injected and which are still available in the template looks very clean. We should definitely do that.

jantimon:
Added `chunks` and `excludeChunks` to the api.

README.md
```diff
@@ -206,3 +206,28 @@ template is rendered. This variable has the following attributes:
 
 - `webpackConfig`: the webpack configuration that was used for this compilation. This
   can be used, for example, to get the `publicPath` (`webpackConfig.output.publicPath`).
+
+
+Filtering chunks
+----------------
+
+To include only certain chunks you can limit the chunks beeing used:
```
ampedandwired: Typo "beeing"

index.js
```diff
@@ -80,8 +84,6 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function(compilation, webp
     // Will contain the html5 appcache manifest files if it exists
     manifest: Object.keys(compilation.assets).filter(function(assetFile){
       return path.extname(assetFile) === '.appcache';
```
ampedandwired: Should the manifest have the hash appended as well if `options.hash` is set?

ampedandwired:
Nice work, looks good!

jantimon:
Okay - ready for merge and publish.
I have already increased the version number to 1.3.0

ampedandwired:
Released 1.3. Thanks again for your awesome work Jan.

jantimon:
Cool :+1:

PR #16, opened 2015-03-13
Allow to pass a function for custom html rendering / templating

jantimon:
This pull request allows to pass a synchronous function to generate the templateContent.

Might also fix #7


ampedandwired:
This only fixes #7 insofar as you can do all the templating yourself using your own template engine instance (could be anything you like, not just blueimp). I'd still like a better fix for #7, but this is useful functionality in its own right.

PR #26, opened 2015-04-18
Fixed gathering CSS chunks with content hash in their names

nartamonov:
This change allows to properly populate default_index with links to css files which contains content hash in their names. For example, when we create separated stylesheets with ExtractTextPlugin:

new ExtractTextPlugin("[name].css?[contenthash]");

Currently html-webpack-plugin will not include such files in generated index.html.

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,7 +99,9 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function(compilation, webp
 
     // Gather all css files
     var css = chunkFiles.filter(function(chunkFile){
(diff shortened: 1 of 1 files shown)


nartamonov:
Such configuration with content hashes in chunk names is used, fro example, by [react-starter](https://github.com/webpack/react-starter/blob/master/make-webpack-config.js#L125)

jantimon:
Cool :+1:  could you please add a test?

ampedandwired:
Should be fixed in 1.3, let me know if you see any problems still.

Issue #28, opened 2015-04-23
Favicon 


cgarvis:
Any suggestions on adding a favicon?

jantimon:
You could use a custom template or templateContent

SimenB:
@jantimon Will Webpack pick up the usage and include the image in it's build?

jantimon:
No unfortunately it won't as webpack gathers only files from the loaders.

SimenB:
@jantimon I just got it working locally, I'll throw together a PR later today

jantimon:
Cool

tomatau:
How does this work? Trying to add a favicon to my webpack dist. Currently trying this:

```
  entry:
    "main" : [
      "#{SRC}/favicon.ico"
    ]
  module:
    loaders: [{
      test: /\.ico$/,
      loader: ExtractTextPlugin.extract('file')
    }
  plugins: [
    new ExtractTextPlugin("favicon.ico",
      allChunks: false
    )
 ]
```
More recent browsers don't need the favicon file specified in the html so nothing needed by HtmlWebpackPlugin.

This gets the favicon and puts it into the correct output but the file isn't valid... 
This is probably outside the remit of your plugin but I can't find any other discussions about this! Any help appreciated!

SimenB:
@tomatau You could try it's own loader for `favicon` (just `file`), then set the name, so that webpack doesn't just give it a MD5 name

jantimon:
The solution for 1.4.0 allows to add favicon (however it hasn't been merged into the master branch yet):

```javascript
new HtmlWebpackPlugin({
  favicon: 'my-favicon.ico'
})
```

Specifying a favicon name allows you to have different icons for different environment. (e.g. development, stage, prod)

tomatau:
does this favicon file get picked up without being added as an entry or required by a webpack bundled codebase elsewhere?

jantimon:
Yes it is - however the name isn't getting hased by now - so if anyone could add this I would love to merge it.

jantimon:
Should we add https://github.com/haydenbleasel/favicons ?

Issue #10, opened 2015-01-24
Allow using a loader


Couto:
I'm not sure if this is a useful idea for other people, but it happens that my index.html file already contains some images. It would be nice to add the images to webpack's dependency graph.

Is it possible to use the [html-loader](https://github.com/webpack/html-loader) in conjunction with this plugin? Or is this idea too farfetched? 

Once again, thanks.

ampedandwired:
This is a pretty interesting idea - I like it. To be honest I'm not too sure if it's possible, but I'll definitely take a look into it.

The webpack stats object actually contains the compiled code for each asset as a string, so conceivably we could take the output of the html-loader and use it as an input template into this plugin.

andersekdahl:
I'd really like this as well. I want to add an `apple-touch-icon` link icon in index.html, and I currently have to require that in JS (even though JS doesn't use it) to force it into the pipeline, and then iterate over all modules in my html template to get the correct path for that file. I'd much rather be able to just do `{% o.require('./myimage.png') %}` in my template.

SimenB:
@andersekdahl Could you post your code? I need a `favicon`, and I'm thinking this is the best way.
Ref #28

jantimon:
So #41 allows to use loaders - could someone please give me some feedback on it?

jantimon:
You can use the beta branch: `npm install html-webpack-plugin@2 --save-dev`

PR #32, opened 2015-05-08
Add support for favicon

SimenB:
Ref #28 

This is supposed to be a temporary solution until #10 is implemented.

A file named `favicon.ico` has to be included manually in some `require`, which I suppose is bad, but until #10 is done, I don't see any way around it


index.js
```diff
@@ -148,6 +148,14 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function(compilation, webp
   // requires the same css.
   assets.css = _.uniq(assets.css);
 
+  if (favicon) {
+    assets.favicon = _.find(webpackStatsJson.modules, function(module) {
+      return /^favicon\.ico($|\?)/.test(path.basename(module.name));
+    });
+    if (assets.favicon) {
+      assets.favicon = webpackStatsJson.publicPath + assets.favicon.assets[0];
```
SimenB: I don't know if webpack has some nice util that will make this less manual?

spec/HtmlWebpackPluginSpec.js
```diff
@@ -486,4 +486,20 @@ describe('HtmlWebpackPlugin', function() {
...
+        path: OUTPUT_DIR,
+        filename: 'index_bundle.js'
+      },
+      module: {
+        loaders: [{ test: /\.ico$/, loader: 'file' }]
+      },
+      plugins: [
+        new HtmlWebpackPlugin({favicon:true})
+      ]
+    }, [/<link rel="shortcut icon" href="0a31c912c8b55c756f7a969982b1ff91.ico">/], null, done);
```
SimenB: I'm really bad at regex, that'd probably be cleaner than adding the has to the test...

jantimon: `[^"]+` will match 1 or more characters that are no double quotes.

jantimon:
Unfortunately this does not work with the new template mode (`inject`: true).

jantimon:
Why do we have to set favicon to true?
Wouldn't it be enough to look for favicons in the assets object?

SimenB:
> Why do we have to set favicon to true?
Wouldn't it be enough to look for favicons in the assets object?

The thought is that it is a potentially heavy operation (my project at work has over 1000 assets). I suppose it's not too much either way, node is pretty fast... What do you think?

spec/HtmlWebpackPluginSpec.js
SimenB: Nice, that works!

jantimon:
We do sth similar with the manifest.
It would be cool if we could keep the same style for the favicon:
https://github.com/SimenB/html-webpack-plugin/blob/favicon/index.js#L86

SimenB:
@jantimon Added support for `inject`. I'll try copying the way manifest is retrieved

jantimon:
Would it be possible to specify a path for favicon in the options and load it like 
`compilation.assets['favicon.ico'] = faviconObject`. (similar to https://github.com/lettertwo/appcache-webpack-plugin/blob/master/src/index.coffee)
If I understood this correct this would allow to add a favicon without requiring the file manually.

jantimon:
It would also be nice if this also allowed to specify files for [`apple-touch-icon`](https://mathiasbynens.be/notes/touch-icons)

SimenB:
@jantimon Unless I want to include _any_ `.ico` file as an icon, I can't use the same way the `manifest` is received, as `compilation.assets` only has the name Webpack gives the file, not the original.

I can see if I can get it working without manually requiring. I can then add the same logic for `apple-touch-icon`.

SimenB:
@jantimon I _think_ that we can't use how `appcache-webpack-plugin` does it, as the `favicon` is an external file, and not something I build up manually.

SimenB:
@jantimon Added support for `apple-touch-icon`. It has to be named `apple-touch-icon.png`. This is the name h5bp has given it.

SimenB:
@jantimon I expanded the PR to also cover any extra file you may want. I can separate it into it's own PR if you'd like, but it builds on this one, so I thought I'd keep them together

jantimon:
yeah I guess the extra file should be in an extra pull request for now.

SimenB:
@jantimon Ok, removed those changes

SimenB:
@jantimon Found a bug if an image was base64-encoded. Should be ready for review now :smile:

index.js
```diff
@@ -70,6 +70,27 @@ HtmlWebpackPlugin.prototype.emitHtml = function(compilation, htmlTemplateContent
...
+        return regexp.test(basename);
+      });
+    })
+    .map(function (module) {
+      // If the assets is not base64-encoded
+      if (module.assets.length) {
+        return [path.parse(module.name).name, publicPath + module.assets[0]];
+      }
+
+      return [path.parse(module.name).name, module.source.substring(18, module.source.length - 1)];
```
SimenB: `module.exports = "` as well as the last `"`

SimenB:
@jantimon Any chance this could get merged? If you're enjoying the weekend, ignore my nagging :smile: 

I'd like to see this in before tomorrow, as I'll need it at work, though.

index.js
```diff
@@ -70,6 +70,27 @@ HtmlWebpackPlugin.prototype.emitHtml = function(compilation, htmlTemplateContent
   };
 };
 
+HtmlWebpackPlugin.prototype.getAssetPathFromModuleName = function(publicPath, modules) {
```
SimenB: It might be better to extract some if these functions into a `lib/` directory or something, as this file is getting a bit bulky, and there's no real need to export anything but the constructor and the `apply`-function

jantimon:
Hi @SimenB thanks for all your work. Unfortunately I am on a conference for the next days but I'll take a look at it.
By now you can simply install it with `npm i --save-dev SimenB/html-webpack-plugin`

SimenB:
@jantimon Ah, OK. The proxy at works kills GitHub dependencies. I'll just publish a fork in the meantime, then deprecate when (if) you merge it :smile:

jantimon:
Here is a first try to solve the favicon (based on your changes)
https://github.com/ampedandwired/html-webpack-plugin/tree/favicon

I removed the apple icon for now as it didn't allow to specify multiple dimensions e.g.
```html
<link rel="apple-touch-icon" sizes="76x76" href="/apple-touch-icon-76x76.png" />
<link rel="apple-touch-icon" sizes="120x120" href="/apple-touch-icon-120x120.png" />
<link rel="apple-touch-icon" sizes="152x152" href="/apple-touch-icon-152x152.png" />
```

Another big change was that it uses a similar way to add the `favicon.ico` like the app manifest plugin.
That means that you don't have to require the favicon.ico in your js code and it won't be part of the resulting output javascript file.

What do you think?

SimenB:
@jantimon Oooh, those changes look much cleaner! Great job!
Could you also expand on the idea I have going in #36, and make it easy to include any arbitrary file, and not just `favicon`? Your `addFileToAssets` should make that pretty easy.
One (possible, but probably out of scope of this plugin) downside is that it won't be passed through a loader. Maybe we can support both formats? Both requiring only for this plugin, but also easily access any asset in the pipeline?

Regarding apple icons, I thought about the different sizes, but as h5bp removed them (https://github.com/h5bp/html5-boilerplate/issues/1367), and now reccomend using just one (https://github.com/h5bp/html5-boilerplate/pull/1622), I thought it was fine to just include the one size.

jantimon:
Although I really like the idea with the extra files I am not sure wether this is just an edge case or a common use case. It is already possible to pass a custom templateContent function which has access to everything needed to use these files.

You are correct the base64 encoding is missing as it doesn't use the loaders.
However I would really prefer not to load the favicon into the javascript.
Maybe you could come up with a good solution for this issue?  

The h5bp is referring only to the file name and is a solid starting point for creating web sites however imho there are still use cases where you would like to provide different icon versions depending on the size.

I rebase this pr into #37 for the new 1.4.0 release but I would like to get the feedback of @ampedandwired before merging.

SimenB:
The us case behind extra files is that I don't want to know how to iterate through the files in my template, or how Webpack inner structure works. I just want this and that file in my template.

I have no idea how to send it through loaders, ref #10. Would be awesome though!

As my use case for this was just favicon, I think it is covered by #37. I'd like to see if we can reach a consensus on #36 though. I'll rebase after `1.4.0` is released :smile:

PR #34, opened 2015-05-09
Use files in package.json

SimenB:
Reduces the number of file a consumer gets when `npm install`ing this package.

Before: http://hastebin.com/kafifaxeqe.sh
After: http://hastebin.com/yejasupoqe.sh


jantimon:
Cool :+1:

PR #35, opened 2015-05-09
Switch out Jasmine with Mocha&Chai

SimenB:
I did this locally to be able to easily debug from IntelliJ. I thought as I already did it, I could provide a PR in case you're interested in doing it yourselves :smile:

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -80,16 +80,5 @@
     "shelljs"       : false,    // ShellJS
     "worker"        : false,    // Web Workers
     "wsh"           : false,    // Windows Scripting Host
-    "yui"           : false,    // Yahoo User Interface
-
-    // Custom Globals
-    "globals"       : {
-      // jasmine helpers
-      "expect": false,
-      "describe": false,
-      "ddescribe": false,
-      "beforeEach": false,
-      "it": false,
-      "iit": false
-    }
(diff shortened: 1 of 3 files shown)


jantimon:
Actually I don't care so I let this decision open to @ampedandwired who chose the testing framework in the first place.

SimenB:
Rebased now, it if matters :laughing:

SpaceK33z:
This makes sense I think, because mocha seems to be standard (webpack also uses it).

MoOx:
My 2 cents https://medium.com/javascript-scene/why-i-use-tape-instead-of-mocha-so-should-you-6aa105d8eaf4

jantimon:
@MoOx don't you think it would be better to use a testing framework for an open source project which is widely known and understood to encourage writing tests?

Tape might be the better tool but people seem to have more experience with Jasmine / Mocha:
http://ashleynolan.co.uk/blog/frontend-tooling-survey-2015-results#q7-javascript-testing

tomatau:
I prefer mocha and chai to tape... the extra assertions improve the semantics as often a deeplyEqual just isn't enough. I've found little setup necessary for mocha also with lots of flexibility -- especially compared to say... jest or karma which both require lots of configuration.

PR #20, opened 2015-04-04
Added support for minification using html-minifier

sebastienbarre:
I discovered ```html-webpack-plugin``` today, great tool, thanks.

This PR adds a ```minify``` option (```true``` or an options ```object```), which enables the minification of all resulting HTML files, using [kangax/html-minifier](https://github.com/kangax/html-minifier). This comes handy when you are working on your own (unminified) template file, but want the smallest final output for performance.

Please note that ```html-minifier``` is very conservative by default, as most of [its options](https://github.com/kangax/html-minifier#options-quick-reference) are set to ```false```.
I suggest using these for example:
```
  var HtmlWebpackPlugin = require('html-webpack-plugin');
  var html_webpack_options = {
    template: 'src/assets/index.template.html'
  };
  if (production) {
    html_webpack_options.minify = {
      removeComments: true,
      collapseWhitespace: true,
      conservativeCollapse: true,
      minifyJS: true,
      minifyCSS: true
    };
  }
  config.plugins.push(new HtmlWebpackPlugin(html_webpack_options));
```
Hope this helps.

PR #37, opened 2015-05-12
1.4.0

jantimon:
This pull request contains #32 and #20.

It adds two new features:
* favicon support (#28 #32)
* html minification (#20)


SimenB:
@jantimon Is it possible to use the `this.minimize` like `html-loader` does? So if I run `webpack -p` from the commandline, it's automatically minified?

https://github.com/webpack/html-loader/blob/master/index.js#L59

sebastienbarre:
@jantimon sorry, didn't get the time to look at it this weekend. Either way I can't replicate what ```html-loader``` is doing, i.e. ```var query = loaderUtils.parseQuery(this.query);```, no such ```this.query``` is passed to ```html-webpack-plugin``` from ```webpack```, that I can see. Though I assume I could always check if ```process.argv.indexOf('-p') > -1```.

SimenB:
Ah ok. Too bad.

How about hooking into `optimize` phase? If I understand this correctly (probably not...) https://github.com/webpack/extract-text-webpack-plugin/blob/master/index.js#L188

sebastienbarre:
@SimenB I'm afraid not, from what I gather. The HTML itself seems to be generated in the ```'emit'``` phase:
https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L13
Sticking a ```compilation.plugin("optimize", function() {``` inside doesn't appear to work, the code is not executed. ```optimize``` is probably a sub-stage that only makes sense during a ```compilation``` stage? Anyway, here is how I handle debug vs. prod in my ```webpack.config.js```
```
var production = process.env.NODE_ENV === 'production' ||
  process.argv.indexOf('-p') > -1;
[...]
  var HtmlWebpackPlugin = require('html-webpack-plugin');
  var html_webpack_options = {
    template: 'src/assets/html/index.template.html'
  };
  if (production) {
    html_webpack_options.minify = {
      removeComments: true,
      collapseWhitespace: true,
      conservativeCollapse: true,
      minifyJS: true,
      minifyCSS: true
    };
  }
  config.plugins.push(new HtmlWebpackPlugin(html_webpack_options));
```

jantimon:
@SimenB the query does something different - if you call a loader like `myloader!myfile?x=1` then `x=1` is your query.
If you call webpack with `-p` it just pushes the `UglifyJsPlugin` into the plugin array: https://github.com/webpack/webpack/blob/master/bin/convert-argv.js#L401

We could check iif the UglifyJsPlugin is part of `compiler.options.plugins`.

SimenB:
@jantimon Anything holding this up?

README.md
```diff
@@ -108,58 +112,64 @@ once in your plugins array:
...
+    }
+  })
+]
+```
+Or the async version:
+```javascript
+plugins: [
+  new HtmlWebpackPlugin({
+    templateContent: function(templateParams, compilation, callback) {
       // Return your template content synchronously here
```
SimenB: missing an `a` :scream:

package.json
```diff
@@ -30,13 +30,16 @@
   "devDependencies": {
     "css-loader": "^0.12.0",
     "extract-text-webpack-plugin": "^0.7.1",
+    "file-loader": "^0.8.1",
     "jasmine-node": "^1.14.5",
     "jshint": "^2.7.0",
     "rimraf": "^2.3.3",
     "style-loader": "^0.12.2",
+    "url-loader": "^0.5.5",
```
SimenB: do you still need this or `file=--loader`?

jantimon: Hm I am not sure why I added them.. You are probably right ;)

SimenB: They are probably left from my commit, as I needed them for tests :smile:

PR #39, opened 2015-05-16
Allow custom compile function to be passed

### LINKED ISSUE #38: Using the "Handlebars" (or similar) instead of "Blueimp".
For example, I want to use "partials" in "index.html" — how can I do this?
Thanks.

bebraw:
Closes #38.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,17 @@ plugins: [
 ]
 ```
 
+Passing custom compiler
+-----------------------
+
+In case you want to use something else than blueimp for templating, you can overwrite the template compiler like this:
+
+```javascript
+plugins: [
+  new HtmlWebpackPlugin({
+    compile: function(template, data) {
+      return Handlebars.compile(template)(data);
+    }
+  })
+]
+```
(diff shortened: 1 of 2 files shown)


bebraw:
Closing... Just noticed `templateContent`.

jantimon:
Yeah - the readme is out of date. I updated the new one on the 1.4 branch however I am waiting for @ampedandwired to merge it and release the new version to npm

bebraw:
Ok, cool.

bebraw:
After some discussion at Gitter we realized something like this might be needed after all! Even though you can process content through `templateContent` hook it will then get processed by blueimp in any case! So if you output content that blueimp recognizes, you are in trouble.

This PR allows you to work around this issue effectively. It's possible the API will have to change somehow if you want to fix this in a nicer way.

jantimon:
we could move the template compilation in an own file however we would have to add all requested template languages to the package json

bebraw:
Yeah, ideally the choice of templating language would remain in user space. Pulling template libs you don't actually use when you install the plugin would be a big no-no in my books.

jantimon:
I have a huge change which allows to use loaders.
This would also fix this issue.

PR #41, opened 2015-05-18
Feature/loaders

jantimon:
This rather big pull request will allow to use loaders #10 :

```js
plugins: [
    new HtmlWebpackPlugin({
      template: 'html!./index.html'
    })
  ]
```

Further changes to get this working:

* It will remove the deprecated `templateParams.htmlWebpackPlugin.assets`.
* It will also remove the `templateContent` feature. (Maybe we can fix this)
* It adds an [examples](https://github.com/ampedandwired/html-webpack-plugin/tree/feature/loaders/examples) - for easier issue debugging and for people to try the plugin
* It sets `inject` to `true` by default

This merge request is work in progress and will need some more testing and cleanup.


SimenB:
Oh, this is awesome!

Regarding `compile`, how about flipping it and calling it `preCompiled`?

jantimon:
I thought something like `postprocess` / `postProcess`. It could be a little bit the replacement for the missing templateContent option.

jantimon:
Open todos: 
+ Move blueimp into a loader (`blueimp!template.html`) - replaces `compile` option
+ More tests
+ Bring back templateContent

index.js
```diff
@@ -1,153 +1,224 @@
...
+    .then(function() {
+      var templateParams = {
+        webpack: webpackStatsJson,
+        webpackConfig: compilation.options,
+        htmlWebpackPlugin: {
+          files: assets,
+          options: self.options,
+        }
+      };
+      if (self.options.compile === true) {
```
bebraw: Why not to allow compiler to be replaced instead of using a flag? That seems like a more flexible alternative to me. It could still default to blue imp for backwards compatibility.

If you want to skip compiling altogether, you would just set `compiler: id` (pass through).

jantimon: hey @bebraw thanks for the feedback.. 
Actually I have a new (hopefully) better idea.
Instead of using a template processing part I will use a blueimp template loader by default.
This will reduce the complexity and will still support blueimp templates

bebraw: Cool. I'm a happy camper if I can write something like `require('html!highlight!markdown!./README.md')` within the template and have it evaluated through Webpack using those loaders.

jantimon: This should already work with this branch however it's not finalised

jantimon:
@bebraw I have just added an example which uses the blueimp loader together with the html loader.
https://github.com/ampedandwired/html-webpack-plugin/tree/feature/loaders/examples/two-loaders

This just an example but should give you an idea how loaders might work together.

SimenB:
This looks really promising, can't wait for it to be released! Anything I can do to speed this up?

A thought, why not drop the blueimp dep, and just go with lodash templates?

jantimon:
@ampedandwired chose blueimp - and I would rather concentrate on the loaders issue first.  
Furthermore you could just use lodash templates with the lodash loader ;)

SimenB:
I use handlebars or jade, so it makes no real difference to me. Just thinking that more people are probably familiar with lodash/underscore templates than blueimp. It also allows you to drop a, albeit small, dependency.

But yeah, different issue :smile:

jantimon:
Before releasing I would like to add more examples and use those examples for integration testing.

So it would be a great support if you could add some more examples (e.g. underscore, jade, handlebars, mocha-loader, using the chunks option and so on)

bebraw:
@jantimon That `{% include(require('html!./partial.html'), {}); %}` looks nice! Great work. :+1:

saberduck:
I have troubles using jade as template, because I have jade registered in loaders array in webpack.config.js  and when I use it for the template it is loaded with blueimp-tmpl. I believe this is because no loaders are detected in the template name? I think, perhaps blueimp-tmpl shouldn't be used by default, only when requested explicitly. Or at least provide option to disable it

jantimon:
@saberduck This is for backwards compatibility.. Maybe we could add the blue-imp loader to the module loaders so it would become a fallback - but I am not sure if this would would work.

saberduck:
Maybe, if you can replace this test 
```
     if(this.options.template.indexOf('!') === -1) {
       this.options.template = 'blueimp-tmpl!' + path.resolve(this.options.template);
     }
```
with some explicit option like `use-blueimp` , which is true by default, but configurable

jantimon:
Why can't you just use jade!template.html ?

saberduck:
because I have jade loader defined as loader for .jade files in module.loaders array. If I specify `jade!` with template, it will try to process file with jade loader twice and it doesn't work. However  'jade!raw!` works, but seems like unnecessary hack.

jantimon:
Why would it load jade twice?

saberduck:
I will try to setup some minimalistic project to show you my problem, maybe I am missing something  and problem is somewhere else, ttyl

jantimon:
Cool - let me know I'll look into it

jantimon:
@saberduck - fixed that. It will use only a loader if you haven't specified one

kbanman:
Excellent work! I'm very much looking forward to being able to use this

saberduck:
Looks great now, thanks! Sorry, I didn't have time for the minimal test-case to reproduce the issue I had before

let4be:
How can I use this to load a jade template? nothing helps me...
`jade!template.jade` doesn't work

let4be:
`template: 'html!jade-html!./src/client/index.jade'` doesn't seem to work either

jantimon:
@let4be The reason is that the jade-loader is not able to load the fs module:
https://github.com/webpack/jade-loader/issues/8

jantimon:
@let4be - I added a jade example. Execute the following commands to try it:
```
git clone -b feature/loaders https://git@github.com/ampedandwired/html-webpack-plugin.git
cd html-webpack-plugin/
npm i
cd examples/jade-loader/
../../node_modules/.bin/webpack
```

aaxx:
Hello, failed to run 'two-templates' example. Got some weird error though the template looks like OK: 
```
ERROR in HtmlWebpackPlugin Error: Error: Child compilation failed:
Module build failed: SyntaxError: Unexpected token ;
    at Function (native)
    at .../node_modules/html-webpack-plugin/node_modules/lodash/index.js:10747:16
    at ...node_modules/html-webpack-plugin/node_modules/lodash/index.js:11020:21
    at ...node_modules/html-webpack-plugin/node_modules/lodash/index.js:8094:31
    at Function.template (...node_modules/html-webpack-plugin/node_modules/lodash/index.js:10746:20)
    at Object.module.exports (.../node_modules/html-webpack-plugin/loader.js:18:20):
SyntaxError: Unexpected token ;
```

aaxx:
Is there any way or proper example how to use the jade or other template engine with partials/include via loaders? It would be really cool to have templates to be resolved from modules, for example

jantimon:
@aaxx fixed the bug in the example please try again ;)

Kagami:
:+1: for this. blueimp is really old, badly supported (doesn't even have "Issues" button) and syntax is weird (IMO).

README.md
```diff
@@ -1,11 +1,11 @@
 HTML Webpack Plugin
-=================== 
+===================
 [![npm version](https://badge.fury.io/js/html-webpack-plugin.svg)](http://badge.fury.io/js/html-webpack-plugin) [![Dependency Status](https://david-dm.org/ampedandwired/html-webpack-plugin.svg)](https://david-dm.org/ampedandwired/html-webpack-plugin) [![bitHound Score](https://www.bithound.io/github/ampedandwired/html-webpack-plugin/badges/score.svg)](https://www.bithound.io/github/ampedandwired/html-webpack-plugin) [![Build status](https://travis-ci.org/ampedandwired/html-webpack-plugin.svg)](https://travis-ci.org/ampedandwired/html-webpack-plugin)
 
 This is a [webpack](http://webpack.github.io/) plugin that simplifies creation of HTML files to serve your
 webpack bundles. This is especially useful for webpack bundles that include
 a hash in the filename which changes every compilation. You can either let the plugin generate an HTML file for you or supply
-your own template (using [blueimp templates](https://github.com/blueimp/JavaScript-Templates)).
+your own template (using lodash/ejs templates.
```
Kagami: Missed closing bracket?

Kagami:
Btw, one more huge :-1: against `blueimp-tmpl`.

I played with relatively large HTML files (~200k) to test inline styles rendering and for some reason my webpack setup gone really slow. After some investigation it turned out that `blueimp-tmpl` **is incredibly slow** (despite that stated in its project description).

Check this simple benchmark:

```
$ npm i blueimp-tmpl
$ for ((i=0;i<30000;i++)) echo "<div>test</div>" >> test.html
$ cat bi.js 
var fs = require("fs");
var tmpl = require("blueimp-tmpl").tmpl;
var html = fs.readFileSync("test.html", {encoding: "utf8"});
console.log(tmpl(html, {a: 1}));
$ time node bi.js | wc -l
30001
node bi.js  23.09s user 0.24s system 99% cpu 23.331 total
wc -l  0.00s user 0.00s system 0% cpu 23.331 total
```

And compare this result for example with handlebars:

```
$ npm i handlebars
$ cat hb.js 
var fs = require("fs");
var Handlebars = require("handlebars");
var html = fs.readFileSync("test.html", {encoding: "utf8"});
console.log(Handlebars.compile(html)({a: 1}));
$ time node hb.js | wc -l
30001
node hb.js  0.00s user 0.00s system 3% cpu 0.113 total
wc -l  0.00s user 0.00s system 0% cpu 0.112 total
```

`blueimp-tmpl` is really doing it wrong.

aaxx:
@jantimon thanks, it helped a lot :+1:

default_index.html
```diff
@@ -1,18 +1,9 @@
...
   <head>
     <meta charset="UTF-8">
-    <title>{%=o.htmlWebpackPlugin.options.title || 'Webpack App'%}</title>
-    {% if (o.htmlWebpackPlugin.files.favicon) { %}
-    <link rel="shortcut icon" href="{%=o.htmlWebpackPlugin.files.favicon%}">
-    {% } %}
-    {% for (var css in o.htmlWebpackPlugin.files.css) { %}
-    <link href="{%=o.htmlWebpackPlugin.files.css[css] %}" rel="stylesheet">
-    {% } %}
+    <title><%= htmlWebpackPlugin.options.title %></title>
```
Kagami: One more request while we are here. It would be nice to have alias for `htmlWebpackPlugin.options`. Now it's really long.

SimenB:
@jantimon Any news here? This is looking fantastic, would love to use it :smile:

jantimon:
@SimenB use it and give feedback ;)

SimenB:
WIll do tomorrow!

SimenB:
@jantimon It works great with handlebars! It nicely deprecates the `favicon` option as well (at least along with `handlebars-loader` which looks up `img` `src`es) :grinning: 
Only thing (not really related to this PR) is it'd be great if you could pick up that minimize is in effect for the whole build, without having to explicitly state it to the plugin.

Any ETA?

index.js
```diff
@@ -1,154 +1,280 @@
...
+      }
+    })
+    // Minify
+    .then(function(html) {
+      if (self.options.minify) {
+        var minify = require('html-minifier').minify;
+        // If `options.minify` is set to true use the default minify options
+        var minifyOptions = _.isObject(self.options.minify) ? self.options.minify : {};
+        try {
+          minify(html, minifyOptions);
```
SimenB: This here destroys `minify` option. Either return the result from `minify`, or assign it to `html`

SimenB: #80

index.js
```diff
@@ -1,154 +1,280 @@
...
+var NodeTemplatePlugin = require('webpack/lib/node/NodeTemplatePlugin');
+var NodeTargetPlugin = require('webpack/lib/node/NodeTargetPlugin');
+var LoaderTargetPlugin = require('webpack/lib/LoaderTargetPlugin');
+var LibraryTemplatePlugin = require('webpack/lib/LibraryTemplatePlugin');
+var SingleEntryPlugin = require('webpack/lib/SingleEntryPlugin');
+
 function HtmlWebpackPlugin(options) {
-  this.options = options || {};
+  // Default options
+  this.options = _.extend({
```
SimenB: You can use `_.defaults` here :smile:

index.js
```diff
@@ -1,154 +1,280 @@
...
+    hash: false,
+    inject: true,
+    compile: true,
+    favicon: false,
+    minify: false,
+    chunks: 'all',
+    excludeChunks: [],
+    title: 'Webpack App'
+  }, options);
+  // If the template doesn't use a loader use the blueimp template loader
```
SimenB: This comment is outdated, isn't it?

index.js
```diff
@@ -1,154 +1,280 @@
...
+  if(this.options.template.indexOf('!') === -1) {
+    this.options.template = require.resolve('./loader.js') + '!' + path.resolve(this.options.template);
+  }
+  // Resolve template path
+  this.options.template = this.options.template.replace(
+    /(\!)([^\/\\][^\!\?]+|[^\/\\!?])($|\?.+$)/,
```
SimenB: Could you comment this? :laughing:

index.js
jantimon: sounds good

index.js
jantimon: yeah should be removed

index.js
jantimon: The regexp? :)

SimenB: Aye, it looks magical

SpaceK33z:
@jantimon, I've tested it with my app, and it works fine!

Error messages don't seem to work (see my comment above), but except this I think this PR makes html-webpack-plugin much better. I think this is more the *webpack* way to do this.

SpaceK33z:
Isn't minifying the html a task of the loader? e.g. `html-loader` has support for this.

jarnix:
I would pay a beer or two for this patch. :beers:

SpaceK33z:
With #83 I fixed the bug I mentioned above (loader errors were not displayed) and fixed the merge conflicts. Is there anything more to do?

jantimon:
thanks a lot - there are still some open topics left which should be adressed so we can release this  2.0 branch:
+ unit tests
+ using it together with the extract-text plugin seems very buggy (might require changes there)
+ performance benchmarks against the current version

ideas which might also be considered:
+ add events so other plugins will be able to add features like live reload, minifications, adding resources like apple icons, dns predetching etc

SpaceK33z:
Okay, I will try to work on some unit tests. We're using your PR in production already in combination with `extract-text-plugin` and so far haven't encountered any issues. What issues are there?

Are benchmarks relevant? This could be a bit slower, but the huge win here is that you're reusing webpack's loaders. Because of this you can use custom template engines, postprocess the html etc.

Why would you add events for other plugins to add such features? Loaders already provide this exact functionality. E.g. `html-loader` can minify the HTML and process images.

index.js
SpaceK33z: What is the purpose of the regexp? I added some extra tests for the loader (will create a PR soon), but the regexp seems to never match anything.

jantimon:
@SpaceK33z events might be useful for the appcache-webpack plugin or for a favicon generator which might add a apple icon, an android icon and so on.
Otherwise we have to include [manifest](https://github.com/lettertwo/appcache-webpack-plugin), [favicon](https://github.com/haydenbleasel/favicons#output), [resource-prefetching](https://en.wikipedia.org/wiki/Link_prefetching#HTML5_prefetching) and so on all in the webpack plugin instead of giving other developers the opportunity to implement this in another plugin and lower the complexity inside the html-webpack-plugin.
Webpack has a event system so all we had to do would think of good event names and trigger them.

Maybe the extract-text-plugin was already fixed but you can try the following: require a css file (which is not required by your javascript files) from your template file and use the extract-text-plugin. 

You are right a benchmark is not that important however it feels like that in dev mode the template is always generated even if nothing changed. This might slow down the compilation especially if you have several html files.

szag:
Wow this feature looks really awesome :)

I found an error while using the plugin together with the `BannerPlugin` like this:
~~~javascript
plugins: [
    new HtmlWebpackPlugin({
        filename: "index.html",
        template: "html!./app/index.tpl.html"
    }),
    new webpack.BannerPlugin("test")
]
~~~

Error:
~~~
ERROR in HtmlWebpackPlugin SyntaxError: evalmachine.<anonymous>:1
 test */
       ^
Invalid regular expression: missing /
~~~

Also the `resolve.root` configuration does not seem to be honored for the `template` path but maybe this is expected behavior, I don't know.

SpaceK33z:
@jantimon, I am not able to reproduce the extract-text-plugin bug.

> require a css file (which is not required by your javascript files) from your template

You mean simply adding a `<link rel="stylesheet" href="not-existing.css">` in the template `<head>`? Or how do you require the css?

I just updated #84 with additional tests :).

jantimon:
@szag - using resolve root is a nice idea. Unfortunately I am not able to reproduce your bug right now.
@SpaceK33z You can require a css file from your underscore template like `<% require('./demo.css'); &>`

szag:
@jantimon Weird. The bug occurred with a bigger config setup. I remember disabling other plugins to see if they interfere but maybe there's another configuration option causing it. I'll check…

jantimon:
I have created a new pre released 2.0.2 and added an installation instruction to the official readme

SpaceK33z:
@szag, I have written a testcase for this and was able to reproduce your issue. I have no idea how to fix it yet unfortunately.

SimenB:
Sweet! :+1:

SpaceK33z:
@szag, I have made a PR #85, which should fix your issue with the bannerplugin. Can you verify this?

szag:
@SpaceK33z Yes the issue is fixed! Thanks for that, awesome :+1:

SpaceK33z:
@jantimon, I have reproduced the issue with `require`-ing a CSS file using a test now. Unfortunately I didn't succeed in squashing this bug. Do you happen to know if the fault is 'ours' or from `extract-text-webpack-plugin`?

That is the latest remaining bug (as far as we know), right? The README needs some updates as well I think. Is it ok if I update the README or do you want to do it yourself?

jantimon:
@SpaceK33z yeah sure feel free to update the readme.

jantimon:
I believe the bug is inside the extract-text-webpack-plugin but the whole plugin contains almost no comments so it's really hard to guess why it was written that way.
Could you please upload your example somewhere and open a ticket in their issue queue?

SpaceK33z:
@jantimon, It should be fixed with #89.

I haven't created a ticket yet on extract-text-webpack-plugin, because I've not yet succeeded at reproducing it without using html-webpack-plugin. I want to rule out as much as possible :).

NekR:
@jantimon
> @SpaceK33z events might be useful for the appcache-webpack plugin or for a favicon generator which might add a apple icon, an android icon and so on.

Works fine with [`offline-plugin`](https://github.com/NekR/offline-plugin)

NekR:
Why `${title}` treats same as `<% title %>` with default template?
Also, is there a way to use templates which return function instead of string?

SpaceK33z:
Do you have an example of a loader that returns a function instead of string and doesn't work? I use this plugin with [underscore-template-loader](https://github.com/emaphp/underscore-template-loader) and it works fine.

NekR:
@SpaceK33z I tried this https://github.com/okonet/ejs-loader which is pretty simple. Will try `underscore-template-loader`

NekR:
> Why `${title}` treats same as `<% title %>` with default template?

Okay, this happens because `lodash` added some weird trick about interpolation with `${ ... }` and problem with returned function was my fail, by some reason loader way applied twice per template.

SimenB:
What;s the current status here? Is it only importing from template missing, or something else? I could try to help out a bit :smile:

SpaceK33z:
@SimenB, the last remaining (known) bug is that `require('test.css')` inside a template fails. I've created a (failing) [test for this here](https://github.com/SpaceK33z/html-webpack-plugin/commit/f2882604e4eeaeaf4f47dd521bb278dee7d8240a). Would be awesome if you can fix that :)! See also [this comment](https://github.com/ampedandwired/html-webpack-plugin/pull/41#issuecomment-141637588).

SimenB:
Are you supposed to be able to require from a template? Can you do that with JS now? Like just chuck a `require('modernizr')` in `head` to avoid repaints?

SpaceK33z:
I personally never use this feature; I just require my css via the entry module.
@jantimon, what's your use case for this?

jantimon:
An isomorph react app might require css and would just crash with this plugin if we can't solve it.

nelix:
Rather than using templates for the html entrypoints, why not use react components so that the app can later be mounted on the document? Also react components have access to require() assets and what not.

NekR:
@nelix are you suggesting React by default, e.g. for everyone? Who wants React components, can just use appropriate loader.

SpaceK33z:
@nelix, not everyone uses React...

NickClark:
So I'm trying this with jade. Template compliation works, but I can't figure out how to use interpolations. This doesn't work for me:

`script( src="<%= o.htmlWebpackPlugin.files.chunks['app'] %>" )`

SpaceK33z:
@NickClark, can you try it using `<%= htmlWebpackPlugin.files.chunks['app'].entry %>`? In my underscore template I do it like this:

```html
<% for (var chunk in htmlWebpackPlugin.files.chunks) { %>
        <script src="<%=htmlWebpackPlugin.files.chunks[chunk].entry %>"></script>
<% } %>
```

NickClark:
It's not substituting at all.

My config look like like this: 

```
    new HtmlWebpackPlugin({
      filename: 'index.html',
      template: 'jade!./src/index.jade',
      ngApp:    'MyAppModule'
      inject:   false
      hash:     false
    })
```

NickClark:
Ok, got it. Had to switch to 'jade-loader', was using 'jade-html-loader' before because I was using for my templates. Also, I was able to use the normal jade interpolations. (ie. `#{}`).

ghislainf:
works well with :
```
loaders: [
  new HtmlWebpackPlugin({
    template: 'html!./client/index.html',
    inject: 'head'
  }),
]
```
but with the `UglifyJsPlugin` plugin, **the scripts and links tags was not append** on `<head>` 
```
loaders: [
  new HtmlWebpackPlugin({
    template: 'html!./client/index.html',
    inject: 'head'
  }),
  new webpack.optimize.UglifyJsPlugin({
    comments: '',
    compressor: {
      warnings: false,
      screw_ie8: true,
    },
  })
]
```

Try to add `UglifyJsPlugin` on [html-loader exemple](https://github.com/ampedandwired/html-webpack-plugin/tree/feature/loaders/examples/html-loader), you will see the issue

avaragado:
@ghislainf I've found a workaround for this problem. See #128.

wmadden:
I found that with multiple instances of the `HtmlWebpackPlugin` my `wepack-dev-server` doesn't rebuild. It outputs `webpack: bundle is now INVALID` and then hangs forever. 

I'd love to give more info for debugging but I don't know where to start - there's no output at all, and I can't find a way to get the dev server to give me more info.

SimenB:
I'm getting some warnings from bluebird when running tests. Doesn't make any difference, but it might be some subtle breakage?

```
Warning: a promise was created in a  handler but was not returned from it
    at processImmediate [as _immediateCallback] (timers.js:383:17)
```

Call stack is useless because of no async in IntelliJ...

image

mlegenhausen:
Some positive feedback. Works like a charm. Even with multiple favicons and manifest configurations inside the template.

In the documentation you should add that the html loader is used when it is already defined in the `loaders` section or you get a `module.exports` wrapped template back.

examples/favicon/webpack.config.js
```diff
@@ -0,0 +1,24 @@
...
+    loaders: [
+      { test: /\.css$/, loader: ExtractTextPlugin.extract('style-loader', 'css-loader') },
+      { test: /\.png$/, loader: 'file-loader' }
+    ]
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
+      title: 'HtmlWebpackPlugin example',
+      favicon: 'favicon.ico',
+      filename: 'favicon.html'
```
rayshan: Should this be `template.html`? Or rename the file to this name.

rayshan:
How do we do a custom favicon path now? I'm using `2.1.0`. I see favicon is removed from the default template. I have this defined:

```
new HtmlWebpackPlugin({
    template: "app/index.html",
    favicon: "app/favicon.png"
})
```

If I don't define an injection point for favicon in my template, it's pulling in the right file, and still injects:
`<link rel="shortcut icon" href="favicon.png">`

If I do define an injection point, such as:
`<link rel="shortcut icon" href=/<%= htmlWebpackPlugin.options.favicon %>>`

I get both of these:
`<link rel="shortcut icon" href="favicon.png">`
`<link rel="shortcut icon" href="/app/favicon.png">`

I would like to end up with the icon at root:
`<link rel="shortcut icon" href="/favicon.png">`

SimenB:
> How do we do a custom favicon path now?

Relative path from the template works for us (not using `favicon` option).

mlegenhausen:
Make sure you have added `link:href` attrs to your html-loader configuration. See the example:

```js
{
  test: /\.html$/,
  loader: 'html',
  query: {
    attrs: ['link:href']
  }
},
```

Should also be added to the documentation ;)

rayshan:
@SimenB do you mean like below, with a custom key name?

```
new HtmlWebpackPlugin({
    template: "app/index.html",
    xxx: "favicon.png"
})
```

`<link rel="shortcut icon" href=/<%= htmlWebpackPlugin.options.xxx %>>`

The result is what I would expect:
`<link rel="shortcut icon" href="/favicon.png">`

But previously with the `favicon` key, the favicon file gets copied over. With a custom key, it doesn't.

SimenB:
Just 
```js
new HtmlWebpackPlugin({
    template: "app/index.html"
})
```

and the `link` tag as 
```html
<link rel="shortcut icon" href="./path/to/favicon.ico">
```

rayshan:
@SimenB thanks for getting back. With your code, it's just hard coding to a predetermined path, no? I suppose we would use it with the `file-loader` to move the favicon.

SimenB:
Yup, that's right. I prefer having it in code and not in configuration, anyways :smile: 
You could always use a `resolve.alias` if you prefer that. In my exp, the `favicon` option is unecesarry if you provide your own template in version 2

jeshtan:
Does `template` loader support query syntax? For example
```
new HtmlWebpackPlugin({
    template: "myloader?hello=world!./app/index.html"
})
```
I tried and it seem not work.

mxstbr:
@jantimon `2.6.0`, getting the following error when I `require(html-webpack-plugin)`. Maybe related to @rayshan's [comment above](https://github.com/ampedandwired/html-webpack-plugin/pull/41#commitcomment-15384404)?

```
Error: Cannot find module './lib/compiler.js'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/a2517/sites/personal/github/react-boilerplate/node_modules/html-webpack-plugin/index.js:7:21)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/a2517/sites/personal/github/react-boilerplate/webpack/makewebpackconfig.js:6:25)
    at Module._compile (module.js:425:26)
    at Object.Module._extensions..js (module.js:432:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
```

The important part:

**`html-webpack-plugin/index.js:7:21`**

jantimon:
@mxstbr thanks for your feedback #164 was just fixed with https://github.com/ampedandwired/html-webpack-plugin/commit/c6cb00c4fe350f1573b35449e7bdffe994e2ea01

SimenB:
@jantimon You can include the whole `lib`-directory to avoid the problem in teh future :smile:

jantimon:
@SimenB Good point :+1: will do :)

SimenB:
Created a PR for it

haf:
I've been trying to grok the samples and this PR's code, but haven't made it yet. I want to transform this: `<link rel="apple-touch-icon" href="/img/favicon.png" />` into `<link rel="apple-touch-icon" href="/2de345f67ff65a43.png" />` where the alpha-numeric string is the webpack-compiled name. Since the html page is not using JS, I can't just do `require('./img/favicon.png')`, so how would I do it? An ico is not enough nowadays.

SpaceK33z:
@haf, you can use something like [html-loader](https://github.com/webpack/html-loader) or [underscore-template-loader](https://github.com/emaphp/underscore-template-loader) for this. When you have this installed, you can just use a `<link href="x.png">` and it will automatically compile it and change the url to the correct path.

jantimon:
@SpaceK33z concidering backwards compatibility and stability of underscore-template-loader - is there any reason why we should not use underscore-template-loader as default loader for this plugin?

SpaceK33z:
@jantimon, I don't think there's a good reason not to use it. It's lightweight and I think stable enough. Using it as default loader would certainly be interesting!

rubyless:
get some error when i run webpack;

rubyless@Mac ~/jobs/html-webpack-plugin/examples/jade-loader ±26c8725⚡ » webpack      
Hash: 8fa1534d971ffd522008
Version: webpack 1.12.2
Time: 919ms
                               Asset       Size  Chunks             Chunk Names
0714810ae3fb211173e2964249507195.png    54.1 kB          [emitted]  
                           bundle.js    21.6 kB       0  [emitted]  main
                         favicon.ico  766 bytes          [emitted]  
                          index.html    2.18 kB          [emitted]  
   [0] ./example.js 280 bytes {0} [built]
    + 8 hidden modules

ERROR in   Error: Parameter 'dependency' must be a Dependency
  
  - Compilation.js:347 Compilation.process [as _addModuleChain]
    [lib]/[webpack]/lib/Compilation.js:347:9
  
  - Compilation.js:423 Compilation.process [as addEntry]
    [lib]/[webpack]/lib/Compilation.js:423:7
  
  - SingleEntryPlugin.js:22 SingleEntryPlugin.<anonymous>
    [html-webpack-plugin]/[webpack]/lib/SingleEntryPlugin.js:22:15
  
  - Tapable.js:107 Compiler.applyPluginsParallel
    [lib]/[webpack]/[tapable]/lib/Tapable.js:107:14
  
  - Compiler.js:394 Compiler.compile
    [lib]/[webpack]/lib/Compiler.js:394:7
  
  - Compiler.js:203 Compiler.runAsChild
    [lib]/[webpack]/lib/Compiler.js:203:7
  
  - compiler.js:56 
    /Users/rubyless/jobs/html-webpack-plugin/lib/compiler.js:56:19
  
  - debuggability.js:180 Promise._execute
    [html-webpack-plugin]/[bluebird]/js/release/debuggability.js:180:9
  
  - promise.js:460 Promise._resolveFromExecutor
    [html-webpack-plugin]/[bluebird]/js/release/promise.js:460:18
  
  - promise.js:76 new Promise
    [html-webpack-plugin]/[bluebird]/js/release/promise.js:76:14
  
  - compiler.js:55 Object.compileTemplate
    /Users/rubyless/jobs/html-webpack-plugin/lib/compiler.js:55:10
  
  - index.js:47 
    /Users/rubyless/jobs/html-webpack-plugin/index.js:47:28
  
  - util.js:11 tryCatcher
    [html-webpack-plugin]/[bluebird]/js/release/util.js:11:23
  
  - promise.js:489 Promise._settlePromiseFromHandler
    [html-webpack-plugin]/[bluebird]/js/release/promise.js:489:31
  
  - promise.js:546 Promise._settlePromise
    [html-webpack-plugin]/[bluebird]/js/release/promise.js:546:18
  
  - promise.js:583 Promise._settlePromiseCtx
    [html-webpack-plugin]/[bluebird]/js/release/promise.js:583:10
  
  - async.js:130 Async._drainQueue
    [html-webpack-plugin]/[bluebird]/js/release/async.js:130:12
  
  - async.js:135 Async._drainQueues
    [html-webpack-plugin]/[bluebird]/js/release/async.js:135:10
  
  - async.js:16 Immediate.Async.drainQueues [as _onImmediate]
    [html-webpack-plugin]/[bluebird]/js/release/async.js:16:14
  


ERROR in ./main.css
Module build failed: Error: Parameter 'dependency' must be a Dependency
    at Compilation.process [as _addModuleChain] (/Users/rubyless/.nvm/versions/node/v4.2.0/lib/node_modules/webpack/lib/Compilation.js:347:9)
    at Compilation.process [as addEntry] (/Users/rubyless/.nvm/versions/node/v4.2.0/lib/node_modules/webpack/lib/Compilation.js:423:7)
    at SingleEntryPlugin.<anonymous> (/Users/rubyless/jobs/html-webpack-plugin/node_modules/webpack/lib/SingleEntryPlugin.js:22:15)
    at Compiler.applyPluginsParallel (/Users/rubyless/.nvm/versions/node/v4.2.0/lib/node_modules/webpack/node_modules/tapable/lib/Tapable.js:107:14)
    at Compiler.compile (/Users/rubyless/.nvm/versions/node/v4.2.0/lib/node_modules/webpack/lib/Compiler.js:394:7)
    at Compiler.runAsChild (/Users/rubyless/.nvm/versions/node/v4.2.0/lib/node_modules/webpack/lib/Compiler.js:203:7)
    at Object.module.exports.pitch (/Users/rubyless/jobs/html-webpack-plugin/node_modules/extract-text-webpack-plugin/loader.js:81:18)

jantimon:
Thanks @rubyless - could you please move this into an issue?

jantimon:
I am about to release 2.x - do you see any open points which should be addressed first?

mxstbr:
@jantimon do it!

SpaceK33z:
Yay! Please do :)

mxstbr:
Fuck yeah, good work @jantimon! 🎉🎉🎉

SimenB:
We had an issue in development on rebuilds after jumping from 2.0.4 to 2.6.something (don't remember exact version ottomh) when using custom handlebars. 
I'm on the bus on my way to work now, so I'll see if it's still a problem with the latest version

PR #45, opened 2015-05-21
Fixes error reporting

cgreening:
Fixes error reporting so that instead of:

Error in undefined

We get:

Error in HtmlWebPackPlugin: htmlWebpackPlugin.assets is deprecated - please use htmlWebpackPlugin.files instead


brumm:
Thanks for fixing this confusing error message. Since this seems to be more like a 'soft deprecation', could this be made into a warning instead?

PR #46, opened 2015-05-23
Use pretest

SimenB:
It's a bit cleaner IMO

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "default_inject_index.html"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "default_inject_index.html"
   ],
   "scripts": {
-    "test": "jshint -c .jshintrc *.js spec && jasmine-node --captureExceptions spec"
+    "pretest": "jshint *.js spec",
```
SimenB: `.jshintrc` is automatically looked up.

PR #49, opened 2015-06-01
Allow to inject javascript files into the head of the html page

jantimon:
Allows to inject javascript resources to the head instead of the body element

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,13 @@ Allowed values are as follows:
    You can specify a subdirectory here too (eg: `assets/admin.html`).
 - `template`: A html template (supports [blueimp templates](https://github.com/blueimp/JavaScript-Templates)).
 - `templateContent`: A html string or a function returning the html  (supports [blueimp templates](https://github.com/blueimp/JavaScript-Templates)).  
-- `inject`: Inject all assets into the given `template` or `templateContent`.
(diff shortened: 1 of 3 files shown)


tomatau:
Would it not make more sense to provide an inject map allowing some scripts to be injected at head and others into the bottom of the body?

This is the method for optimising above the fold loading for large SPAs that I feel would be the most common use case for head loading scripts.

Also it would be great to allow some script/link inlining in the head when very very small content as per Google's CSS optimisation guidelines.

jantimon:
@tomatau - that's a great idea. 

Could you propose a clean API which allows placing all, or only some picked css/js files to the head/body?

tomatau:
Keep the same options for true/false or placing all at 'head'/'body'.

In addition to the current API, possibly something like:
```js
inject: {
  'head': ['chunkname1', /file-name-regex1/],
  'body': ['chunkname2', /file-name-regex2/],
}
```

This would allow for the easy inclusion of chunks which I feel would be cleanest but also allow for specific files.

This could cause problems when a file/chunk is selected for both head and body?

jantimon:
This looks quite nice.
I guess all not specified js files should be injected to the body and all not specified css files should go into the head. - What do you think?

If the configuration is wrong it should just add some warnings that the file will be placed to the head and to the body element.

Do you have time to create a merge request for this?

tomatau:
Re merge req: I would really like to - however I'm not familiar with the internals and time isn't on my side right now so it would take me a very long time to put this together.

PR #58, opened 2015-06-10
add support for managing 'live-reload' url via options

tony-kerz:


tony-kerz:
hi @jantimon,

this is just to get any early feedback, i'm just in the process of adding a test as well.

best,
tony.

jantimon:
Looks like this conflicts with webpack --inline. See http://stackoverflow.com/questions/31448534/inject-webpack-dev-server-js

tony-kerz:
i haven't been able to get that `--inline` option to work, have you ever used it successfully?

jantimon:
Did you try http://127.0.0.1:8080/webpack-dev-server/

PR #66, opened 2015-06-30
Fix example template

cesarandreu:
I copied the example, it gave me errors and it took me a moment to notice why.


jantimon:
Thanks!

PR #81, opened 2015-08-14
Passing true to minify does nothing

SimenB:
All `html-minifier` options are `false` by default, so passing just true does nothing to the output

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Allowed values are as follows:
 - `templateContent`: A html string or a function returning the html  (supports [blueimp templates](https://github.com/blueimp/JavaScript-Templates)).  
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
 - `favicon`: Adds the given favicon path to the output html.
-- `minify`: `true | {...} | false` Set to true or pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -130,9 +130,7 @@ HtmlWebpackPlugin.prototype.emitHtml = function(compilation, htmlTemplateContent
   // Minify the html output
   if (this.options.minify) {
     var minify = require('html-minifier').minify;
-    // If `options.minify` is set to true use the default minify options
-    var minifyOptions = _.isObject(this.options.minify) ? this.options.minify : {};
-    html = minify(html, minifyOptions);
+    html = minify(html, this.options.minify);
```
SimenB: IDK if this change is necessary, as it might break someone's stuff. Passing in an empty object if it was just `true` before did nothing to the output though, so I'm guessing people using the `minify`-option do pass an object

README.md
```diff
@@ -67,7 +67,7 @@ Allowed values are as follows:
 - `templateContent`: A html string or a function returning the html  (supports [blueimp templates](https://github.com/blueimp/JavaScript-Templates)).  
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
 - `favicon`: Adds the given favicon path to the output html.
-- `minify`: `true | {...} | false` Set to true or pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
+- `minify`: `{...} | false` Pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
```
SimenB: Is it "an html" or "a html"? :P

PR #84, opened 2015-09-05
Add more tests and cleanup for PR #41

SpaceK33z:
I added some more tests that use the `html-loader` and `underscore-template-loader` to test if loaders work.

Also, I accidentally fixed the `manifest` option, and added some tests for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,6 @@ Allowed values are as follows:
 - `template`: Path to the template. Supports loaders e.g. `html!./index.html`.
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
 - `favicon`: Adds the given favicon path to the output html.
-- `manifest`: Adds the given manifest path to the output html. 
 - `minify`: `{...} | false` Pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
 - `hash`: `true | false` if `true` then append a unique webpack compilation hash to all
   included scripts and css files. This is useful for cache busting.
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ Allowed values are as follows:
 - `template`: Path to the template. Supports loaders e.g. `html!./index.html`.
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
 - `favicon`: Adds the given favicon path to the output html.
+- `manifest`: Adds the given manifest path to the output html. 
 - `minify`: `{...} | false` Pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
 - `hash`: `true | false` if `true` then append a unique webpack compilation hash to all
   included scripts and css files. This is useful for cache busting.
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -185,10 +185,6 @@ HtmlWebpackPlugin.prototype.evaluateCompilationResult = function(compilation, co
   if(!compilationResult) {
     return Promise.reject('The child compilation didn\'t provide a result');
   }
-  if(typeof source === 'string') {
```
jantimon: this part allows to pass text html files instead of templates - why would you like to remove it?

SpaceK33z: Ah sorry, did not know that. I ran the code through Istanbul and found out that this line was not tested. I couldn't find a usecase for it so I removed it. Could you give an example of when this would be used? Then I can write a test for it.

index.js
```diff
@@ -67,6 +61,12 @@ HtmlWebpackPlugin.prototype.apply = function(compiler) {
               assets.favicon = faviconBasename;
             });
         }
+        if (self.options.manifest) {
```
jantimon: This isn't necessary as you can use a webpack app cache plugin - like [appcache-webpack-plugin](https://github.com/lettertwo/appcache-webpack-plugin) will add a `.appcache` to the resources and will be picked up by the html-webpack-plugin: https://github.com/ampedandwired/html-webpack-plugin/blob/feature/loaders/index.js#L349

index.js
```diff
@@ -31,12 +31,6 @@ function HtmlWebpackPlugin(options) {
   if(this.options.template.indexOf('!') === -1) {
     this.options.template = require.resolve('./loader.js') + '!' + path.resolve(this.options.template);
   }
-  // Resolve template path
```
jantimon: Why would you remove this?
It would probably break the rendering.

README.md
```diff
@@ -66,6 +66,7 @@ Allowed values are as follows:
 - `template`: Path to the template. Supports loaders e.g. `html!./index.html`.
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
 - `favicon`: Adds the given favicon path to the output html.
+- `manifest`: Adds the given manifest path to the output html.
```
jantimon: Not needed - will be done by a webpack-appcache plugin

jantimon:
Wow cool looks like a lot progress! :)

index.js
SpaceK33z: After I added tests, I ran the code through Istanbul. It showed that this regex never matches anything to replace.
After removing it, all tests still work. Could you give an example when it would replace anything?

jantimon: Maybe I am wrong but I believe this is necessary if this plugin is used as a submodule and the template lies in the parent module (which is true for most use cases).

SpaceK33z: Okay, but these lines are not tested. Could you maybe write some tests for that scenario? I have no idea how to write a test for that usecase.

SpaceK33z:
I've fixed your point about the manifest now, and added a test that uses `appcache-webpack-plugin` to generate a manifest.

About the lines I removed from `index.js`; this was because these lines were not tested and I couldn't find a usecase for it. Do you maybe have the time to write tests for these? If yes, do you want me to re-add these lines?

jantimon:
Yeah please add those deletes lines again so I can merge it :+1:

SpaceK33z:
Okay, the `this.options.template` might still be necessary. But in the [second thing I deleted](https://github.com/ampedandwired/html-webpack-plugin/blob/feature/loaders/index.js#L188), the `source` variable is _always_ undefined. It is only defined right under the `if` statement. Do you really want me to put that back? Looks like a bug.

jantimon:
Oh you are right - looks like old code which was left after a refactoring.

SpaceK33z:
Cool, I have know re-added the template path regex. Could you write a test for it, or give me a usecase? All of the tests pass with this path regex removed, and it is really hard to read.

PR #85, opened 2015-09-09
Add support for webpack.BannerPlugin

SpaceK33z:
Previously the first 3 characters were stripped from the beginning of the source; it assumed these characters are `var`. When using the webpack.BannerPlugin, the source starts with the given comment and after that `var result` is used. This is fixed now and I've added a test for this.

This may not be the most clean solution, but it works.

### DIFF
--- a/index.js
+++ b/index.js
@@ -145,7 +145,7 @@ HtmlWebpackPlugin.prototype.compileTemplate = function(template, outputFilename,
   childCompiler.apply(
     new NodeTemplatePlugin(outputOptions),
     new NodeTargetPlugin(),
-    new LibraryTemplatePlugin('result', 'var'),
+    new LibraryTemplatePlugin('HTML_WEBPACK_PLUGIN_RESULT', 'var'),
     new SingleEntryPlugin(this.context, template),
     new LoaderTargetPlugin('node'),
     new webpack.DefinePlugin({ HTML_WEBPACK_PLUGIN : 'true' })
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -185,8 +185,16 @@ HtmlWebpackPlugin.prototype.evaluateCompilationResult = function(compilation, co
   if(!compilationResult) {
     return Promise.reject('The child compilation didn\'t provide a result');
   }
-  // Strip the leading 'var '
-  var source = compilationResult.source().substr(3);
+  var source = compilationResult.source();
+  // Strip the leading 'var ' if present.
+  // If webpack.BannerPlugin is used, `source` starts with given comment.
+  if (_.startsWith(source, 'var')) {
```
jantimon: I like your idea and comment - however I would not add the if statement.
Just add `source = source.replace('var result =', '');` for both cases

SpaceK33z: That would be very tricky, as it's possible `var result = ` exists in the source code itself.

SpaceK33z:
It seems that `LibraryTemplatePlugin ` does not work anymore if you change `result` to another variable. What would be the advantage of changing this?

jantimon:
It would allow us to use a more specific replace like `source = source.replace('var WEBPACK_HTML_PLUGIN_RESULT =', '');`.

What does not work?

SpaceK33z:
Ah nice! That's a lot better. I've implemented this now.

jantimon:
Released in 2.0.3

PR #89, opened 2015-09-28
Fix publicPath for feature/loaders

SpaceK33z:
I've incorporated 396cdedf8a554465d00797bb4d279b085e2e05d8 into the `feature/loaders` branch, which should fix the failing tests for old Node.js versions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -335,7 +335,7 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function(compilation, chun
       path.relative(path.dirname(self.options.filename), '.');
 
   if (publicPath.length && publicPath.substr(-1, 1) !== '/') {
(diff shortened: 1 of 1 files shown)


jantimon:
Actually I don't get why we need the url module at all.
`publicPath += '/'` should be just fine - why did you change it?

SpaceK33z:
I didn't change it, I just merged the change from master at 4809fa5804a800ec0b1b7440dbe53293ce6279d2. The tests also pass with `publicPath += '/'`, but I don't know if they also pass on Windows.

jantimon:
Public path is a web url isn't it?

SpaceK33z:
That's right. I've changed it to `publicPath += '/'`. If it does not work for Windows, someone who uses Windows should write a test for it that also fails on non-windows machines.

PR #90, opened 2015-09-29
Append new line to inserted tags

L42y:


jantimon:
Sorry there is no need for those line breaks - you can write a custom template if you need them.

dominique-mueller:
Any specific reason this one got declined? I really feel that line breaks could improve readability of the build output. If needed, the mifier could then go on removing those line-breaks.

rshingleton:
The fact that there are no line breaks is pretty annoying. It's, as dominique-mueller eluded to, like minifying the script/style tags by default, which feels like unexpected behavior.

tsurumure:
eh.. Guys. So do I felling that line breaks is importance. obsessive compulsive disorder :）

PR #92, opened 2015-10-02
FIX missing [hash] interpolation in publicPath

thomasguillory:
Webpack gives the possibility to use the build [hash] in
output.publicPath option:
https://webpack.github.io/docs/long-term-caching.html

However HtmlWebpackPlugin is reading directly the raw option, preventing the
user to use this option when using it.


jantimon:
Hey @thomasguillory do you have the time to port your fix to the 2.0 version?
Would be awesome: https://github.com/ampedandwired/html-webpack-plugin/tree/feature/loaders

chrillo:
+1 I need this feature as well

PR #93, opened 2015-10-04
Fix minify option

NekR:
It used `this` instead of `self` and there was not errors because they there caught and not forwarded back to promise (forgotten `return` before `Promise.reject()`).

Btw, why need to `try..catch` if you are already in `Promise`, shouldn't it be auto-rejected on error?


jantimon:
Correct we should drop the try catch

NekR:
@jantimon okay, let me update it.

NekR:
@jantimon :+1:

PR #102, opened 2015-10-18
Don't explode on missing object to minify, just warn

### LINKED ISSUE #97: Regression when using minify in v1.6.2
#81 is a breaking change, it shouldn't be released in v1.6.2. I was passing in true or false depending on my environment, (admittedly I hadn't looked at my output, I just assumed it was being minified) and the v1.6.2 release broke my builds.

Adding a check such as: `minify = minify === true ? {} : minify` could be released as v1.6.3 and it would preserve backwards compatibility.

SimenB:
Fixes #97

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 var fs = require('fs');
 var path = require('path');
- var urlModule = require('url');
+var urlModule = require('url');
 var _ = require('lodash');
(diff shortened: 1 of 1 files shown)


SimenB:
@ampedandwired @jantimon ping

jantimon:
This was just an issue when upgrading and setting minify to true - should be solved by now.

PR #107, opened 2015-10-28
Add a body template parameter to the default template

insin:
This makes it more convenient to add placeholder content or a render target in `<body>` while making use of everything else the default template provides and without having to define an entire custom template around it.


jantimon:
I would say that the inject feature with a really small index.html offers a really easy and clean solution for this.

bebraw:
@jantimon Can you give a small example? This use case comes up with React in particular.

PR #131, opened 2015-12-02
Turn off bluebird warnings for "runaway" promise.

nicksrandall:
The issue isn't causing any problems so I'm ok with just turning off the warning. This probably won't be a problem in V2 anyway.


jantimon:
Would prefer to fix it instead of turning of the warnings

PR #149, opened 2015-12-27
Introduce injectDivWithId-option

SimenB:
The only reason we have our own template is to add a div to be used as container for our react app. This option allows us to avoid that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Allowed values are as follows:
    You can specify a subdirectory here too (eg: `assets/admin.html`).
 - `template`: Path to the template. Supports loaders e.g. `html!./index.html`.
 - `inject`: `true | 'head' | 'body' | false` Inject all assets into the given `template` or `templateContent` - When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element.
+- `injectDivWithId`: `'some id' | false` Inject a `div` with the given `id` into the body of the document. Only works in conjunction with the `inject` option. This is useful for creating a container for React applications.
 - `favicon`: Adds the given favicon path to the output html.
 - `minify`: `{...} | false` Pass a [html-minifier](https://github.com/kangax/html-minifier#options-quick-reference) options object to minify the output.
 - `hash`: `true | false` if `true` then append a unique webpack compilation hash to all
--- a/index.js
+++ b/index.js
@@ -20,6 +20,7 @@ function HtmlWebpackPlugin(options) {
     filename: 'index.html',
     hash: false,
     inject: true,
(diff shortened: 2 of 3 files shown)


bebraw:
I feel that's very specific. Maybe it would be better to give more control over the attributes to inject? I mean something like:

```javascript
injectDiv: {
  // attributes go here
  id: 'foobar'
}
```

Something even more generic could work.

Btw, [html-webpack-template](https://www.npmjs.com/package/html-webpack-template) seems to solve this, and some other problems, well. One option would be to skip this PR and link to a user space solution like that instead.

SimenB:
I'm not sure what other attributes you'd like, but having it accept both a string for id, and object for everything else shouldn't be a problem. But if so I'd like to get someone on the project giving it the go ahead.

Did not know of the template, could use that instead!

bebraw:
Yeah. Let's see what @jantimon has to say.

jantimon:
First of all thanks for all the work @SimenB 

Imho the plugin is already doing to much - we should rather add some events so people can add mobile favicons or google analytics by their own. 
E.g.  `compilation.applyPluginsAsync('html-plugin-before-inject', ...)`
What do you think?

bebraw:
@jantimon Hooks sound like a good idea as it allows you to keep the core neat. Perhaps you can even drop some functionality in the next major release through this? Simplicity is good. :+1:

SimenB:
That does sound like a very good idea!

jantimon:
@bebraw @SimenB Added three events in 2.3.0: https://github.com/ampedandwired/html-webpack-plugin/tree/feature/loaders#events - would love to get your feedback

bebraw:
@jantimon That `htmlPluginData.html` is just a plain string, right? Is the idea that you use something like [cheerio](https://www.npmjs.com/package/cheerio) to inject the data you want? I don't know if there's a better way, though. Cheerio isn't too bad for something like this.

Another question is how to compose? If you see that as a good extension point and it's possible to compose, then you can push common logic to separate packages and glue them together. A bit like with [html-webpack-template](https://www.npmjs.com/package/html-webpack-template).

jantimon:
@bebraw I didn't use cheerio for performance reasons. Whenever a file is changed (css, js, png, .., whatever) it would have to parse the html file with cheerio for every html plugin instance.
I absolutely see an advantage of a cheerio instance over a html string.
Do you think it is fast enough and worth it?

bebraw:
@jantimon I guess the current API is enough for now. It's easy enough to deal with it in the user space. The plugin is big already so no need to introduce yet another dep perhaps. :smile:

I know from experience that it comes with some overhead, especially if your files are big. That's probably not an issue here.

PR #154, opened 2016-01-03
Fixed broken output.publicPath when using url

cementovoz:
When using url to cdn in output.publicPath, as example - http://cdn.example.com - path.join replace two slash to one - http:/cdn.example.com, it's broken url, and resources don't load.


jantimon:
This was already introduced however as you see in the broken tests it doesn't work for relative files.

cementovoz:
Hm, I started investigation, and found that my issue reproduced when publicPath is url and without last slash, what is wrong. This changes not needed.

PR #159, opened 2016-01-09
improve test case

island205:
make correct `expect`.


jantimon:
Wow nice!

Could you please move the `done()` below the `existsSync`?

PR #160, opened 2016-01-09
Support lodash template's HTML "escape" delimiter (<%- %>)

anjianshi:
Without this, it will throw `ERROR in Template execution failed: ReferenceError: _ is not defined` when use `<%- something %>` in template.
The template functions referenced `_.escape`, and it lost the reference when transform it to string.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ function HtmlWebpackPlugin(options) {
     excludeChunks: [],
     title: 'Webpack App'
   }, options);
-  // If the template doesn't use a loader use the blueimp template loader
+  // If the template doesn't use a loader use the lodash template loader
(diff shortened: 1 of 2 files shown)


jantimon:
Okay sounds valid - You could also have used the `tpl-loader` to achieve the same.

jantimon:
Do you think we should also change https://github.com/ampedandwired/html-webpack-plugin/blob/feature/loaders/default_index.html#L5 to escape the title?

anjianshi:
I think it's a good idea. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title
> The HTML <title> element defines the title of the document, shown in a browser's title bar or on the page's tab. It can only contain text and any contained tags are not interpreted.

PR #168, opened 2016-01-13
support access compiled assets content

kerryChen95:
Support insert compiled assets content into html template like: 
`<script><%= htmlWebpackPlugin.files.chunks.main.source() %></script>`

### DIFF
--- a/index.js
+++ b/index.js
@@ -370,7 +370,7 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function(compilation, chun
     var chunkName = chunk.names[0];
 
     assets.chunks[chunkName] = {};
-    assets.chunks[chunkName].source = compilation.assets[chunk.files[0]].source.bind(compilation.assets[chunk.files[0]])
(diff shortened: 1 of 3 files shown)


jantimon:
Cool idea :+1: 
I fear this would make it really hard to modify the assets object like in the following test/example:
https://github.com/ampedandwired/html-webpack-plugin/blob/feature/loaders/spec/HtmlWebpackPluginSpec.js#L687

kerryChen95:
I deep into `compilation` object and find methods output compiled assets content, so I just mount those methods to be accessed from html template context, haha~

kerryChen95:
@jantimon So would you accept this pull request? Many people hope that the html-webpack-plugin package on NPM can support inline compiled assets into html template

jantimon:
This doesn't look like a complete solution as it works only for chunks (so you can't inline css) and it makes it harder to modify the assets object with the hook system.

jantimon:
What do you think about something like https://github.com/ampedandwired/html-webpack-plugin/pull/191 ?

jantimon:
It would work for css js manifests files and does support altering the assets object with the events.

kerryChen95:
Cool, I agree with #191.
There is also desire to inline css and webpack manifest files.

PR #173, opened 2016-01-15
Remove lodash as dependency, fix #172

### LINKED ISSUE #172: upgrade incompatible
When I upgrade `html-webpack-plugin` from `v2.1.0` to `v2.6.0`. I get an error:

```bash
ERROR in ./~/html-webpack-plugin/lib/loader.js!./app/index.html
    Module not found: Error: Cannot resolve module 'lodash' in /Users/creeper/work/projects/react-base/app
     @ ./~/html-webpack-plugin/lib/loader.js!./app/index.html 1:8-25
```

I checked the source code, and found the reason:

```js
// lib/loader.js line26
return 'var _ = require("lodash"); module.exports = ' + template;
```

So it makes the `lodash` the **unneeded** dependency. And this change is from `v2.5.0`.

creeperyang:
This PR removes changes included with `commit b3913d04001f031342794c71acfc4a1bae4e01b3`.

Why?

- `commit b3913d` makes `lodash` a dependency for user's app. Not every user will use `lodash`. Those who don't use lodash will run into error.
- When I upgrade `html-webpack-plugin@2.1` to `html-webpack-plugin@2.6`, my build script crashed because of reason above. The `commit b3913d` leads to incompatibility.
- I think `escape` is not that important and we have other approach.

More details:

- <https://github.com/ampedandwired/html-webpack-plugin/pull/160> 
- <https://github.com/ampedandwired/html-webpack-plugin/issues/172>

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -22,5 +22,5 @@ module.exports = function (source) {
   // Use underscore for a minimalistic loader
   var options = loaderUtils.parseQuery(this.query);
   var template = _.template(source, options);
-  return 'var _ = require("lodash"); module.exports = ' + template;
+  return 'var _ = require("' + require.resolve('lodash') + '");module.exports = ' + template;
 };


jantimon:
What about `'var _ = require("' + require.resolve('lodash') + '");'`?

creeperyang:
@jantimon A better solution :smile:

PR #179, opened 2016-01-18
Child compilation must use a separate cache object

sokra:
fix for https://github.com/webpack/webpack/issues/1890

yes, I know, this should be done automatically by webpack when creating a child compilation...
I'll add it in the next major version...

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -50,6 +50,13 @@ module.exports.compileTemplate = function compileTemplate(template, context, out
     new SingleEntryPlugin(this.context, template),
     new LoaderTargetPlugin('node')
   );
+  childCompiler.plugin("compilation", function(compilation) {
+    if(compilation.cache) {
+      if(!compilation.cache[compilerName])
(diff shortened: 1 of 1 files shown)


jantimon:
I saw this in extract-text-plugin however there was no comment (like in this merge request) which explains what this code does and why

jantimon:
Thanks a lot @sokra this fixed different issues including incomplete cache modules causing "HtmlWebpackPlugin Error: No source available"

joetidee:
How is it possible to view which version of the html-webpack-plugin this fix was committed to (in order to set the correct version in my package.json) ?

aimettii:
Update html-webpack-plugin did not help me. I use Laravel-mix, for hot replacement https://github.com/JeffreyWay/laravel-mix/blob/master/docs/hot-module-replacement.md

digitaltopo:
Page does not reload when I make a change to a template or a partial (handlebars), using custom templates in html-webpack-plugin.

PR #187, opened 2016-01-21
Add jade dev dependency to fix tests

SpaceK33z:
3 tests failed without this dependency.


jantimon:
This is only necessary for npm3 but I guess dev dependencies wont hurt.

PR #192, opened 2016-01-25
Implemented new chunk sorting mode 'dependency'

ruehl:
This pull request attempts to solve the chunk ordering problem by automatically calculating a proper order of chunks to be included in the HTML file based on the dependencies between all chunks.

`chunksSortMode` now accepts the new value "dependency", which orders chunks based on their relation within the dependency tree as devised by Webpack. Technically it performs a topological sorting over the directed acyclic graph (DAG), which is defined by all chunks (vertices) and parent-child relations (edges) as devised by Webpack. This results in lowest-level chunks (which have no incoming edges) to be listed before higher-level chunks (which have one or more incoming edges each).

Sorting fails, 

* if the dependency graph contains on or more cycles.
* chunks do not contain (correct) dependency information, i.e. have no "parents" property set

Added unit test for new sorting mode.

Fixed a few typos in other unit tests...sorry, couldn't resist ;-)


jantimon:
Wow that looks really nice :)

Just a minor thing - could you please adapt the code style to match the other lib files?

Instead of 

```js
return {
     sortChunksByDependencies: sortChunksByDependencies
   };
 };
```

Just 
```js
module.exports.sortChunksByDependencies = function sortChunksByDependencies() {

}
```

https://github.com/ampedandwired/html-webpack-plugin/blob/master/lib/compiler.js#L32

And we might move all other sorting methods to the chunksorter.js e.g.:

```
module.exports.none = function (chunks) { 
 return chunks;
};
modules.exports.auto = function() {
  // sorting
};
module.exports.dependency = function() {
  // sorting
};
modules.exports.default = module.exports.auto;
```

What do you think?

lib/chunksorter.js
```diff
@@ -0,0 +1,72 @@
...
+    (containing the chunk id) and "parents" (array containing the ids of the
+    parent chunks). Must not be null/undefined or empty
+
+    @return A topologically sorted version of the input chunks, or null if
+    no such order could be calculated (e.g. because the chunks and their
+    parent relations did not define an directed acyclic graph).
+  */
+  function sortChunksByDependencies (chunks) {
+    if (!chunks) {
+      return null;
```
jantimon: return chunks

lib/chunksorter.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    });
+
+    // We now perform a topological sorting on the input chunks and built edges
+    var sortedVertices = null;
+
+    try {
+      sortedVertices = toposort.array(chunks, edges);
+    } catch (err) {
+      return null;  // Error during sort
```
jantimon: fall back to auto (id) sort

lib/chunksorter.js
ruehl: I agree. I will rewrite the chunk sorter according your suggestions.

lib/chunksorter.js
ruehl: The only problem I have with that is that if a developer explicitly selects a chunkSortMode of "dependency" and the Chunk Sorter internally falls back to another sorting mode, the developer might get confused (as the returned order might not be equal to the expected topological sorting). 

Maybe if you decide one day to make dependency sort the default sort, we could define "auto" as "dependency sort, followed by id-sort as fallback" explicitly in the documentation?

PR #199, opened 2016-01-28
implemented possibility to specify bodyContent

philipshurpik:
Implemented possibility to specify some specific bodyContent.
For example root element in ReactJS apps: '<div id="root"></div>'


jantimon:
Just use it together with the https://github.com/jaketrent/html-webpack-template :)

PR #201, opened 2016-01-31
Compilation fails with extract-text-webpack-plugin

pirelenito:
```
Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
```

The use case that triggered this behavior for me was doing React’s `renderToString` of a component that has a dependency to a `.css` file.

Apparently the `ExtractTextPlugin` loader is configured properly, but the plugin not when we [create the createChildCompiler](https://github.com/ampedandwired/html-webpack-plugin/blob/master/lib/compiler.js#L44)

**This commit implements a test that demonstrates the issue.** but I'm not sure where to begging to fix it. Could you provide me some direction to start? I would gladly update this PR with a solution.

### DIFF
--- a/spec/HtmlWebpackPluginSpec.js
+++ b/spec/HtmlWebpackPluginSpec.js
@@ -356,6 +356,28 @@ describe('HtmlWebpackPlugin', function () {
     }, ['<link href="styles.css"'], null, done);
   });
 
+  it('should work with the css extract plugin while using a JavaScript template', function (done) {
+    var ExtractTextPlugin = require('extract-text-webpack-plugin');
+    testHtmlPlugin({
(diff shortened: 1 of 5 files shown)


jantimon:
This can only be fixed by the extract-text-webpack-plugin.

jantimon:
What if we would add an ExtractTextPlugin with {disabled:true} to the child compilation?

calumjames:
Hi all,

What's the status on this PR being merged? I'm seeing the same issue with this plugin when using a JavaScript template for server-side rendering a React component in to the index.html.

jantimon:
This pr is not a fix but illustrates the problem.

As it is quite outdated I will close it for now

Issue #205, opened 2016-02-03
Universal (//domain.com/asset.css) publicPath and ExtractTextPlugin: css is missing


smolyakoff:
In my webpack configuration I have a public path configured in such a way that it is not dependent on HTTP/HTTPS server

```js
{
   output: {
      publicPath: '//localhost:8080/'
   }
}
```

The output chunk names look like this:
```js
[ '//localhost:8080/vendor-3a8b12e879b7905f4358.js',
  '//localhost:8080/vendor-3a8b12e879b7905f4358.js.map' ]
[ '//localhost:8080/app-3a8b12e879b7905f4358.js',
  '//localhost:8080/styles.css',
  '//localhost:8080/app-3a8b12e879b7905f4358.js.map',
  '//localhost:8080/styles.css.map' ]
```

CSS is not inserted into the output HTML, probably because *path.extname()* is used to get an .css extension

Can we add a special check to cover this case?

jantimon:
In webpack you should always specify `http` or `https` in your public path during development as it will break background images in css files if you use source maps.
But feel free to create pull request for this issue :+1:

smolyakoff:
@jantimon Thanks for you notice, haven't faced background images yet. This could still be useful when building a package for deployment, urls should work in this case. I've created a pull request, could you take a look?

PR #207, opened 2016-02-03
Protocol relative urls support

smolyakoff:
Support for protocol relative urls, for ex. '//domain.com/styles.css'. See issue #205

### DIFF
--- a/index.js
+++ b/index.js
@@ -378,9 +378,11 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
 
     // Gather all css files
     var css = chunkFiles.filter(function (chunkFile) {
+      // Supporting protocol relative urls, for ex. '//domain.com/styles.css'
+      var cssChunkName = chunkFile.replace(/^\/{2}/, '');
       // Some chunks may contain content hash in their names, for ex. 'main.css?1e7cac4e4d8b52fd5ccd2541146ef03f'.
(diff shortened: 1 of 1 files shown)


jantimon:
Does this really solve your issue?

Extname does not care about the path:

```
> var path = require('path');
> path.extname('##¿¿¿¿\\\//a/b.css')
'.css'
```

smolyakoff:
My case is solved, as chunk name is //domain.com/styles.css 
Do you think there is a better way to extract an extension? '''str.split()''' should also work

jantimon:
Which node version do you use?
`path.extname` does not care about the protocol.

smolyakoff:
```cmd
C:\Windows\System32>node -v
v4.2.1

C:\Windows\System32>node
> require('path').extname('//domain.com/styles.css')
require('path').extname('//domain.com/styles.css')
''
>
```

Returns empty string in this case

smolyakoff:
By the way, I found a workaround to remove publicPath from webpack config when **not** in development. Then it just generates relative urls (without any protocols and domains) and seems to work. 

Can close  the issue, if you think that chunk names with protocol relative urls is not a valid use case for html-webpack-plugin

jantimon:
Ah - windows ;)
Could you please add a unit test for it so it won't be broken again by future changes?

jantimon:
Changed it a little bit and merged it in 2.8.2!
Thamks for your help

PR #215, opened 2016-02-10
Fix favicon location

jirikolarik:
Consider publicPath in `favicon.ico` location

Resolves https://github.com/ampedandwired/html-webpack-plugin/issues/208


jantimon:
Does this also work if you specify the pats `/some` with out the trailing slash?

jirikolarik:
It doesn't. I think that we should have absolute url for favicon if publicPath is not specified. I can add these two updates in another PR

jantimon:
Why should it be absolute? this would destroy the backwards compatibility

jirikolarik:
Because current implementation is wrong. Favicon, same as assets (js, css) should be still on the same location and not relative

PR #216, opened 2016-02-11
Webpack 2.0.7-beta failing compile

RahavLussato:


RahavLussato:
i also upgraded lodash to the maintained version

foglerek:
This will break the plugin for any users who provide a custom template to the internal lodash template compiler, i.e:

```
plugins: [
  new HtmlWebpackPlugin({
    title: 'Custom template',
    template: 'my-index.ejs', // Load a custom template
    inject: 'body' // Inject all scripts into the body
  })
]
```

and force them to rewrite their templates to append `data`. There are less invasive solutions to this problem which will keep the plugin compatible between Webpack 1 and 2, although they require more work. See https://github.com/ampedandwired/html-webpack-plugin/issues/213

jantimon:
Closed because of #219

PR #219, opened 2016-02-12
Fix Webpack2 compatibility. Closes #213.

### LINKED ISSUE #213: Webpack 2.0.7-beta failing compile
Hi,

I'm having issues compiling with a very basic setup using Webpack 2. Getting two sets of errors ([...] = redacted):
```
ERROR in   Error: Child compilation failed:
  Module parse failed: [...]/node_modules/html-webpack-plugin/lib/loader.js!  [...]/node_modules/html-webpack-plugin/default_index.ejs Unexpected token   (1:0)
  You may need an appropriate loader to handle this file type.
  | <!DOCTYPE html>
  | <html>
  |   <head>:
  SyntaxError: Unexpected token (1:0)

  - compiler.js:78
    [...]/[html-webpack-plugin]/lib/compiler.js:78:16

  - Compiler.js:243 Compiler.<anonymous>
    [...]/[webpack]/lib/Compiler.js:243:10

  - Compiler.js:432
    [...]/[webpack]/lib/Compiler.js:432:12

...
```
and 
```
Child html-webpack-plugin for "index.html":

    ERROR in ./~/html-webpack-plugin/lib/loader.js!./~/html-webpack-plugin/default_index.ejs
    Module parse failed: [...]/node_modules/html-webpack-plugin/lib/loader.js![...]/node_modules/html-webpack-plugin/default_index.ejs Unexpected token (1:0)
    You may need an appropriate loader to handle this file type.
    | <!DOCTYPE html>
    | <html>
    |   <head>
```

The config is 
```
    context: __dirname + '/app',
    entry: { 'app' : './entry' },
    output: {
        path: __dirname + '/public/assets/[hash]',
        publicPath: "/assets/[hash]",
        filename: '[name].js'
    },
...
    devServer: {
        contentBase: "./public",
    }
...
    plugins: [
        new HtmlWebpackPlugin()
    ],
```

Any ideas?

foglerek:
Fixes the issues with lodash templates using `with()` not being compatible with Webpack 2, while maintaining backwards compatibility.

Do not be put off by the Function() constructor, that is how the lodash template works internally as well.

The only downside to this fix is that, should arguments passed to the template functions change, they need to be added both in `index.js` and `template.js`. I refactored out the template compiler to a separate file for this reason, to ease this.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -3,6 +3,7 @@
 
 var _ = require('lodash');
 var loaderUtils = require('loader-utils');
+var templateConstructor = require('./template');
 
 module.exports = function (source) {
   if (this.cacheable) {
@@ -22,6 +23,8 @@ module.exports = function (source) {
   }
   // Use underscore for a minimalistic loader
   var options = loaderUtils.parseQuery(this.query);
-  var template = _.template(source, options);
-  return 'var _ = require(' + loaderUtils.stringifyRequest(this, require.resolve('lodash')) + ');module.exports = ' + template;
+  // Workaround for Webpack 1 & 2 compatibility.
+  // See issue#213.
+  var template = _.template(source, _.defaults(options, { variable: 'data' }));
+  return 'var _ = require(' + loaderUtils.stringifyRequest(this, require.resolve('lodash')) + ');module.exports = ' + templateConstructor(template);
 };
--- a/lib/template.js
+++ b/lib/template.js
@@ -1,0 +1,21 @@
+/* This module returns a function constructor. It is necessary to maintain
+   backward compatibility with Webpack 1 while supporting Webpack 2.
+   Webpack 2 does not allow with() statements, which lodash templates use to unwrap
+   the parameters passed to the compiled template inside the scope. We therefore
+   need to unwrap them ourselves here. This is essentially what lodash does internally
+   (see lodash/template).
+
+   See issue#213 for more information.
+*/
+'use strict';
+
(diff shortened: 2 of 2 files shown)


jantimon:
sorry this is not backwards compatible :/

foglerek:
Hmm, could you tell me which cases it breaks?

jantimon:
@foglerek - sorry I was wrong. After reviewing this with a little bit more time I got it.

You are just exposing all values to get rid of the with.
`executeTemplate`
https://github.com/ampedandwired/html-webpack-plugin/blob/63e4ad6fc7d0fb11eb2c5813b88a22db973c0062/index.js#L197-L204

Could you please move this into the loader itself?
Do you know any way to run the tests for webpack1 and for webpack2?

foglerek:
I tried to find a way to move it into the loader, but the issue is that those variables are not available at the time the loader gets parsed. The loader gets parsed by Webpack if I understand it correctly, while these variables that you pass to the compiled template in executeTemplate become available at a later stage to the plugin itself.

Besides, don't the variables in executeTemplate need to be there in case another loader is used? (not the one provided by the plugin).

As for running the tests for webpack1 and webpack2, that's a bit tricky. I had a look at this: https://www.google.com/#q=run+tests+for+different+library+versions+npm

There's one way where you create a subpackage with a dependency for webpack2 which you can export as webpack. But you'd need to run all the tests twice, probably.

jantimon:
Moved on a branch (#225)

PR #233, opened 2016-02-20
Update README.md

PredokMiF:
Add hack to insert only needed links to the template, when we have multiple webpack entry points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,19 @@ This will generate a file `dist/index.html` containing the following:
 ```
 
 If you have multiple webpack entry points, they will all be included with `script`
-tags in the generated HTML.
+tags in the generated HTML. But you can create custom template, set `inject: false` and insert scripts and styles any you want:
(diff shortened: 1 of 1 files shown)


jantimon:
Why don't you use the entry point option?

PredokMiF:
Hi! "If you have multiple webpack entry points, they will all be included with `script`" it is from documentation. So if I have mono/multiple entry points - I have one result HTML file, with all scripts and styles (ok, not one, but with equal scripts list). Perhaps, I dont know something about webpack entry points?
Actually, I dont use webpack before and dont know it alot

jantimon:
Sorry wasn't explicit enough. The webpack plugin already offers a filtering option:
https://github.com/ampedandwired/html-webpack-plugin#filtering-chunks

PredokMiF:
Yes, this is exactly that I need! Thanks alot!

PR #235, opened 2016-02-23
Add chunk hash to `assets.chunks`

mc-zone:
Add chunk hash to `assets.chunks` so can use it in template.
For some version control requirements, chunk hash may be more useful than compilation hash.
For example, if there are two chunks in page, but only one of them is changed, use chunk hash is more cache-friendly.


mc-zone:
CI Checks failed:
```
Failures:
  1) HtmlWebpackPlugin fires the html-webpack-plugin-before-html-processing event
   Message:
     timeout: timed out after 5000 msec waiting for spec to complete
   Stacktrace:
     undefined
```
Timeout? Weird. I have passed all tests in local.

PR #240, opened 2016-02-27
Add before-html-generation event

SimenB:
This allows you to add assets to the html when not using injection


jantimon:
Shouldn't we better move the `html-webpack-plugin-before-html-processing` event up so it can deal with both ways?

SimenB:
Then you have no way to modify the HTML before injection

PR #241, opened 2016-02-28
Upgrade jasmine to 2

SimenB:
Focused specs make the upgrade worth it, IMO

### DIFF
--- a/spec/support/jasmine.json
+++ b/spec/support/jasmine.json
@@ -1,0 +1,8 @@
+{
+  "spec_dir": "spec",
+  "spec_files": [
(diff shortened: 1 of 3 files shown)


spec/support/jasmine.json
```diff
@@ -0,0 +1,8 @@
+{
+  "spec_dir": "spec",
+  "spec_files": [
+    "**/*[sS]pec.js"
+  ],
+  "stopSpecOnExpectationFailure": false,
+  "random": true
```
SimenB: only change from default generated file. In addition to removing `helpers`

Issue #244, opened 2016-03-01
Only generate HTML once


SimenB:
I've noticed that if I exclude `html-webpack-plugin` from tests, the re-compile time is much lower. And while tests don't need the html at all, I need them for regular development.

So I'm wondering if it's possible to generate the HTML only once, and not on rebuilds?

SimenB:
I was able to hack it together, and recompiles went from 10-12 seconds to less than 7. (we have a very slow recompile, working on it!)

```patch
 index.js | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)

diff --git a/index.js b/index.js
index 7714d4e..70a8d19 100644
--- a/index.js
+++ b/index.js
@@ -23,7 +23,8 @@ function HtmlWebpackPlugin (options) {
     showErrors: true,
     chunks: 'all',
     excludeChunks: [],
-    title: 'Webpack App'
+    title: 'Webpack App',
+    noRecompiles: false
   }, options);
 }
 
@@ -35,6 +36,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   this.options.template = this.getFullTemplatePath(this.options.template, compiler.context);
 
   compiler.plugin('make', function (compilation, callback) {
+    if (self.options.noRecompiles && self.compiled) return callback();
     // Compile the template (queued)
     compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, self.options.filename, compilation)
       .catch(function (err) {
@@ -53,6 +55,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   });
 
   compiler.plugin('emit', function (compilation, callback) {
+    if (self.options.noRecompiles && self.compiled) return callback();
     var applyPluginsAsyncWaterfall = Promise.promisify(compilation.applyPluginsAsyncWaterfall, {context: compilation});
     // Get all chunks
     var chunks = self.filterChunks(compilation.getStats().toJson(), self.options.chunks, self.options.excludeChunks);
@@ -160,6 +163,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       })
       // Let webpack continue with it
       .finally(function () {
+        self.compiled = true;
         callback();
         // Tell blue bird that we don't want to wait for callback.
         // Fixes "Warning: a promise was created in a handler but none were returned from it"
2.6.2
```

SimenB:
See linked PR

jantimon:
[...]

PR #245, opened 2016-03-01
Add option to skip recompiles

SimenB:
I'll just create a PR from #244 in case you're interested.
I'm unsure how to test this, doesn't seem like any tests runs the compilation twice


SimenB:
@jantimon Thoughts? I think that `cache` is not sufficient for my use case (I know the name s of all files, so I only need the html output once), and `cache` tries to be smart and see if there are any new assets. When I as a user know that there's not, I can apply this optimization.

jantimon:
Would prefer to fix the caching instead of implementing a work a round

PR #249, opened 2016-03-09
build public path with correct hash value

sandeep45:
Currently if my webpack config has: `publicPath: "/dist/[hash]/"` then my output html file will have:  `<script src="/dist/[hash]/commons_11e1b2480d26bdc6c179.js">`

WIth this fix [hash] gets replaced with the actual hash value.


jantimon:
Could you please add a unit test?

sandeep45:
@jantimon I have added a unit test. Please review.

PR #254, opened 2016-03-15
Update changelog

mistadikay:
fixes https://github.com/ampedandwired/html-webpack-plugin/issues/253


jantimon:
Wow - thank you so much!

PR #255, opened 2016-03-15
Add xhtml option to make link tags self-closing

koenpunt:
Some apps require xhtml tags to be generated (like here https://github.com/webpack/extract-text-webpack-plugin/issues/160)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Allowed values are as follows:
 - `chunks`: Allows you to add only some chunks (e.g. only the unit-test chunk)
 - `chunksSortMode`: Allows to control how chunks should be sorted before they are included to the html. Allowed values: 'none' | 'auto' | 'dependency' | {function} - default: 'auto'
 - `excludeChunks`: Allows you to skip some chunks (e.g. don't add the unit-test chunk)
+- `xhtml`: `true | false` If `true` render the `link` tags as self-closing, XHTML compliant. Default is `false`
 
 Here's an example webpack config illustrating how to use these options:
 ```javascript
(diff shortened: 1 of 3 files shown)


jantimon:
Could you please add a note to the readme and a unit test?

koenpunt:
:rocket:

jantimon:
Just thinking - would `selfClosingTags` be better then just `xml` ?

koenpunt:
`xhtml` could more than just closing tags. Maybe eventually there are some required attributes on tags in xhtml, which are not necessary for just html

jantimon:
True but this wouldn't be backwards compatible if we extend xml later on.

koenpunt:
> this wouldn't be backwards compatible

In what way backwards compatible? 

> extend xml later on

You're planning on xml support? How does that work? And in terms of naming, then the package name would be incorrect too I guess.

jantimon:
You said you might want to extend the `xml` feature set for the `xml` option eventually in future.
This would probably break backwards compatibility for people who set `xml` to `true` to get the self closing tags.

koenpunt:
I didn't mention `xml`, only `xhtml`

jantimon:
Sorry for the confusion I meant xhtml.

koenpunt:
Setting `xhtml` to `true` would mean you expect the output to be XHTML compliant, so as long as that's what is does, there won't be any backwards incompatibility further down the line. 
`selfClosingTags` would imply there are other reason than being XHTML compliant for self-closing tags to be present.

jantimon:
Not complete sold but sounds okay

koenpunt:
Cool, could you release a new version of the package? Thanks!

PR #256, opened 2016-03-15
Inline

numical:
Hi - my first GitHub pull request so apologies for anything missing and thanks in advance for any advice.
I have added in-lining the CSS as <style> tags in the generated HTML as opposed to using ExtractTextPlugin to pull into a separate CSS file and <link>'ing it.  Not everyone's Best Practice, but good for minimising number of requests and getting quick page loads in specific cases (especially with async scripts).
The technique is the same as ExtractTextPlugin (thanks Sokra!) - using a loader.  See the updated README.md for more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,3 +253,52 @@ compilation.plugin('html-webpack-plugin-before-html-processing', function(htmlPl
   callback();
 });
 ```
+
+Inline CSS
+----------
+
+CSS can be inlined into the head element of the HTML by using the inline loader.
+Note that the plugin itself does not need any options set up for inlining:
+
+```javascript
+module: {
+  loaders: [
+    { test: /\.css$/, loader: HtmlWebpackPlugin.inline() }
+  ]           
+},
+plugins: [
+  new HtmlWebpackPlugin()
+]  
+```
+The inline loader can be the end of a chain of CSS loaders, though note that any prior loader must output pure CSS, not javascript.  This means that [css-loader](https://www.npmjs.com/package/css-loader) and [style-loader](https://www.npmjs.com/package/style-loader) cannot be used, though the very useful [post-css](https://www.npmjs.com/package/postcss-loader) loader can be:
+
+```javascript
+module: {
+  loaders: [
+    { test: /\.css$/, loader: HtmlWebpackPlugin.inline('postcss-loader') }
+  ]           
+},
+postcss: [
+  require('autoprefixer')
+],
+plugins: [
+  new HtmlWebpackPlugin()
+]  
+```
+
+You may also wish to minify the in-lined CSS; this requires specific `minify` options:
+
+```javascript
+module: {
+  loaders: [
+    { test: /\.css$/, loader: HtmlWebpackPlugin.inline() }
+  ]           
+},
+plugins: [
+  new HtmlWebpackPlugin({
+   minify: {
+    minifyCSS: true
+   }
+  })
+]  
+```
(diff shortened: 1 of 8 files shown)


jantimon:
Why don't you just render the css of the extract text plugin?

numical:
Yup - I actually did that in an early commit on my local repo.  But then I realised it was pig ugly.  For the user, they in effect have to tell Webpack to remove all CSS to an external css file and then extract it form there back into the HTML.  In practice it is ugly too, because having done this you then have to remove css assets and chunks from the Compilation.  Yuk.  My suggested approach is intended to parallel the 'ExtractTextPlugin' approach whilst being an obvious alternative.  Your call though :-)

jantimon:
Why would you use a store? That's an antipattern. Think of multiple webpack-plugins or multiple compilations. Your store would be full of old data.

jantimon:
Maybe you could move this pull request into it's own plugin - the html-webpack-plugin offers hooks to modify the html output afterwards.

numical:
Re: using a store.
I thought I could get away with a naive implementation of the store as each invocation of webpack would cause a new instantiation.  However it looks like I'm the one being naive - in hindsight it is obvious as something like the 'watch' state will cause multiple compilations from a single instantiation.  Not sure I follow the use case of 'multiple webpack plugins' though...
Anyhow, back to the drawing board there - how to pass the CSS to the plugin. I am not keen to add to the Compilation as:
1: Conceptually it seems wrong: my understanding of the Compilation's purpose is to produce javascript and javascript-wrapped assets.  Attaching CSS which will have nothing to do with any of the Compilation's assets seems to be mixing concerns.
2: And more importantly: I am terrified of modifying core Webpack entities as I really do not yet have a grasp of the possible Unforeseen Consequences..
Any suggestions most welcome :-)

re: a separate plugin
There's an idea.  I'll take a look.  But please note that I have tried hard to minimize changes to html-webpack-plugin code so as to make this Pull request as risk-free as possible.  Once I sort out a store alternative I hope it might make it into the plugin.  Better for the user a single plugin with more options than two separate plugins to keep in sync.

numical:
PS : thanks for the feedback.  Much appreciated.

jantimon:
I believe this plugin is to bloated already so I would prefer to focus on separation of concerns and let it just generate the basic html skeleton ;)
If you want to do more than that then you might use the hooks or write a custom template like 
https://github.com/jaketrent/html-webpack-template or https://github.com/SimenB/add-asset-html-webpack-plugin

numical:
Hi,
Happy to close this PR and move the in-lining to a separate plugin.
Would you be willing to accept a PR that moves any minify processing until after the 'html-webpack-plugin-after-html-processing' event?  I *think* that makes sense.
If you are worried about backward compatibility, I can think of two options: 1) another event, 2) an option on the minify option that defines when it happens.
Thoughts?

jantimon:
I guess I would add another event - why could minification be a problem?

jantimon:
Would love to see the plugin :+1:

numical:
Minification not a problem per se.  
However one of the nice things about this PR is that it worked with the current minify options; i.e. including:
```javascript
new HtmlWebpackPlugin({
  minify: { {
    minifyCSS: true
  }
});
```
would minify the inlined CSS as well.
Once the inlining is moved into a plugin fired by the 'html-webpack-plugin-after-html-processing' event, it is too late to use this option.  Obviously there are solutions to this, just none as clean as an new event firing between the 'after html processing' and before the call to html-minifier.
I'll get on with that plugin then.

Issue #121, opened 2015-11-21
Ability to include [hash] in filename


frenchie4111:
I want to have my index.html also include the project hash.

so
```
new HtmlWebpackPlugin( {
    filename: 'index.[hash].html'
} )
```

Would produce a file with the name something like this: `index.3017a9749fcdc8c694cd.html`

I could then pipe it into `gulp-cloudfront` to set it as the new root object on my cloudfront distribution.

PR #257, opened 2016-03-16
add support for [hash] in filename

dotch:
This PR adds support for replacing `[hash]` in the filename with the hash of the compiled template. 

This behavior was brought up in #121 and would be pretty helpful for me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,8 +43,12 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   }
 
   compiler.plugin('make', function (compilation, callback) {
+    // Remove occurences of '[hash]' in the filename so the child
+    // compiler does not try to replace it.
+    var escapedFilename = self.options.filename.replace('[hash]', '');
+
     // Compile the template (queued)
-    compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, self.options.filename, compilation)
+    compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, escapedFilename, compilation)
       .catch(function (err) {
         compilation.errors.push(prettyError(err, compiler.context).toString());
         return {
@@ -149,8 +153,9 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
         return self.options.showErrors ? prettyError(err, compiler.context).toHtml() : 'ERROR';
       })
       .then(function (html) {
+        var outputFilename = self.options.filename.replace('[hash]', self.childCompilerHash);
         // Replace the compilation result with the evaluated html code
-        compilation.assets[self.options.filename] = {
+        compilation.assets[outputFilename] = {
           source: function () {
             return html;
           },
@@ -158,11 +163,12 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
             return html.length;
           }
         };
+        return outputFilename;
       })
-      .then(function () {
+      .then(function (outputFilename) {
         // Let other plugins know that we are done:
         return applyPluginsAsyncWaterfall('html-webpack-plugin-after-emit', {
-          html: compilation.assets[self.options.filename],
+          html: compilation.assets[outputFilename],
           plugin: self
         });
       })
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -38,6 +38,14 @@ function testHtmlPlugin (webpackConfig, expectedResults, outputFile, done, expec
     } else {
       expect(compilationWarnings).toBe('');
     }
+    if (outputFile instanceof RegExp) {
+      var matches = Object.keys(stats.compilation.assets).filter(function (item) {
+        return outputFile.test(item);
+      });
+      expect(matches.length).toBe(1);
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -37,7 +37,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
 
   compiler.plugin('make', function (compilation, callback) {
     // Compile the template (queued)
-    compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, self.options.filename, compilation)
+    compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, 'template', compilation)
```
dotch: The name was replaced here, because the child compiler would otherwise replace the `[hash]` by itself, causing an error. The actual provided output name for this function seems to not matter and be a temporary thing anyway. Please correct me if my assumption is wrong.

jantimon:
Is this different from #249?

dotch:
Yes, this PR targets the output html file, #249 targets the included assets. 
An example usage for this PR would be:
``` JS
new HtmlWebpackPlugin({
      template: 'index.html.template',
      filename: 'build/index-[hash].html',
}),
```
`[hash]` would then be replaced with the hash of the compiled template. 
This is useful in cases where the final output html is by itself only an asset that also needs cache busting. E.g. when using it as an iFrame or with html-imports.

index.js
jantimon: Webpack will overwrite an existing file with the name 'template'. This could also conflict if you would have two instances of the webpack-html-plugin.

index.js
```diff
@@ -48,6 +48,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
         // If the compilation change didnt change the cache is valid
         isCompilationCached = compilationResult.hash && self.childCompilerHash === compilationResult.hash;
         self.childCompilerHash = compilationResult.hash;
+        self.options.filename = self.options.filename.replace(/\[hash\]/g, compilationResult.hash);
```
jantimon: Do we have to change the user options here?

dotch: We could store the filename in a different property such as `this.modifiedFilename` and leave the option as is for clarity.

jantimon: Can't we just use `self.childCompilerHash = compilationResult.hash;` ?

dotch:
I added a commit using `self.childCompilerHash` to build the filename at the position where it gets used.

index.js
dotch: Hm I guess that makes my approach rather pointless. We can not hand in the filename with the hash before we even have compiled the template.

alextreppass:
I ran into a similar issue; the easiest way for me was to:

- 1) [Get webpack to do the hashing](https://webpack.github.io/docs/long-term-caching.html)
- 2) Remove `hash: true` from HtmlWebpackPlugin config options

jantimon:
Looks quite good - however I am still not sure about the hardcoded `template` name.
It might cause issues with recompilation and caching.
Could we instead just escape the template name?

Maybe you could also add a test to the caching spec.

dotch:
Agree with both of your points and will address them and document the functionality in the readme.

dotch:
@jantimon Mind having another look at this?
I escaped the filename before handing it to the child compiler, by just removing `[hash]`.
I added a test for correct handling of `[hash]` in the output file name.
And I run the whole caching spec suite with a `[hash]` in the filename.
Sorry for the massive diff due to the indentation change in `CachingSpec.js`.

krukid:
@dotch This PR is exactly what I need for an IFrame, but one thing is not clear to me: when the plugin generates a hash-based file name, how do I get it programmatically? I mean ideally, I could just `require(file_name)` (or template_name?) and get the compiled HTML asset URL, but something tells me that is not how it works. Am I wrong?

jantimon:
Could you please revert the indentation change - it's really impossible to read otherwise.

dotch:
@jantimon 
Sorry about that. I reverted the whitespace changes and put them in a second commit for now.

jantimon:
Looks like it is still broken

index.js
```diff
@@ -43,8 +43,12 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   }
 
   compiler.plugin('make', function (compilation, callback) {
+    // Remove occurences of '[hash]' in the filename so the child
+    // compiler does not try to replace it.
+    var escapedFilename = self.options.filename.replace('[hash]', '');
```
jantimon: Wouldnt it be better to replace `[` and `]` to reduce the chance of collisions and support also other webpack variables?

We could move the escaping to a small file in the library folder which does only the webpack filename escaping and unescaping.

What about `[` -> `{[{` and `]` -> `}]}` ?

Should still be a valid filename: https://en.wikipedia.org/wiki/Filename

jantimon:
After some research I tried to use the webpack way of handling the filename: #279.
What do you think?

dotch:
Looks good to me, I like that approach!

jantimon:
Moved to #279

PR #259, opened 2016-03-16
Async attribute for <script> tags

numical:
You probably have solutions for this as the functional change is trivial.
However this is my go at a convenient options syntax to permit flexibility in defining *which* scripts are marked as 'async'.


numical:
Now implemented as extension plugin [script-ext-html-webpack-plugin](https://github.com/numical/script-ext-html-webpack-plugin)

PR #263, opened 2016-03-20
Add caching tests

jantimon:
Tests to ensure that the html is compiled only once (see #244)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "style-loader": "^0.13.0",
     "underscore-template-loader": "^0.7.2",
     "url-loader": "^0.5.7",
-    "webpack": "^1.12.14"
+    "webpack": "^1.12.14",
+    "webpack-recompilation-simulator": "^1.3.0"
   },
   "dependencies": {
     "bluebird": "^3.3.4",
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -1,3 +1,7 @@
+/*
+ * Integration and unit tests for all features but caching
+ */
+
 /* eslint-env jasmine */
 'use strict';
 
--- a/spec/CachingSpec.js
+++ b/spec/CachingSpec.js
@@ -1,0 +1,109 @@
+/*
+ * Integration tests for caching
+ */
+
+/* eslint-env jasmine */
+'use strict';
+
+// Polyfill promisses for node 0.10.x
+if (!global.Promise) {
+  require('es6-promise').polyfill();
+}
+
+var path = require('path');
+var webpack = require('webpack');
+var rm_rf = require('rimraf');
+var WebpackRecompilationSimulator = require('webpack-recompilation-simulator');
+var HtmlWebpackPlugin = require('../index.js');
+
+var OUTPUT_DIR = path.join(__dirname, '../dist');
+
+function setUpCompiler (htmlWebpackPlugin) {
+  spyOn(htmlWebpackPlugin, 'evaluateCompilationResult').and.callThrough();
+  var compiler = new WebpackRecompilationSimulator(webpack({
+    entry: path.join(__dirname, 'fixtures/index.js'),
+    output: {
+      path: OUTPUT_DIR,
+      filename: 'index_bundle.js'
+    },
+    plugins: [htmlWebpackPlugin]
+  }));
+  return compiler;
+}
+
+describe('HtmlWebpackPluginCaching', function () {
+  beforeEach(function (done) {
+    rm_rf(OUTPUT_DIR, done);
+  });
+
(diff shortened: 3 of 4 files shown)


SimenB:
Not able to create test, but I made a small sample repo showing the issue.

https://github.com/SimenB/webpack-html-double-compile

image

Just run `npm start`, then change the `index.js` file.

jantimon:
Sorry I don't understand what you mean exactly - the webpack dev server doesn't write to disk.

SimenB:
That's why I said "disk".

Running webpack in watch-mode shows the same. Updated with `npm run watch`.

jantimon:
Is this a child compilation issue or a webpack-html-plugin issue?

SimenB:
I don't know...

But I just found out that it's the `make` phase that messes us up, so the cache option does its job.

```diff
diff --git i/index.js w/index.js
index ab7a5f0..563d6d3 100644
--- i/index.js
+++ w/index.js
@@ -32,10 +32,14 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   var self = this;
   var isCompilationCached = false;
   var compilationPromise;
+  var compiled = false;

   this.options.template = this.getFullTemplatePath(this.options.template, compiler.context);

   compiler.plugin('make', function (compilation, callback) {
+    if (compiled) return callback();
+    compiled = true;
+
     // Compile the template (queued)
     compilationPromise = childCompiler.compileTemplate(self.options.template, compiler.context, self.options.filename, compilation)
       .catch(function (err) {
```

image

jantimon:
So what can we do now?
The childCompilation also uses the cache (probably) as you can see in this test: https://github.com/ampedandwired/html-webpack-plugin/blob/feature/recompilation-tests/spec/CachingSpec.js#L78-L80

SimenB:
I think that test only asserts that the output remains the same, not that it's only output once.

It might be that the child compiler always adds the feil to the compilation, and even though it's unchanged, webpack "overwrites" the old file?

What about introducing a new option (similar to my last PR, but only for the `make` phase), to not recompile the template?I don't know how often the template is changed (I'm guessing it's very infrequent), but I think that's a fair solution.

jantimon:
I added another test which counts the build files during the built:

```js
function getCompiledModuleCount (statsJson) {
  return statsJson.modules.filter(function (webpackModule) {
    return webpackModule.built;
  }).length + statsJson.children.reduce(function (sum, childCompilationStats) {
    return sum + getCompiledModuleCount(childCompilationStats);
  }, 0);
}
```

It shows that the child compilation is cached:

```js
  it('should compile nothing if no file was changed', function (done) {
    var template = path.join(__dirname, 'fixtures/plain.html');
    var htmlWebpackPlugin = new HtmlWebpackPlugin({
      template: template
    });
    var childCompilerHash;
    var compiler = setUpCompiler(htmlWebpackPlugin);
    compiler.run()
      // Change the template file and compile again
      .then(function () {
        childCompilerHash = htmlWebpackPlugin.childCompilerHash;
        return compiler.run();
      })
      .then(function (stats) {
        // Verify that no file was built
        expect(getCompiledModuleCount(stats.toJson()))
          .toBe(0); // <---- the child compilation compiled 0 files
      })
      .then(done);
   });
});

SimenB:
What about instead of creating a childcompiler on every change webpack sees, to just watch the passed in template file yourself? Using e.g. https://github.com/webpack/watchpack

jantimon:
This wouldn't work with partials - where exactly do we lose performance and how much? 
Because it seems like everything comes from the cache.

jantimon:
I guess we should ask @sokra for the intended webpack caching way instead of trying to find workarounds. Maybe also here: https://hashnode.com/ama/with-webpack-team-cilqks5re00038g53qqgpvqza#cils0m1oq000rkb53sgfg2egr

SimenB:
I'm unsure about how much actual perf is lost. The huge slowdown for us was using 2.1.0 of this plugin, it's much better with newer versions with caching enabled.

Still, there's no need to create a child compiler unless the template itself changes, and then we avoid extra writes as well, so that might be an optimization we could do?

jantimon:
If there are no performance issues we shouldn't try to fix them ;)
Performance optimisations are always bad for maintainability..

SimenB:
@jantimon simple change in js file compiles with 2.5-2.8 seconds on my machine when skipping extra trips with html-webpack-plugin. With the added output of 4 html files, it's a second more, between 3.4-3.6. This is over a course of 10 builds with and without a change.

And yes, it's much better than it was, but I'd still like to have the option to skip recompiles. Thoughts?

jantimon:
Could you try to narrow it down which part is causing the peformance leak? 
Maybe by using `console.time`. It looks like this is an webpack issue not a plugin issue.

Issue #265, opened 2016-03-23
Absolute path for `filename` option?


yyx990803:
I am trying to generate the HTML file at a different location from other webpack built assets. However if I specify an absolute path for the `filename` option, the generated HTML file is still nested under webpack's output path. I assume this is due to how webpack interprets assets, but is there anyway to get around it?

jantimon:
duplicate of #234

yyx990803:
That's not quite what I want - I want to generate the HTML to a location unrelated to the webpack output path. Also, specifying a nested path in `output.filename` is a less-than-optimal solution, because then I also have to do so for all other assets such as files processed by `file-loader` and CSS extracted with `extract-text-webpack-plugin`.

It also feels "not working as expected" when you specify an absolute path and instead see it nested under another directory. In fact the fix is pretty simple, as implemented in this PR: https://github.com/ampedandwired/html-webpack-plugin/pull/266

yussan:
i think i found the solution , i need absolute path too, because i have to deploy in netlify and combine via _redirects file.

my solution : 
in webpack config. set `output.publicPath` to `/`
```
...
output: {
    filename: "app/[name]-[hash].js",
    path: "dist",
    publicPath: "/"
  },
...
```
ref : https://stackoverflow.com/questions/36122972/preserving-folder-structure-in-output-directory-with-webpack

PR #266, opened 2016-03-23
support absolute filenames

yyx990803:
solves #265

### DIFF
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -571,6 +571,19 @@ describe('HtmlWebpackPlugin', function () {
     }, ['<script src="index_bundle.js"'], 'test.html', done);
   });
 
+  it('allows you to use an absolute output filename', function (done) {
+    testHtmlPlugin({
+      entry: path.join(__dirname, 'fixtures/index.js'),
+      output: {
+        path: OUTPUT_DIR,
+        filename: 'index_bundle.js'
+      },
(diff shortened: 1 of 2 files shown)


spec/BasicSpec.js
```diff
@@ -571,6 +571,19 @@ describe('HtmlWebpackPlugin', function () {
...
 
+  it('allows you to use an absolute output filename', function (done) {
+    testHtmlPlugin({
+      entry: path.join(__dirname, 'fixtures/index.js'),
+      output: {
+        path: OUTPUT_DIR,
+        filename: 'index_bundle.js'
+      },
+      plugins: [new HtmlWebpackPlugin({
+        filename: path.resolve(OUTPUT_DIR, 'test.html')
```
jantimon: Would this also work for `path.resolve(OUTPUT_DIR, 'subfolder', 'test.html')`?

jantimon:
Looks very promissing - I'll look into it.

spec/BasicSpec.js
yyx990803: yes, it should work with arbitrary absolute paths.

jantimon:
Had to fix one thing but looks good:
https://github.com/ampedandwired/html-webpack-plugin/tree/yyx990803-master

jantimon:
I am on the road right now and my internet is quite bad - Probably won't be able to publish this before tonight.

PR #274, opened 2016-03-29
Keep favicon.ico at Root of Project

Schlesiger:
I have an issue where the favicon.ico is missing in subdirectories. Since this plugin only copies the favicon.ico to the root of the project, the path should be absolute.


jantimon:
I see your point but this will break a lot for people who are already using the plugin.
The plugin copies the favicon to the same location as the index.html.

PR #281, opened 2016-04-02
Don’t emit the html for hot-update compilation.

dtinth:
webpack-dev-server@2.0.0-beta seems to be emitting hot update bundles separately from the full bundle. This makes HtmlWebpackPlugin mistake it for the real bundle, and generates a HTML file that looks like this:

```html
<script src="0.8f9c33574d86dd9eab0d.hot-update.js"></scriptt>
```

This commit uses quite a hack to check if the emitted bundle is a hot-update bundle, and skip emitting `index.html` file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,6 +68,12 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
     chunks = self.sortChunks(chunks, self.options.chunksSortMode);
     // Get assets
     var assets = self.htmlWebpackPluginAssets(compilation, chunks);
+    // If this is a hot update compilation, move on!
+    // This solves a problem where an `index.html` file is generated for hot-update js files
+    // It only happens in Webpack 2, where hot updates are emitted separately before the full bundle
+    if (self.isHotUpdateCompilation(assets)) {
+      return callback();
+    }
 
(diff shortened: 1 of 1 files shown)


jantimon:
Cool would love to merge this - could you please add a test? Should be possible with the caching test spec.

MoOx:
Not sure how to test this but here is my scenario:

- start my app with --hot and --inline
- rebuild a file that have a change that is not handled by the HMR (eg: a very simple change in your entry point)
- webpack refresh the full page, but this plugin is serving a hot-update instead of the expected chunks

This PR definitely solved the issue for me, hope to get it merged :)

MoOx:
This is related to the fact that webpack 2 now use a [2 passes](https://github.com/webpack/webpack/pull/856) build step for hot loaded stuff

> Child html-webpack-plugin for "index.html":
>    chunk    {0} index.html 505 kB
> **Compilation needs an additional pass and will compile again.**
> webpack: bundle is now VALID.

Here is a raw debug where you can see the chunks details

screen shot 2016-04-27 at 13 07 59

Hope that helps to understand the problem

jantimon:
I will try to write a test with https://github.com/jantimon/webpack-recompilation-simulator

PR #282, opened 2016-04-04
Issue with publicPath and custom template

kos9k:
Hello. Please review this bug fix.

## Reason:

I have custom html template which proceed though my loader like html-loader (this bug was replicated with html-loader also)
my template have this line
```
<link rel="icon" href="/images/favicon.ico" type="image/x-icon" />
```
(way with favicon option does not suit me)

href attr converted to [my_dist_directory_with_hash]/images/favicon.ico
my publicPath option for webpack is '/dist/somepath/[hash]/'

after html was emit, attr converted to 

[html_plugin_child_compiler_webpack_publicPath]/images/favicon.ico
html-webpack-plugin

not my publicPath;
I was found only this way for fix this issue)

P.S. Sorry for my English...)))

### DIFF
--- a/index.js
+++ b/index.js
@@ -189,7 +189,7 @@ HtmlWebpackPlugin.prototype.evaluateCompilationResult = function (compilation, s
   // The LibraryTemplatePlugin stores the template result in a local variable.
   // To extract the result during the evaluation this part has to be removed.
   source = source.replace('var HTML_WEBPACK_PLUGIN_RESULT =', '')
-    .replace(/(__webpack_require__\.p\s=\s\")[\s\S]*/, ['$1', compilation.mainTemplate.getPublicPath({hash: compilation.hash }), '"'].join(''));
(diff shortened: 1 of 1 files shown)


jantimon:
Why can't you use the url-loader or file-loader instead?

kos9k:
I have 2 way to compile my project (new with webpack and old) and i need support old compilation procedure. I use file-loader for static files, this described in webpack.config; 

```
{
    test: /\.(ico)$/,
    loader: 'file?name=[path][name].[ext]'
}
```

After emit my files located in my dist directory with hash like
```
dist/[hash]/static_files
mybuild.[hash].js
```

jantimon:
Use the file loader in your html: (won't work with the html-loader)

`<link rel="icon" href="<%= require('./images/favicon.ico') %>" type="image/x-icon" />`

kos9k:
This is not suit for me(((.

jantimon:
Why?

kos9k:
Because i use my own loader (like html-loader) for file.tpl file (like file.html). In this file all static request transforms in require('static/file'), It is the same as you advised me.

PR #295, opened 2016-04-18
add demo code about events

zzuieliyaoli:
When I want to use `events` to alter htmls, the usage of `events` can't help me directly.

I spent much time on finding that what's the `compilation`. That's so sad.


jantimon:
Thanks - however the additional `when` doesn't make much sense to me

numical:
More detailed code example +1
Additional 'when' -1
Also make reference to working example such as, ahem, [style-ext-html-webpack-plugin](https://github.com/numical/style-ext-html-webpack-plugin) ... :-)

jantimon:
I also added a link to https://github.com/jantimon/html-webpack-harddisk-plugin which shows a very simple implementation.

PR #303, opened 2016-04-23
Add dns-prefetch option to generated style element

seanrkerr:
HI, Thought this might be a good edition to your plug-in. It has great browser support and can boost performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Allowed values are as follows:
 - `chunksSortMode`: Allows to control how chunks should be sorted before they are included to the html. Allowed values: 'none' | 'auto' | 'dependency' | {function} - default: 'auto'
 - `excludeChunks`: Allows you to skip some chunks (e.g. don't add the unit-test chunk)
 - `xhtml`: `true | false` If `true` render the `link` tags as self-closing, XHTML compliant. Default is `false`
+- `preFetch`: 'true' Gives a hint to the browser to perform a DNS lookup in the background to improve performance if false then 'stylesheet' attribute is added by default.
 
 Here's an example webpack config illustrating how to use these options:
 ```javascript
--- a/index.js
+++ b/index.js
@@ -24,7 +24,8 @@ function HtmlWebpackPlugin (options) {
     chunks: 'all',
     excludeChunks: [],
     title: 'Webpack App',
-    xhtml: false
+    xhtml: false,
+    preFetch: false
   }, options);
 }
 
@@ -436,6 +437,14 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
  * Injects the assets into the given html string
  */
 HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets) {
+  //check to see if we have the prefetch option in our options object
+  var prefetch = this.options.preFetch;
+  //this is the defay
+  var strCSSAttribute = 'stylesheet';
+  //if the prefetch option is set to true then add it to the stylesheet element
+  if (prefetch) {
+    strCSSAttribute = 'dns-prefetch';
+  }
   // Turn script files into script tags
   var scripts = assets.js.map(function (scriptPath) {
     return '<script src="' + scriptPath + '"></script>';
@@ -444,7 +453,7 @@ HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets) {
   var xhtml = this.options.xhtml ? '/' : '';
   // Turn css files into link tags
   var styles = assets.css.map(function (stylePath) {
-    return '<link href="' + stylePath + '" rel="stylesheet"' + xhtml + '>';
+    return '<link href="' + stylePath + '" rel="'+ strCSSAttribute + xhtml + '>';
   });
   // Injections
   var htmlRegExp = /(<html[^>]*>)/i;
--- a/index.js
+++ b/index.js
@@ -454,9 +454,8 @@ HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets) {
 
   // Turn css files into link tags
   var styles = assets.css.map(function (stylePath) {
-    return '<link href="' + stylePath + '" rel="' + strCSSAttribute + '" ' + xhtml + '>';
+    return '<link href="' + stylePath + '" rel="' + strCSSAttribute + '"' + xhtml + '>';
   });
-  console.log(styles);
   // Injections
   var htmlRegExp = /(<html[^>]*>)/i;
   var head = [];
--- a/index.js
+++ b/index.js
@@ -437,11 +437,11 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
  * Injects the assets into the given html string
  */
 HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets) {
-  //check to see if we have the prefetch option in our options object
+  // check to see if we have the prefetch option in our options object
   var prefetch = this.options.preFetch;
-  //this is the defay
+  // this is the defay
   var strCSSAttribute = 'stylesheet';
-  //if the prefetch option is set to true then add it to the stylesheet element
+  // if the prefetch option is set to true then add it to the stylesheet element
   if (prefetch) {
     strCSSAttribute = 'dns-prefetch';
   }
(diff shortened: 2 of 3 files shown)


jantimon:
But what has prefetch to do with the styling?
Could you give an example on how this might work?

seanrkerr:
@jantimon dns-prefetch allows the pre resolution of resources, in this case, stylesheets. In terms of the article you sent, for which, I didn't read all of it. I try to follow standards and implement changes that have support for more than just one browser. It's entirely up to you to implement my change. I appreciate your feedback though.

jantimon:
But prefetching a style sheet doesn't make any sense to me - the browser has to fetch the dns anyway..

seanrkerr:
Thats right the browser does have to fetch the DNS. Consider the following example:

 <link rel="stylesheet" type="text/css" href="/sites/css">
 <link rel="stylesheet" type="text/css" href="www.mydomain.com/another.css">
 <link rel="stylesheet" type="text/css" href="www.anotherdomain.com/yetanother.css">

dns-prefetching notifies the client that there are assets we'll need later from a specific URL, so the browser can resolve the DNS as quickly as possible. In this case, it will resolve the additional css files that we have on our page, which come from different domains ahead of time. 

Have a look at the specification:
[spec](https://www.w3.org/TR/resource-hints/#dns-prefetch)

jantimon:
If I got you correctly you don't want to add the stylesheets in the head but load them later on using javascript. In this case your dns prefetch does make sense.
But as this is a really rare use case which can already be achieved by turning the inject feature off together with a custom template like https://github.com/jaketrent/html-webpack-template/blob/master/index.ejs this pull request won't be merged.

seanrkerr:
You completely missed the point. That's ok. Thanks for your time.

jantimon:
Okay sorry then please try to explain it once again.
When should the resolving take place and when should the download take place? Both on page load?

seanrkerr:
How about I do some performance testing? Probably better to prove that it works rather than reading the specs. I'll do this if you re-open my pull. It's only fair.

jantimon:
does this mean you didn't test this pr and you don't know if it works at all?

seanrkerr:
no it means that for some reason you're skeptical about it's implementation. There is evidence of this technique all over the place. 

I don't understand why this is all so hard. I can't be the only person that has heard of this and other techniques....

I'll create some tests and post some findings. If you don't like the results, you can close the pull and thats that.

numical:
To be fair,  and notwithstanding the implementation, this PR does not follow the coding standards of the project and has no additional tests for the added functionality.  These should at least be addressed before getting onto the conceptual stuff of what `prefetch` is about.

jantimon:
Okay as it reads from the spec dns prefetching is for dns resolution of foreign files e.g. javascript, images, stylesheets especially when using CDNs or external links.

However this only useful if they are not loaded during the page load.
Resources load during the page load are already discovered by modern browsers.

Your change would turn all 

` '<link href="' + stylePath + '" rel="stylesheet"' + xhtml + '>'` 

into

 `'<link href="' + stylePath + '" rel="dns-prefetch"' + xhtml + '>'`

So it would **remove all stylesheet** entries from the page for a performance gain which doesn't exist.

Please correct me if anything of this is wrong.

seanrkerr:
@numical where do I find the guidelines you mention?

jantimon:
Semistandard - to test on your machine that you are compliant run `npm test`

Have you alread done some tests on the change?

seanrkerr:
ah ok. I was looking at the wiki for them. Might help if some where created.

seanrkerr:
Will get back to you with the test and make some changes after executing the test runner.

seanrkerr:
Hey Jan, got a few errors when running the tests. Will investigate and post a work pull.

jantimon:
Looks like the webpack2 tests fail

seanrkerr:
Compatibility issue?

seanrkerr:
wait a sec. I think I can see the issue...

jantimon:
You have to install webpack2 and execute `npm run build-examples`

seanrkerr:
they should be passing now... I'll need to update them anyway to test the new attribute.

seanrkerr:
hmm. it failed with a peer dependency issue.

seanrkerr:
apologies, just saw the comment / realised deps have changed too.

seanrkerr:
@jantimon I'll add some tests for the new attribute. Then if that's ok I post some benchmarks.

jantimon:
And please verify that the stylesheets still work ;)

seanrkerr:
will do.

seanrkerr:
@jantimon Did some testing, it does seem to make a slight difference with the attribute. Will modify the specs to include tests for this param + ensure the stylesheets work.

prefetch
no-prefetch

jantimon:
Could you please share the entire test sources and results - and which gains you could find for the various browsers?

seanrkerr:
Do you have a definitive list of browsers you target? This attribute is supported by all of the current browsers, including IE.

jantimon:
Can I close this?

seanrkerr:
You've taken so long to approve this, now it has conflicts with the build.

jantimon:
Sorry I was waiting for the test sources and results..

seanrkerr:
And I was wait for device and browser versions. 

You make it really difficult to contribute. Whats the issue?

jantimon:
This project has over 250.000 downloads a month and you want to change it's core without being able to explain the benefits in detail and removing `rel="stylesheet"` from all stylesheets might cause a lot of harm.. So I won't merge this until I am convinced that this is no issue for all the browsers out there.
The other reason is that I don't believe there is any advantage in adding prefetch as this resource is not lazy loaded.

It's nothing personal! And I would be happy to learn something new if you could proof me wrong. 
I have prepared the code for your own npm webpack plugin which would do the same thing as your proposed change (using the html webpack plugin events).

This would allow you and others to test and use 'dns-prefetch' in their projects.

```js
function YourPlugin() {

}

YourPlugin.prototype.apply = function apply(compiler) {

  function filterTag(tag) {
    return tag.tagName === 'link';
  }

  function processTag(tag) {
    tag.rel = 'dns-prefetch';
  }

  compiler.plugin('compilation', function compilationPlugin(compilation) {
    compilation.plugin('html-webpack-plugin-alter-asset-tags', function supportHtmlWebpack(pluginArgs, callback) {
      pluginArgs.head.filter(filterTag).forEach(processTag);
      pluginArgs.body.filter(filterTag).forEach(processTag);
      callback(null);
    });
  });
};

module.exports = YourPlugin;
```

Usage in your config:
```js
{
    plugins: [
        new HtmlWebpackPlugin(),
        new YourPlugin(),
    ],
}
```

Hopefully you don't feel offended - I am just trying to do the right thing for everyone using the plugin ;)

seanrkerr:
Hi, no offence taken. It's good to be cautious. In terms of testing and browser support it's pretty much supported across the board as a standard. 

1. The screen shots above illustrate the performance gain
2. It has nothing to do with lazy loading, it's telling the browser to prefetch the resource in parallel 

Without understanding the browser support matrix you have, I wouldn't have been able to provide benchmarks away.

Cheers, 

Sean

jantimon:
From Wikipedia:

> The Domain Name System (DNS) is a hierarchical decentralized naming system for computers, services, or any resource connected to the Internet or a private network. It associates various information with domain names assigned to each of the participating entities. Most prominently, it translates more readily memorized domain names to the numerical IP addresses needed for the purpose of locating and identifying computer services and devices with the underlying network protocols. By providing a worldwide, distributed directory service, the Domain Name System is an essential component of the functionality of the Internet.

What is the aim of DNS prefetching?

from https://www.chromium.org/developers/design-documents/dns-prefetching :

> DNS prefetching is an attempt to resolve domain names before a user tries to follow a link. 

So all it does is turning a domaing e.g. `google.com` into `216.58.208.46`. It does not download any content.

This also means, that this feature is not testable on localhost as localhost is resolved using your etc hosts file.

Do you think you will publish a dns-prefetch-webpack plugin with the code I prepared?

seanrkerr:
Yep, you posted that stuff before and I know what DNS is.

> Pre-fetching is a way of hinting to the browser about resources that are definitely going to or might be used in the future, some hints apply to the current page, others to possible future pages.
> 
> As developers, we know our applications better than the browser does. We can use this information to inform the browser about core resources.

> This works for css, images, fonts, assets hosted on other servers / CDN's

1. In terms of Chromium, this is not the only browser that people are using. There are many others, the docs, that you sent through are vague at best. You have not even refuted my claims by substantiated, just copy and pasted from the docs.  

>You: This also means, that this feature is not testable on localhost as localhost is resolved using your etc hosts file.

>You: Why would u test on localhost? 
Maybe this is something you, yourself would do but not me.

The code I published at the time was fine, it was tested, and passing, you asked me for test results

At some point you'll learn about this stuff (dns-prefetch, preconnect, and prerender) and maybe even use it.
>> You: Do you think you will publish a dns-prefetch-webpack plugin with the code I prepared?

1. I would assist with tests and look at what it does
2. I would supply a list of target browsers, versions both major and minor

It would be optional, a boolean. Of course I would test it out first.

No need to post anymore Chromium excerpts, I've read them already.

jantimon:
Cool please send me a link to your github repo of the plugin and please let me know when you published it with some tests 👍

seanrkerr:
Of course @jantimon. I have no problems doing this. I see an opportunity. Any contributions by yourself would need to fit in with the guidelines.... ✋

jantimon:
Cool if you need further support to write your plugin for html-webpack-plugin just let me know but I guess the example from above should get you started :)

seanrkerr:
thanks, much appreciated. I think I could write one pretty easily. Would be a good little project. 👍

PR #311, opened 2016-04-28
Add type attribute to script element

### LINKED ISSUE #309: Safari 9.1.1. Issue - "A problem repeatedly occurred"
Earlier this month I noticed an issue with some websites I built HTML for using this plugin: when landed in Safari, the page is reloaded several times and finally, shows "A problem repeatedly occurred." 

screen shot 2016-04-27 at 10 32 14 am

Initially, I thought it was an issue with permissions from AWS S3 or an issue with my code. However, after digging a bit further, I discovered something rather alarming: if you use `<script src="/app.js"></script>` the page reloads. If you use `<script type="text/javascript" src="/app.js"></script>`, the page loads as expected. Considering it is [a part of the HTML5 standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) to exclude the `type` attribute, I am wondering... is this actually a Safari bug or am I missing something here? Is anyone else having this issue?

Note, this also occurs when I do not use this plugin (such as when I am using local development server). However, considering the script tag is central to this plugin, I figured I'd make an issue here to see if anyone else encountered this.

Edit, some examples of websites:
- http://pburtchaell.com
- http://lawnchair.io
- http://segment.social

pburtchaell:
* Update index.js to add type attribute
* Update tests
* Build examples to update HTML
* Update CHANGELOG

Resolves #309.


jantimon:
Please make sure it still passes the tests `npm run test`.

pburtchaell:
Whoops! I’ll get the tests updated.

jantimon:
👍

pburtchaell:
@jantimon Updated so tests pass.

pburtchaell:
@jantimon Any idea why the tests fail on CI? They pass locally for me.

jantimon:
If you rebase your branch against master you will get a better CI feedback.

To test it on your local machine:
```
npm rm webpack
npm i webpack@beta
npm test
```

Ignore the peer dependency errors ;)

jantimon:
Merged into 2.17.0

pburtchaell:
So it turns out this wasn't entirely true. What was breaking my apps was the `font-feature-settings: normal;` CSS property. If you use it anywhere in Safari, the page does not load.

NekR:
It is weird that this was even accepted. Browsers support script tags without type attribute for many-many years already.

jantimon:
going back would break backwards compatibility..

NekR:
Btw, HTML minifiers remove type attribute anyway.
Not sure I get it, why it will break back-compat?

jantimon:
If you are using the current version and your javascript is filtering for script files e.g. `document.querySelector('script[type]')` it would stop working as soon as we revert this.

NekR:
Okay, sounds fair (_although I don't see why someone would do it_).

I see you don't ever bump major version, but you can grab all potential changes and do that (_just for example_).

jantimon:
Many people don't upgrade to new major versions and as you said before if those `type` tags bother you just use the minifier options

PR #313, opened 2016-05-03
Fix a bug where not all CSS files were injected

jhnns:
This commit fixes a bug where CSS files were not recognized that have been exported with the file-loader.

Instead of searching all chunks for referenced CSS files and deduping it afterwards, we just skim all `compilation.assets` for CSS files. This allows us to also pick up CSS files that have been exported via the file-loader.

Use cases for this change:
- CSS files that have been processed with the [extract-loader](https://github.com/peerigon/extract-loader)
- CSS files that should be hashed and emitted to the output dir without processing it via the css-loader (because they are ready-to-use, for instance)

Since all previous tests are still working without changing them, I assume that this is not a breaking change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -388,7 +388,15 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
     // Will contain all js files
     js: [],
     // Will contain all css files
-    css: [],
+    css: Object.keys(compilation.assets).filter(function (assetFile) {
+      return path.extname(assetFile) === '.css';
+    }).map(function (cssFile) {
+      cssFile = publicPath + cssFile;
+      if (self.options.hash) {
+        cssFile = self.appendHash(cssFile, webpackStatsJson.hash);
+      }
+      return cssFile;
+    }),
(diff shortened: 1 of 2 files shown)


jantimon:
Got to review this in detail but wouldn't this conflict with filtering css files with the `chunks` and `excludeChunks` option?

jhnns:
Yes, that's correct. With this PR, filtering CSS files with `chunks` or `excludeChunks` will not work anymore. Apparently there were no tests for this use-case.

jantimon:
We should all css files which aren't part of the chunks.
Whould that be okay for you?

jhnns:
> Whould that be okay for you?

Yes, I guess so.

I wonder if there are cases, where this is actually not desired? Like, you don't want to add all stylesheets on all generated HTML... but if you didn't want that, you'd probably attach these stylesheets to specific chunks?

jantimon:
As using the file loader for css is really uncommon we could also move this to a small plugin for the html-webpack-plugin similar to https://github.com/jantimon/html-webpack-harddisk-plugin

PR #319, opened 2016-05-06
∃ ∀ hot-update.js for isHotUpdateCompilation

graingert:


jantimon:
Wouldn't it be smarter (faster) to check for the length instead of searching through the array twice?

graingert:
Yeah
On 9 May 2016 9:27 am, "Jan Nicklas" <notifications@github.com> wrote:

Wouldn't it be smarter (faster) to check for the length instead of
searching through the array twice?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/ampedandwired/html-webpack-plugin/pull/319#issuecomment-217805182>

jantimon:
Merged 👍

PR #338, opened 2016-06-02
jade -> pug

doriandrn:
This does not work, any ideas? Seems to be something wrong with the PUG loader. I've tried "pug-html-loader" also, still broken. Would love to see a resolution here!


jantimon:
I can't fix the loader here sorry.

Start with correct template paths:

```
+  Entry module not found: Error: Cannot resolve 'file' or 'directory' /Users/dor  ian/Downloads/html-webpack-plugin-master/examples/inline/template.pug in /User  s/dorian/Downloads/html-webpack-plugin-master/examples/inline:
```

doriandrn:
Ehm, forgot the dist file in and commited in rush, didn't even see the file mode changes. Anyways, you should try compiling it yourself to see what I mean. I'm sure paths are right.

jantimon:
Can I close this?

PR #345, opened 2016-06-09
Add 'html-webpack-plugin-alter-asset-tags'

jantimon:


index.js
```diff
@@ -143,8 +143,15 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
           });
       })
       .then(function (html) {
-        // Add the stylesheets, scripts and so on to the resulting html
-        return self.postProcessHtml(html, assets);
+        // Prepare script and link tags
+        var assetTags = self.generateAssetTags(assets);
+        var pluginArgs = {head: assetTags.head, body: assetTags.body, plugin: self, outputName: self.childCompilationOutputName};
```
SPSpwetter: Please add: context: compilation to this.

jantimon: Well that is already available in the plugin context - so you could do sth like:
```js
compiler.plugin('compilation', function(compilation) {
   compilation.plugin('html-webpack-plugin-alter-asset-tags', callback.bind(this, compilation));
}).bind(this);

function callback(compilation, htmlWebpackPluginData) {

}
```

jantimon:
Will probably land in 2.21.0

SPSpwetter:
Great. I think I'm close on the SRIPlugin side, (code is written), I just have to figure out how to debug webpack plugins.

index.js
SPSpwetter: Ah, thanks. Doing that. 

-- 
Pierce Wetter
Sent with Airmail

jantimon:
I am using iron-node and set breakpoints.
And integration tests ;)

PR #351, opened 2016-06-11
Allow scripts to be injected in a <div inject></div> tag

nathanhleung:
If we want to add extra scripts after our webpack script (without creating a new script file), we can now do this:

```jade
doctype html
html
  head
    title Home
    meta(charset='utf-8')
  body
    h1 Welcome
    //- In our webpack index.js, we do window.$ = require('jquery'), etc.
    div(inject)
    script.
      $('body').addClass('home');
      // other homepage-specific scripts


jantimon:
Please use https://github.com/jaketrent/html-webpack-template or create a similar template.

PR #352, opened 2016-06-14
add peerDeps, fix Cannot find module

jaredleechn:
`Error: Cannot find module 'webpack/lib/node/NodeTemplatePlugin'`

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,9 @@
     "webpack": "^1.13.0",
     "webpack-recompilation-simulator": "^1.3.0"
   },
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


jantimon:
Hmm could you change the peerDependency version to `*`? 
Or do you think this might cause unintended issues?

jaredleechn:
> Or do you think this might cause unintended issues?

yep, the major version maybe including some break changes, so I think keep `1.x` is better

jantimon:
But we have also support for webpack 2

jaredleechn:
how about `>= 1.12.0 < 3`

Issue #353, opened 2016-06-15
promise return by compileTemplate should handle err before any logic


vincentngthu:
html-webpack-plugin/lib/compiler.js:L71
should handle err before any logic. Otherwise the error stack will lost.
```javascript
  return new Promise(function (resolve, reject) {
    childCompiler.runAsChild(function (err, entries, childCompilation) {
      if(err) {reject(err); return;} // handle err
      // Replace [hash] placeholders in filename
      var outputName = compilation.mainTemplate.applyPluginsWaterfall('asset-path', outputOptions.filename, {
        hash: childCompilation.hash,
        chunk: entries[0]
      });
      ...
    })
  }
```

jantimon:
Sounds reasonable could you create a pull request?

PR #354, opened 2016-06-15
Handle err before any logic in promise return by compileTemplate

vincentngthu:
ref #353 handle err before any logic in promise return by compileTemplate


jantimon:
please make sure you pass the tests.
why don't you move the existing error handler up?

jantimon:
https://github.com/vincentngthu/html-webpack-plugin/blob/e493348953a6c3903b5e2ad11b314ae4f71af239/lib/compiler.js#L88-L94

PR #357, opened 2016-06-21
Added examples of using the plugin to compile into a Web Component html

homeslicesolutions:
I figured a way to make it to compile to a web component.  I think this is very useful for those who are working with web components!  Let me know what you think!

### DIFF
--- a/examples/web-component/template.ejs
+++ b/examples/web-component/template.ejs
@@ -1,7 +1,9 @@
 <template>
-  <% htmlWebpackPlugin.files.css.forEach(function(cssFile) { %>
-  <style>@import "<%=cssFile%>";</style>
-  <% }) %>
+  <style>
+    <% htmlWebpackPlugin.files.css.forEach(function(cssFile) { %>
+      @import "<%=cssFile%>";
+    <% }) %>
+  </style>
   <div id="mountTarget"></div>
 </template>
 <% htmlWebpackPlugin.files.js.forEach(function(jsFile) { %>
--- a/examples/web-component/template.ejs
+++ b/examples/web-component/template.ejs
@@ -1,0 +1,9 @@
+<template>
+  <% htmlWebpackPlugin.files.css.forEach(function(cssFile) { %>
+  <style>@import "<%=cssFile%>";</style>
+  <% }) %>
+  <div id="mountTarget"></div>
+</template>
+<% htmlWebpackPlugin.files.js.forEach(function(jsFile) { %>
+<script type="text/javascript" src="<%=jsFile %>"></script>
+<% }) %>
--- a/examples/web-component/webpack.config.js
+++ b/examples/web-component/webpack.config.js
@@ -1,0 +1,25 @@
+var HtmlWebpackPlugin = require('../..');
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
+var webpackMajorVersion = require('webpack/package.json').version.split('.')[0];
+module.exports = {
+  entry: './example.js',
+  output: {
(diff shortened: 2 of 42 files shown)


examples/web-component/webpack.config.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+  module: {
+    loaders: [
+      { test: /\.css$/, loader: ExtractTextPlugin.extract('style-loader', 'css-loader') }
+    ]
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
+      inject: false,
+      cache: false,
```
jantimon: why do you have to disable the cache?

homeslicesolutions: There actually no need for that.  I just copied and pasted from one of the examples.  Sorry for the sloppiness.  No need to disable cache.

examples/web-component/template.ejs
```diff
@@ -0,0 +1,9 @@
+<template>
+  <% htmlWebpackPlugin.files.css.forEach(function(cssFile) { %>
+  <style>@import "<%=cssFile%>";</style>
```
jantimon: What about surrounding the loop with the style tag?

jantimon:
Looks cool - the webpack 2 tests are missing right now.
Would this also work for multiple web components?

homeslicesolutions:
@jantimon Yes there are ways to do that.  I can include that example as well.  I imagine the desire would have one importable HTML web component per entry point?

examples/web-component/template.ejs
homeslicesolutions: Touche.  Good look!

homeslicesolutions:
@jantimon I have added new changes and an example of "multiple" web components per entry point

homeslicesolutions:
@jantimon Do you know why these tests are failing one after the other?  Are these errors a false positive?

jantimon:
You have to remove the webpack 1 module install webpack 2 module and build your new examples

homeslicesolutions:
I see the issue, so the "semistandard" prepublish fails because it can't find "HTMLElement" a webcomponent feature.  How can I get it to ignore that check?  I tried environment targeting to browser but no fix.  This is getting sort of ridiculous just adding a few examples...

jantimon:
Sorry :)
We use the examples for integration tests to ensure that every example will work after we changed the plugin.. Don't know what you mean by HTMLElement

homeslicesolutions:
So in my examples, I am showing that you can use this plugin to bundle valid HTML Web Components.  Web components figure 4 new technologies that exists native-ly in a few popular browsers including Chrome and Firefox.  One main feature of Web Components is "Custom Element" which exposes an API class called "HTMLElement"  which you can extend and create your own custom element.  semistandard is using eslint or standard or something to make a check in the JS files and doesn't recognize "HTMLElement" as a valid class possibly there's a setting or the library is outdated, hence it fails.  How familiar are you with the semistandard library and are there ways to stub in the HTMLElement class?

jantimon:
Just use `window.HTMLElement`.

homeslicesolutions:
Wow ok! =)  it worked.  Be back with an update!

homeslicesolutions:
I almost want to give up hahah =p.  What's going on?  So there's a Appcache failure in the test build.  And locally, i noticed that appcache isn't compatible with webpack 2.  Whats going on here?  Seems like a mess.

homeslicesolutions:
image

jantimon:
Welcome to my world ;)
Just ignore those errors - they work fine 👍

homeslicesolutions:
alright so you bless this pull request? ;)

Issue #389, opened 2016-07-20
[Feature request] Add injection of assets at specific position


chrisber:
I'm facing the issue that I need to inject the assets at a specific position within the head tag.
example:
```
<script type="text/javascript" src="django.i18.gettext"></script>
<script type="text/javascript" src="polyfills.bundle.js"></script>
<script type="text/javascript" src="vendor.bundle.js"></script>
<script type="text/javascript" src="main.bundle.js"></script>
```
It would be helpful if I could insert an html tag that is replaced by the `html-webpack-plugin`
example:
```
<script type="text/javascript" src="django.i18.gettext"></script>
<!--HtmlWebpackPlugin-Inject-Position-->
```

jantimon:
You could easily use a custom loader, the html-webpack-plugin hooks or the template system to implement this.

jantimon:
If you need assistance to move this feature into a plugin for the html-webpack-plugin please let me know

jehon:
Hello,

Having the same issue, I didn't find out how to do it.

I was looking for something like this, but this does not works anymore in the new version of the plugin:

```
  <!-- begin CSS files from webpack -->
  {% for (var css in o.htmlWebpackPlugin.files.css) { %}
  <link href="{%=o.htmlWebpackPlugin.files.css[css] %}" rel="stylesheet">
  {% } %}
  <!-- end   CSS files from webpack -->
  <!-- begin JS files from webpack -->
  {% for (var chunk in o.htmlWebpackPlugin.files.chunks) { %}
  <script src="{%=o.htmlWebpackPlugin.files.chunks[chunk].entry %}"></script>
  {% } %}
 <!-- end   JS files from webpack -->

```

How to do this in the new templating system?

Thanks

jantimon:
The o of o.htmlWebpackPlugin was removed

PR #390, opened 2016-07-20
[Feature request] Add injection of assets at specific position

chrisber:
This pull request adds the feature to inject assets at a specific position within the head or body element
ref #389


jantimon:
The inject feature is only a shortcut for the underlieing template system. If inject does not fulfill your needs please use the template variables.. they allow to solve every use case and can be used with any webpack supported template language even with pure JavaScript or react

PR #421, opened 2016-09-07
Add CSS to body if inject="body"

hampusohlsson:


jantimon:
Sorry this a breaking change and doesn't make much sense.
If you need this please use a custom template

hampusohlsson:
Gotcha, thanks for quick reply.

Some friendly feedback, the setting is confusing because on would think that specifying `inject:'body'` would have the desired effect of the PR / put both js/css into the body tag (instead of having two options `true` and `body` which are essentially the same).

The reason I'm suggesting it is because if you run a html-webpack file through [Google PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/) your score will be significantly decreased by having the css in the `<head>`-tag.

jantimon:
Is there any case where you would not want to have the above the fold css in your head?

hampusohlsson:
Yeah you are right, FOUC would happen if I put all css in the body. I guess an optimal solution would need some more sophisticated work... Like splitting the css into different chunks based on what's above-the-fold, inlining the most important parts in the head and then put the under-the-fold css in the body tag.

PR #423, opened 2016-09-08
order of chunks if specifying none is determined by order of includedChunks

willerup:
I have a use case where I would like the order of the chunks that is injected into the html to be as specified by the "chunks" array in the options. The problem is that when specifying 'none' as an sort method, the order becomes unpredictable if the entries in the webpack config is specified as a hash. 

I have added a test that shows the use case and also code that resolves it.
Let me know if you need more information.

### DIFF
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -1140,6 +1140,26 @@ describe('HtmlWebpackPlugin', function () {
       /<script type="text\/javascript" src="common_bundle.js">.+<script type="text\/javascript" src="util_bundle.js">.+<script type="text\/javascript" src="index_bundle.js">/], null, done);
   });
 
+  it('should not sort the chunks in the order provided in none mode', function (done) {
+    testHtmlPlugin({
+      entry: {
+        index: path.join(__dirname, 'fixtures/index.js'),
+        util: path.join(__dirname, 'fixtures/util.js')
+      },
+      output: {
+        path: OUTPUT_DIR,
+        filename: '[name]_bundle.js'
+      },
+      plugins: [
+        new HtmlWebpackPlugin({
+          chunks: ['util', 'index'],
+          chunksSortMode: 'none'
+        })
+      ]
+    }, [
+      /<script type="text\/javascript" src="util_bundle.js">.+<script type="text\/javascript" src="index_bundle.js">/], null, done);
+  });
+
   it('should sort the chunks in custom (reverse alphabetical) order', function (done) {
     testHtmlPlugin({
       entry: {
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -1140,11 +1140,11 @@ describe('HtmlWebpackPlugin', function () {
       /<script type="text\/javascript" src="common_bundle.js">.+<script type="text\/javascript" src="util_bundle.js">.+<script type="text\/javascript" src="index_bundle.js">/], null, done);
   });
 
-  it('should not sort the chunks in the order provided in none mode', function (done) {
+  it('should sort the chunks in the order provided in none mode', function (done) {
     testHtmlPlugin({
       entry: {
-        index: path.join(__dirname, 'fixtures/index.js'),
-        util: path.join(__dirname, 'fixtures/util.js')
+        util: path.join(__dirname, 'fixtures/util.js'),
+        index: path.join(__dirname, 'fixtures/index.js')
       },
       output: {
         path: OUTPUT_DIR,
--- a/index.js
+++ b/index.js
@@ -344,19 +344,19 @@ HtmlWebpackPlugin.prototype.sortChunks = function (chunks, sortMode) {
  * Return all chunks from the compilation result which match the exclude and include filters
  */
 HtmlWebpackPlugin.prototype.filterChunks = function (webpackStatsJson, includedChunks, excludedChunks) {
-    if (Array.isArray(includedChunks)) {
-        // Looping through the includedChunks instead of filtering,
-        // ensures that the order of the chunks is order of includedChunks
-        return includedChunks.map(function(name) {
-            for (var i=0; i < webpackStatsJson.chunks.length; i++) {
-                var chunk = webpackStatsJson.chunks[i];
-                if (chunk.names[0] === name) {
-                    return chunk;
-                }
-            }
-            // unknown chunk, ignore
-        })
-    };
+  if (Array.isArray(includedChunks)) {
+    // Looping through the includedChunks instead of filtering,
+    // ensures that the order of the chunks is order of includedChunks
+    return includedChunks.map(function(name) {
+      for (var i=0; i < webpackStatsJson.chunks.length; i++) {
+        var chunk = webpackStatsJson.chunks[i];
+        if (chunk.names[0] === name) {
+          return chunk;
+        }
+      }
+      // unknown chunk, ignore
+    });
+  }
 
   return webpackStatsJson.chunks.filter(function (chunk) {
     var chunkName = chunk.names[0];
--- a/index.js
+++ b/index.js
@@ -344,6 +344,20 @@ HtmlWebpackPlugin.prototype.sortChunks = function (chunks, sortMode) {
  * Return all chunks from the compilation result which match the exclude and include filters
  */
 HtmlWebpackPlugin.prototype.filterChunks = function (webpackStatsJson, includedChunks, excludedChunks) {
+    if (Array.isArray(includedChunks)) {
+        // Looping through the includedChunks instead of filtering,
+        // ensures that the order of the chunks is order of includedChunks
+        return includedChunks.map(function(name) {
+            for (var i=0; i < webpackStatsJson.chunks.length; i++) {
+                var chunk = webpackStatsJson.chunks[i];
+                if (chunk.names[0] === name) {
+                    return chunk;
+                }
+            }
+            // unknown chunk, ignore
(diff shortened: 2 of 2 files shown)


jantimon:
Chunks is not related to sorting but only filtering

willerup:
The problem is there is no way to control the order of chunks in this scenario:

```
entry: {
    a: "a.js",
    b: "b.js",
    c: ["c.js", "d.js"]
},
...
plugins: [
         new HtmlWebpackPlugin({
           chunks: ['a', 'b'],
           chunksSortMode: 'none'
         })
       ]
```

I want chunks to appear in the html in the order a then b. But since the entries are specified with a hash, the order that they come through to the chunks filter is unpredictable. So I found myself without any reasonable way to control the order. Is there another way to do this?

jantimon:
```js
entry: {
    a: "a.js",
    b: "b.js",
    c: ["c.js", "d.js"]
},
...
plugins: [
         new HtmlWebpackPlugin({
           chunksSortMode: function(entry1, entry2) {
               return 1; // <-- your fancy array sort method goes here :)
            }
         })
       ]

**Comment by willerup** [comment]
Yes this is actually what I ended up doing, but it seems over-complex to provide a sort function for a scenario where the sorting order is provided by a sequence of elements. This is my actual webpack.config file for one project:

```
  var orderByList = function(list) {
    return function(chunk1, chunk2) {
      var index1 = list.indexOf(chunk1.names[0]);
      var index2 = list.indexOf(chunk2.names[0]);
      if (index2 == -1 || index1 < index2) {
        return -1;
      }
      if (index1 == -1 || index1 > index2) {
        return 1;
      }
      return 0;
    };
  };

      config.plugins.push(
        new HtmlWebpackPlugin({
          filename: 'admin.html',
          chunks: ['admin-vendor', 'admin'],
          chunksSortMode: orderByList(['admin-vendor', 'admin']),
          template: './src/public/admin.html',
          inject: true
        }),
        new HtmlWebpackPlugin({
          filename: 'admin-local.html',
          chunks: ['admin-vendor', 'admin' ,'local'],
          chunksSortMode: orderByList(['admin-vendor', 'admin', 'local']),
          template: './src/public/admin.html',
          inject: true
        }),
        new HtmlWebpackPlugin({
          filename: 'index.html',
          chunks: ['main-vendor', 'main'],
          chunksSortMode: orderByList(['main-vendor', 'main']),
          template: './src/public/index.html',
          inject: true
        }),
        new HtmlWebpackPlugin({
          filename: 'index-local.html',
          chunks: ['main-vendor', 'main' ,'local'],
          chunksSortMode: orderByList(['main-vendor', 'leanlaw', 'local']),
          template: './src/public/index.html',
          inject: true
        }),
```

Seems over complicated, so I am thinking it would be much cleaner if the plugin supported this.
If you want to keep the `chunks` option as strictly a filter, would it be an idea to allow an array on the `chunksSortMode` option which provides the order, like this:

```
 new HtmlWebpackPlugin({
           chunks: ['a', 'b'],
           chunksSortMode: ['a', 'b']
         })
```

It would work but duplication is ugly. So perhaps introduce a new sort mode which makes it more explicit that the sorting is done based on the `chunks` option:

```
 new HtmlWebpackPlugin({
           chunks: ['a', 'b'],
           chunksSortMode: 'explicit'
         })
```

If you like any of these options, I'd be happy to submit them. 

I should add that my assumption going into this was that the sort order if 'none' is specified is derived from the entries which is also what you say [here](https://github.com/ampedandwired/html-webpack-plugin/issues/4) but again that only works if entries are specified as an array, not if they are specified as a hash.

**Comment by jantimon** [comment]
I still don't get what you mean with 'hash'.
It should just follow your webpack default order if you set it to none.
(as said before chunk options is not related in any way with sorting)
What is your entry configuration?

**Comment by willerup** [comment]
```
config.entry = {
      'main-vendor': './src/app/vendor.js',
      'main': './src/app/app.js',
      'admin-vendor': './src/admin/vendor.js',
      'admin': './src/admin/app.js',
      'local': './src/app.local.js'
  }
```

With this entry config and this HtmlWebpackPlugin config

```
        new HtmlWebpackPlugin({
          filename: 'admin-local.html',
          chunks: ['admin-vendor', 'admin' ,'local'],
          chunksSortMode: 'none',
          template: './src/public/admin.html',
          inject: true
        }),
```

The html generated is this:

```
<script type="text/javascript" src="/admin.1929c00f0d758eb1c41d.js"></script>
<script type="text/javascript" src="/admin-vendor.1929c00f0d758eb1c41d.js"></script>
<script type="text/javascript" src="/local.1929c00f0d758eb1c41d.js"></script>
```

Noticed that the order is not the same as specified in the entry config. Maybe what I am missing is that there is a way to explicitly specify order of entries in webpack when they are specified as hash?

**Comment by jantimon** [comment]
That's really strange don't know why webpack would change the order in that case.. that's definitely not the intended behaviour

File: spec/BasicSpec.js [hunk 1]
```diff
@@ -1140,6 +1140,26 @@ describe('HtmlWebpackPlugin', function () {
...
+        util: path.join(__dirname, 'fixtures/util.js'),
+        index: path.join(__dirname, 'fixtures/index.js')
+      },
+      output: {
+        path: OUTPUT_DIR,
+        filename: '[name]_bundle.js'
+      },
+      plugins: [
+        new HtmlWebpackPlugin({
+          chunks: ['util', 'index'],
```
jantimon: can you please remove this line as it is not related to sorting

willerup:
I updated the test case in the PR to clarify the problem. This test fails without my code change:

```
  it('should sort the chunks in the order provided in none mode', function (done) {
    testHtmlPlugin({
      entry: {
        util: path.join(__dirname, 'fixtures/util.js'),
        index: path.join(__dirname, 'fixtures/index.js')
      },
      output: {
        path: OUTPUT_DIR,
        filename: '[name]_bundle.js'
      },
      plugins: [
        new HtmlWebpackPlugin({
          chunks: ['util', 'index'],
          chunksSortMode: 'none'
        })
      ]
    }, [
      /<script type="text\/javascript" src="util_bundle.js">.+<script type="text\/javascript" src="index_bundle.js">/], null, done);
  });
```

I actually think it makes total sense. The "order" of the entry config is undefined if you specify it as an object as opposed to an array. Just because it says `entry: { util: ..., index: ... }` doesn't mean that `util` comes before `index`. They are just two properties in the object. http://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

spec/BasicSpec.js
willerup: You mean remove `chunks: ['util', 'index']`? If I remove that line, the test will fail.

willerup: You mean remove `chunks: ['util', 'index']`? If I remove that line, the test will fail.

jantimon: Then the test doesn't make sense as chunks should be not related to the sort order

willerup: Sorry but that's the whole point of my PR. If you leave out `chunks:['util','index']` the test will fail, with or without my code. It is not possible to specifying an explicit sort order ysing en entry config when it is specified as an object. Providing `entry: { 'util':.., 'index': ...}` and sortorder `none` will result in html where index is included before util.

I see the following options:

- Don't address it, requiring the caller to provide a sort function along the lines of what I quoted above. In that case we can just close this PR and move along.
- Address it by making the `chunks` option not only work as a filter but **also** as a way to specify the sort order. This is the solution I chose in this PR, and that is why the `chunks` option is part of the test because it specifies the sort order and that is what we are testing.
- Use `chunkSortMode` to specify an explicit sort order (`chunkSortModel: ['util','index']`)
- Other ideas?

willerup: I should add that for option 2, it might be better to introduce a new sort mode (as suggested earlier in the thread), for example `chunkSortMode: 'explicit'`. This way `chunkSortMode: 'none'` still works the way it used to and when specifying `explicit`, then `chunks` option is required and used to do the sorting. I am happy to update the PR to work this way if you like.

jantimon:
It should just use the order of your entry points

RPDeshaies:
Any status on this ?

tsteuwer:
For those of you having an issue with sort order of assets through webpack html plugin, I was able to get this to work the way I want by specifying:

```
chunks: ['thirdparty', 'angular', 'app'],
chunksSortMode: 'manual'
```

And it put them in that order (1) thirdparty, (2) angular, (3) app.

nillith:
I can't just manually reorder it. It's gives `webpackJsonp is not defined` in the console If I don't let the plugin do the ordering itself.

alernerdev:
I have followed @tsteuwer suggestion, and it worked. My HtmlWebpackPlugin looks as following:
```
new HtmlWebpackPlugin({
            template: "src/index.html",
            // I want script references to be injected into the head element
		inject: 'head',
		chunks: ['vendor', 'main'],
		chunksSortMode: 'manual'       
}),
```
where 'vendor' and 'main' come from

  ```
entry: {
        vendor: path.resolve(__dirname, 'src/vendor'),
        main: path.resolve(__dirname, 'src/index')
    },
```
Terrific.  Now, if someone could tell me how to inject those script references at the TOP of the <head> block rather than the bottom?  I have angular init code in the head element, and the vendor file with the angular reference is at the bottom of the head element --- so angular is not defined.  I could move my code to the top of the <body> element so it all falls into place, but I am wondering if there is a better way?  Thank you

PR #427, opened 2016-09-12
add options.injectOrder: specfiy your orders of js and css in output  html

MwumLi:
the options.inject can insert js at the bottom of the body element and css at  at the bottom of the head element, but the order of js and css are not what we expect . so I add it `options.injectOrder`,  which is a object, such as : 

``` javascript
        injectOrder:  {
              js: {
                  keys: ['commons', 'vendor', 'style', 'polyfills', 'app'],
                  func: orderFunc
              },
              css: {
                  keys: ['style', 'app'],
                  func: orderFunc
              }
        }
```

So, we can specify the order of js and css in output html, Look forward to merge


jantimon:
there is already a sort option

MwumLi:
@jantimon you mean  `chunksSortMode` ?

jantimon:
exactly 👍

jantimon:
You can also sort them using `html-webpack-plugin-alter-asset-tags` event but the chunksSortMode should be enough.

MwumLi:
@jantimon  Thanks
 `chunksSortMode` cann't meet my need, but 
Your words gave me inspiration, so I finished this plugin [html-webpack-alter-asset-plugin](https://www.npmjs.com/package/html-webpack-alter-asset-plugin), which based on `html-webpack-plugin-before-html-generation`

jantimon:
Cool :)

PR #434, opened 2016-09-18
Fix manifest not to add publicPath

lzxb:


jantimon:
Could you please add a unit test?

lzxb:
@jantimon 
可以，不过要过两天才有时间
Yes, but it will take two days to have time.

PR #441, opened 2016-09-24
replace [hash] in publicPath of favicon

chrillo:
[hash] replacement works for pretty much all asset types but if the publicPath includes the build hash it does not get replaced and the final html output for the favicon and still contains the [hash] placeholder. this change replaces the [hash] variable in the publicPath for the favicon.

### DIFF
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -1007,6 +1007,22 @@ describe('HtmlWebpackPlugin', function () {
     }, [/<link rel="shortcut icon" href="\/some\/+[^"]+\.ico">/], null, done);
   });
 
+  it('adds a favicon with a publichPath set to [hash]/ and replaces the hash', function (done) {
+    testHtmlPlugin({
+      entry: path.join(__dirname, 'fixtures/index.js'),
+      output: {
+        path: OUTPUT_DIR,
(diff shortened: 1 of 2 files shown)


jantimon:
Looks nice - could you please add a unit test?

chrillo:
@jantimon added a unittest, the test it self can't really check for the exact hash value the way the tests are set up since the hash is not available to the individual specs but the testHtmlPlugin method verifies the hash replacement anyways so I hope this is fine.

spec/BasicSpec.js
```diff
@@ -1007,6 +1007,22 @@ describe('HtmlWebpackPlugin', function () {
...
+        path: OUTPUT_DIR,
+        publicPath: '/[hash]/',
+        filename: 'index_bundle.js'
+      },
+      plugins: [
+        new HtmlWebpackPlugin({
+          favicon: path.join(__dirname, 'fixtures/favicon.ico')
+        })
+      ]
+    }, [/<link rel="shortcut icon" href="\/*\/+[^"]+\.ico">/], null, done);
```
jantimon: Could you try to be a little bit less general?

E.g. `\/[A-F0-9]{6}\/favicon\.ico/`

PR #450, opened 2016-10-06
Add an option that prevents head tag from being generated automatically.

eladchen:
**Description:**
When working with this plugin to generate multiple html files for a single-page-application
Missing `<head></head>` tags are added to the output.

**Issue:**
single page applications usually have a wireframe html file serving as the base / entry point, and multiple html files with nothing but raw html ( no html / body tags ).

The current way this plugin injects assets designated to the head section of a document, is to check whether an appropriate `<head>` tag exists before adding the relevant resources.
This can be problematic, since it can result in two files having `<head>` tags.

For example when entering the first page in an SPA: 
 - wireframe.html is processed and now contains `<head>` tag
 - profile.html is processed and now also contains a `<head>` tag

Such a scenario can create an invalid html layout.

I added an option called "createMissingTags" and set its default value to 'true'
This will make sure no breaking changes are added to the code base, and introduce an option
That will allow future control on the creating of missing tags in a scope larger than just the `<head>` 

I also added unit tests for this, as well as updated the docs.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Allowed values are as follows:
   included scripts and CSS files. This is useful for cache busting.
 - `cache`: `true | false` if `true` (default) try to emit the file only if it was changed.
 - `showErrors`: `true | false` if `true` (default) errors details will be written into the html page.
-- `createMissingTags`: `true | false` if `true` whether to create a `<head>` tag ( if it does not exists ) when injecting assets. default: 'true'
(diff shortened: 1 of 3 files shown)


jantimon:
Just set `inject` to false

eladchen:
You are not following me. 

Setting inject to false doesn't solve the problem, it actually makes it worse.

**I want assets to be injected** I just don't want tags I did not expect to be added brutally by the api. 

To clarify I'm using html-loader to disable the built in ejs template, and use the default behavior.

PR #478, opened 2016-10-28
Resolve 'html-webpack-plugin-after-emit' deprecated warnings

ascariandrea:
@jantimon I excluded the `html-webpack-plugin-after-emit` event from returned result check and fixed tests.

Take a look :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,7 +117,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       // Allow plugins to make changes to the assets before invoking the template
       // This only makes sense to use if `inject` is `false`
       .then(function (compilationResult) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -626,7 +626,7 @@ HtmlWebpackPlugin.prototype.applyPluginsAsyncWaterfall = function (compilation)
   return function (eventName, pluginArgs) {
     return promisedApplyPluginsAsyncWaterfall(eventName, pluginArgs)
       .then(function (result) {
-        if (!result) {
+        if (!result && eventName !== 'html-webpack-plugin-after-emit') {
```
jantimon: please don't hardcode event names into a util functions

how about having two util functions or an argument - `requiresResult` or sth similar?

ascariandrea: You're definitely right!
I've corrected my code

PR #492, opened 2016-11-13
Type check options.minify instead of silently failing

mohsen1:
Wasn't sure where you check for options type. Please let me know how this should be done according to current project architecture. 

I would also suggest using JSON Schema to enforce options.


jantimon:
This should be part of the 'html-minifier' project as we just pass the options:

```
if (self.options.minify) {
        var minify = require('html-minifier').minify;
        return minify(html, self.options.minify);
      }
```

PR #494, opened 2016-11-14
Document how to pass in template variables

likethemammal:


jantimon:
Could we instead update https://github.com/ampedandwired/html-webpack-plugin#writing-your-own-templates ?

PR #497, opened 2016-11-19
Add the ability to exclude assets based on RegExp patterns

jamesjieye:
When adding an entry, for example `style: ['bootstrap/dist/css/bootstrap.css']`, with third party css files to webpack, the injected scripts include `style.js` or `style.[chunkhash].js`. The `excludeChunks` will exclude both `style.css` and `style.js`. With `excludeAssets`, you can keep `style.css` in and `style.js` out.


jamesjieye:
My local test passed. Not sure why the travis-ci failed.

screen shot 2016-11-19 at 10 53 12 pm

jantimon:
What about moving that into a custom plugin like https://github.com/DustinJackson/html-webpack-inline-source-plugin/blob/master/index.js

jamesjieye:
@jantimon, good idea! Let me get it moved.

jamesjieye:
@jantimon, it's done. Moved to https://github.com/jamesjieye/html-webpack-exclude-assets-plugin.

jantimon:
Nice :)

PR #500, opened 2016-11-23
Update docs for html-loader minimize option

zhulongzheng:
The minimization of html-loader can be disabled by setting a query.

### DIFF
--- a/docs/template-option.md
+++ b/docs/template-option.md
@@ -51,7 +51,7 @@ new HtmlWebpackPlugin({
 ```
 
 However this also means that in the following example webpack will use the html loader for your template.
-This will **cause html minification** and it will also **disable the ejs fallback** loader.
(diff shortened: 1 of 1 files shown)


jantimon:
I guess this is already part of the html-loader docs :)

zhulongzheng:
Actually the html-loader doc didn't mention this option, and I've already created a [pull request](https://github.com/webpack/html-loader/pull/104) for html-loader doc.
And even if it's part of docs, I think it is also **necessary** to let the html-webpack-plugin users know that the there's one option to disable the html-loader minimization, or maybe if you want to use html-loader with html-webpack-plugin you should always specify the `minimize: false` cause if you use `minify` of html-webpack-plugin the html template will be minimized for twice, and it might cause bug as the [pull request](https://github.com/webpack/html-loader/pull/104) says.

PR #514, opened 2016-12-05
Enable unvalued element attributes.

numical:
Minor fix to [createHtmlTag](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L583) function to alow elements to have attributes with no values.
This is an enabler for add-on plugins (such as [ScriptExtHtmlWebpackPlugin](https://github.com/numical/script-ext-html-webpack-plugin/issues/3)).

The fix is trivial.  A whole additional Spec object has been created as, conditional on acceptance of this PR, another PR will be submitted with additional Add-On tests to replace [PR#510](https://github.com/ampedandwired/html-webpack-plugin/pull/510).

This PR has been tested locally against node v4,v5,v6.

### DIFF
--- a/index.js
+++ b/index.js
@@ -582,7 +582,8 @@ HtmlWebpackPlugin.prototype.appendHash = function (url, hash) {
  */
 HtmlWebpackPlugin.prototype.createHtmlTag = function (tagDefinition) {
   var attributes = Object.keys(tagDefinition.attributes || {}).map(function (attributeName) {
-    return attributeName + '="' + tagDefinition.attributes[attributeName] + '"';
+    const value = tagDefinition.attributes[attributeName];
+    return value ? attributeName + '="' + tagDefinition.attributes[attributeName] + '"' : attributeName;
   });
   return '<' + [tagDefinition.tagName].concat(attributes).join(' ') + (tagDefinition.selfClosingTag ? '/' : '') + '>' +
     (tagDefinition.innerHTML || '') +
--- a/spec/AddOnSpec.js
+++ b/spec/AddOnSpec.js
@@ -1,0 +1,123 @@
+/*
+ * Integration and unit tests for all features but caching
+ */
+
+/* eslint-env jasmine */
+'use strict';
+
+// Workaround for css-loader issue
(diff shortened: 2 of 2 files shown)


jantimon:
Why do we need such a complex test to test such a simple feature?

numical:
Hi Jan,
Good to see you back.
My reasoning:
1. I created a whole new Spec file to test add-on functionality - if you accept this PR I will follow it up with an additional PR with further tests in this new Spec file (this would be a replacement for [rejected PR510](https://github.com/ampedandwired/html-webpack-plugin/pull/510));
2. I could have created a minimal Spec file dealing with just this one test.  Instead I copied the testing functionality provided in [`BasicSpec.js`](https://github.com/ampedandwired/html-webpack-plugin/blob/master/spec/BasicSpec.js).  I copied because:
3. the tests would be more familiar to yourself;
4. if I did not, I might just have to reintroduce the test functionality in the follow-on PR.

If you are happier I will instead write a minimal new Spec file, or even just append the test to 'BasicSpec.js` - though I don't recommend that as the nature of the test is rather different to all the other tests in that file.

numical:
I have created an [alternative PR](https://github.com/ampedandwired/html-webpack-plugin/pull/529) with a simpler AddOnSpec file.

stevewillard:
+1 looking to inline some of my scripts, and it looks like this is holding things up.

jantimon:
@numical thanks a lot for your work :)
I am really sorry for beeing not that responsive right now.

We already have addon tests in the basic spec - could you just copy one of these?
https://github.com/ampedandwired/html-webpack-plugin/blob/master/spec/BasicSpec.js#L813

numical:
No worries!  It's Christmas.
Sure, I can copy one of the BasicSpec tests - I was actually going to extract _them_ into my new AddOnSpec in a follow-on PR as, to my mind, BasicSpec is getting rather unwieldy.  
However you're the boss.  Will do, hopefully before the new year.

numical:
Replaced by [PR531](https://github.com/ampedandwired/html-webpack-plugin/pull/531)

PR #522, opened 2016-12-08
Simplify script and meta tags

tkrotoff:
Simplifications:

- `<script type="text/javascript" ...>` replaced by `<script ...>`

See https://www.w3.org/TR/html5/scripting-1.html#attr-script-type *“The default, which is used if the attribute is absent, is "text/javascript””*

FYI The old HTML 4 specification requires `type` attribute but even in HTML 4 mode (`<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">`), IE6 works without specifying the `type` attribute (tested myself).

- `<meta http-equiv="Content-type" content="text/html; charset=utf-8"/>` replaced by `<meta charset="utf-8"/>`

See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Obsolete_things_to_avoid
See http://stackoverflow.com/questions/4696499

All tests passing locally.

Less code is always good :)


tkrotoff:
No idea why tests fail with Travis CI, on my Mac `npm run test` runs just fine.

tkrotoff:
@numical do you really believe that <script> without `type` does not work under Safari 9.1?? Half the web omits the `type` attribute, even http://www.apple.com/ does not use it. All minifiers remove this attribute.

The guy just did something wrong in his code, that's all.

tkrotoff:
@numical LOL
> So it turns out this wasn't entirely true. What was breaking my apps was the `font-feature-settings: normal` CSS property.

=> no shit...

> It is weird that this was even accepted

> Btw, HTML minifiers remove type attribute anyway

numical:
Just pointing out the history. 
You will have seen @jantimon's reason for not backing out the PR, however mistaken, so I imagine he is unlikely to accept this one.

tkrotoff:
> Just pointing out the history

Thx for that. I genuinely made this PR thinking `type` was here by mistake. This PR also features `<meta>` tag simplification.

> I imagine he is unlikely to accept this one

Maybe he will reconsider

> @jantimon: [...] If your javascript is filtering for script files e.g. `document.querySelector('script[type]')` it would stop working

If someone wrote this - highly improbable IMHO - it will break sooner or later (HTML minifier, changing from html-webpack-plugin to another tool...).

The longer `<script type="text/javascript" ...>` stays, higher is the risk to see someone writing `$('script[type]')`.

If you are afraid someone might have written `$('script[type]')`, the solution is to bump semver major version number.

SpaceK33z:
Oh I just now see this PR. I made a PR removing only the `type="text/javascript"` thingies (#572) because this was discussed in #558. The difference is that my PR builds successfully though.

tkrotoff:
> The difference is that my PR builds successfully though

This one too: Travis CI failed for other reasons.

jantimon:
@SpaceK33z were merged into the [version-3.x branch](https://github.com/ampedandwired/html-webpack-plugin/tree/version-3.x) Thanks @tkrotoff  @SpaceK33z

tkrotoff:
@jantimon There are 2 other commits in this PR:
- https://github.com/ampedandwired/html-webpack-plugin/pull/522/commits/3c03d3ad649fbdb24f7611f98fcd55693334a2c1
- https://github.com/ampedandwired/html-webpack-plugin/pull/522/commits/ee2df7c9e1065c369d6e8780f96286ede85a5028

that would be a nice addition

jantimon:
Fixed thanks a lot 👍

Daniel-Khodabakhsh:
I'm on 3.0.6 and html-webpack-plugin is still outputting `<script type="text/javascript" ...`.

PR #531, opened 2016-12-27
Enable unvalued element attributes

numical:
Minor fix to [createHtmlTag](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L583) function to allow elements to have attributes with no values.
This is an enabler for add-on plugins (such as [ScriptExtHtmlWebpackPlugin](https://github.com/numical/script-ext-html-webpack-plugin/issues/3)).
Single test added to BaseSpec.js.

This is a replacement for PR's [514](https://github.com/ampedandwired/html-webpack-plugin/pull/514) and [529](https://github.com/ampedandwired/html-webpack-plugin/pull/529)


numical:
@jantimon - just a reminder for you to take a look at [this PR](https://github.com/ampedandwired/html-webpack-plugin/pull/531).  It's a low risk change for html-webpack-plugin but a blocker for the new release of [script-ext-html-webpack-plugin](https://github.com/numical/sscript-ext-html-webpack-plugin) with no less than [three issues](https://github.com/numical/script-ext-html-webpack-plugin/issues) hanging off it

jantimon:
@numical sorry for the delay :)

What do you think about the following change:

https://github.com/ampedandwired/html-webpack-plugin/compare/numical-enable-unvalued-element-attributes-base-spec-test?expand=1

numical:
So I would have to pass the value `true` after all? It seems a little counter-intuitive but if you consider the ability to remove an attribute with a `false` value useful, I can live with it!
BTW - we both fall down if the user is trying to produce XHTML - is this part of html-webpack-plugin's scope?

jantimon:
I am not 100% sure but it seems to be better than setting `async` to `undefined` which is falsy to declare that the `script` is `async`.

Your proposal was if `value` so async='' or async=null or async=0 or async=false would have been a pure attribute.

What is the problem with xhml?

numical:
I was going down the official HTML viewpoint:
_"Some attributes play the role of boolean variables (e.g., the selected attribute for the OPTION element). Their appearance in the start tag of an element implies that the value of the attribute is "true". Their absence implies a value of "false"."_([w3c](https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.3.4.2))  However this is JavaScript so I'm happy to go along with your take on it.

Regarding XHTML - XML does not allow valueless attributes so, again, same W3C reference:
_"Boolean attributes may legally take a single value: the name of the attribute itself (e.g., selected="selected")."_ - this is specifically to accommodate XHTML which our solution does not.
I think that if XHTML becomes an issue, we leave your code as-is and I add an 'xhtml' flag to script-ext-html-webpack-plugin to pass a valued attribute when necessary.

jantimon:
HTML: Okay cool according to those docs true/false would be a perfect fit.

XHTML: There is already a xhtml option in the html-webpack-plugin which you can use :)

jantimon:
Merged

PR #535, opened 2016-12-29
Allow config to specify the hash's query string variable name

willmorgan:
I'd like to be able to set the the query string variable for caching, so that if I specify a whitelist of cache busting query string variables in CloudFront, I can use this plugin to automatically bust the CloudFront cache.

Hopefully this is a good change 👍


willmorgan:
@ampedandwired / @jantimon any thoughts on this? Travis failure is due to some other factors it seems.

willmorgan:
Cheers for the pointer @jantimon.

Issue #541, opened 2017-01-10
Template path resolving error if passing loader query parameters


PeachScript:
Hello, I got an error `fullTemplatePath` when I passing loader query parameters like the following code:

```js
new HtmlWebpackPlugin({
  filename: 'index.html',
  template: '!!html!swig?raw=true!./src/tmpl/index.html'
})
```

Then I got an error `fullTemplatePath`:
```
!!html!/Absolute/path/to/context/swig?raw=true!./src/tmpl/index.swig
```

So the `Child compilation` certainly failed, I read the source code and debug it, I think this problem caused by regexp in the `getFullTemplatePath` function, because the current regexp no deal with loader query parameters, it will match the `swig` substring and resolve it as a file path, is that right?

PeachScript:
I created a pull request to fix it, please review it.

And this PR has been tested locally against Node.js v4, v5, v6, v7, but I don't know why the `travis-ci` failed...

PeachScript:
PR has been merged, close this issue now.

PR #542, opened 2017-01-10
Fix template path resolving regexp to support loader query parameters

PeachScript:
More details in this issue: #541

### DIFF
--- a/index.js
+++ b/index.js
@@ -606,7 +606,7 @@ HtmlWebpackPlugin.prototype.getFullTemplatePath = function (template, context) {
   }
   // Resolve template path
   return template.replace(
-    /([!])([^/\\][^!?]+|[^/\\!?])($|\?.+$)/,
+    /([!])([^/\\][^!?]+|[^/\\!?])($|\?[^!?\n]+$)/,
     function (match, prefix, filepath, postfix) {
       return prefix + path.resolve(filepath) + postfix;
     });


jantimon:
Could you tell a little bit more about the changed regexp?

PeachScript:
Of course, I have described the problem in the issue #541 .

jantimon:
Can you please also explain the solution?

PeachScript:
Before change, the regular expression /([!])([^/\\][^!?]+|[^/\\!?])($|\?.+$)/ will match the following substrings in `!!html!swig?raw=true!./src/tmpl/index.html`:
1. `!` - prefix
2. `swig` - filepath
3. `?raw=true!./src/tmpl/index.html` - postfix

And the `getFullTemplatePath` function will return the error full template path `!!html!/Absolute/path/to/context/swig?raw=true!./src/tmpl/index.swig`, because of the regular expression ignore the remaining string if it read a `?`, and the substring `swig` be treated as a file path, the real file path be treated as a part of the postfix.

After change, the regular expression /([!])([^/\\][^!?]+|[^/\\!?])($|\?**[^!?\n]**+$)/ (Only need to look at the bold content) will match the following substrings in `!!html!swig?raw=true!./src/tmpl/index.html`:
1. `!` - prefix
2. `./src/tmpl/index.html` - filepath
3. `` - postfix

And the `getFullTemplatePath` function will return the correct full template path `!!html!swig?raw=true!/Absolute/path/to/context/src/tmpl/index.swig`, because of the new regular expression doesn't match `!` and `?` after `?`, so it will not treat the loader query parameters as the postfix, this is the solution.

---

The key source code of the `getFullTemplatePath` function:
```js
// Resolve template path
return template.replace(
  /([!])([^/\\][^!?]+|[^/\\!?])($|\?[^!?\n]+$)/,
  function (match, prefix, filepath, postfix) {
    return prefix + path.resolve(filepath) + postfix;
  });
```

jantimon:
That's quite smart - thanks a lot 👍

PeachScript:
Glad you like it :D

PeachScript:
By the way, when do you plan to release the new version?

jantimon:
Released

PR #546, opened 2017-01-16
Feature/control head tag injection

A0dh:
There are cases when `<head>` tag is not preferred to be automatically inserted into the html template. Hope `injectHeadTag` option name is a good choice to be able to control the injection behaviour

### DIFF
--- a/index.js
+++ b/index.js
@@ -521,22 +521,22 @@ HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets, asset
   var htmlRegExp = /(<html[^>]*>)/i;
   var headRegExp = /(<\/head>)/i;
   var bodyRegExp = /(<\/body>)/i;
-  var body = assetTags.body.map(this.createHtmlTag);
-  var head = assetTags.head.map(this.createHtmlTag);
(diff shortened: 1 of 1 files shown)


jantimon:
In that case just disable the `inject` feature and use the template way to add the scripts (see more in the readme)

A0dh:
It was not for scripts, but for styles. I believe it forces the `head` tag if one does not exist. And by looks of it the template approach will not help here.

jantimon:
Why? the template approach can do all what the inject approach can do.. but with more control

A0dh:
@jantimon thanks for taking your time to discuss. I still don't get it. Currently I am using handlebars templates and `html-webpack -plugin` to generate partial html templates. However there is no way I can disable auto adding of `head` tag. I checked the `html-webpack -plugin` code, in particular `injectAssetsIntoHtml` method. And it looks like it will add the `head` tag regardless and there is no way to stop it. Am I looking in the wrong direction? Can you please point me to the right one? 
Thanks in advance!

jantimon:
Hey @A0dh - as said before the entire `inject`-option which will also add the `head` tag can be entirely disabled.

PR #552, opened 2017-01-20
Fix travis-ci: set tty size

spuf:
`PrettyErrors`'s  dependecy `RenderKid` needs TTY columns to word-wrap lines to console.


jantimon:
That's quite awesome

Issue #558, opened 2017-01-23
Version 3.0


jantimon:
At all contributors
@SpaceK33z, @SimenB, @numical, @kennyt, @okhomenko, @cgreening, @anjianshi, @mistadikay, @mc-zone, @JSteunou, @spuf, @vincentngthu, @NekR, @mateatslc, @chiplay, @ascariandrea, @graingert, @zzuieliyaoli, @foglerek, @sandeep45, @koenpunt, @sokra, @shprink, @cesarandreu, @oliviertassinari, @sebastienbarre, @jamesjieye, @AndrewRayCode, @kurtharriger, @ai, @PeachScript, @kinday, @TalAter, @chrillo, @timoxley, @insin, @adjohnson916, @Madalosso, @mgol, @victorwpbastos, @pburtchaell, @dtinth, @yyx990803, @michael-wolfenden, @denis-sokolov, @jirikolarik, @jpetitcolas, @ruehl, @Foxandxss, @creeperyang, @vyorkin, @thomasguillory, @tlrobinson, @nartamonov

I am planning to move the webpack-html-plugin forward to 3.0 and would like to gather your input which (breaking-)changes you would like to see

Planed changes:

- Use `stats.entrypoints` to determine which links/scripts should be added to the html. (Unfortunately this will stop the support for webpack 1 as @sokra won't port this feature back into webpack 1)
- Finally jump up some node versions to node 4 (the official supported version for webpack 2)
- Remove `type: 'text/javascript'`
- Remove html minifcation (could still be done by a plugin if needed)
- Remove chunksSortMode (as `stats.entrypoints` are already sorted)

Open for discussion:

- Switch to `ava`
- Generate multiple html files (e.g. by using [name] as output name)

graingert:
these changes won't break my build.

SpaceK33z:
Sounds like good changes, I agree with all of them 👍 . Let me know if you need any help with the changes, I maybe have some time next week.

victorwpbastos:
Better favicon support would be fantastic.

Em 23 de jan de 2017 21:18, "Kees Kluskens" <notifications@github.com>
escreveu:

Sounds like good changes, I agree with all of them 👍 . Let me know if you
need any help with the changes, I maybe have some time next week.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ampedandwired/html-webpack-plugin/issues/558#issuecomment-274649505>,
or mute the thread

[...]

PR #572, opened 2017-01-29
[v3] Remove `text/javascript` from all script tags

SpaceK33z:
As (shortly) discussed in #558, this removes `text/javascript` from all script tags.


jantimon:
Merged into the [version-3.x branch](https://github.com/ampedandwired/html-webpack-plugin/tree/version-3.x)

PR #573, opened 2017-01-29
[v3] Remove html minification

SpaceK33z:
As discussed in #558, this removes html minification.


jantimon:
Merged into the [version-3-x branch](https://github.com/ampedandwired/html-webpack-plugin/tree/version-3.x)

PR #574, opened 2017-01-29
chunks passed to alter-assets event

numical:
Minor bug fix to ensure `chunks` param passed to `alter-assets` event - will also allow implementation of  [style-ext-html-webpack-plugin](https://github.com/numical/style-ext-html-webpack-plugin) to be simplified.

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,7 +144,6 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       .then(function (result) {
         var html = result.html;
         var assets = result.assets;
-        var chunks = result.chunks;
         // Prepare script and link tags
         var assetTags = self.generateAssetTags(assets);
         var pluginArgs = {head: assetTags.head, body: assetTags.body, plugin: self, chunks: chunks, outputName: self.childCompilationOutputName};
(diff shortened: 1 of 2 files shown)


jantimon:
Can you please go a little bit in detail what your code change actually does?

numical:
Sure - it's a bug fix.
Each event is passed selected parts of `html-webpack-plugin` state.
So the `...before-html...` events gets passed the `assets`, the `...after-event...` event gets passed the `html` etc.
The `...alter-assets...` event is meant to be passed the `chunks` - which is quite useful.  However there is a bug - the `chunks` value which is passed is [taken from the previous event's `result`](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L147).  The previous event  is [not actually passed this value](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L141)  so `undefined` is always sent.
The fix is to simply send the [value generated by the `...alter-chunks...` event ](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L74).  This, usefully, is already is scope so that is needed is to remove [the incorrect assignment](https://github.com/ampedandwired/html-webpack-plugin/blob/master/index.js#L147).

jantimon:
Cool - thanks a lot. Also merged into 3.x

PR #575, opened 2017-01-29
Version 3.x

jantimon:
See #558


package.json
```diff
@@ -56,13 +56,15 @@
   },
   "dependencies": {
     "bluebird": "^3.4.7",
```
wtgtybhertgeghgtwtg: Aren't promises native in the required versions?

jantimon: Right now the code depends on http://bluebirdjs.com/docs/api/promise.promisify.html

wtgtybhertgeghgtwtg: Would you accept a PR to change that?

PR #583, opened 2017-02-05
Infer link element media attribute from CSS filenames

### LINKED ISSUE #582: Add support for setting the media attribute on stylesheet link tags
There is a `media` [link element attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Attributes) that hints to the browser if it should download the CSS file. It would be awesome if there was a way to tell this plugin to add that information to the injected element.

I was thinking that perhaps grabbing this information in Base64 from the filename could be a possibility. Like if you use the ExtractTextPlugin and set the output filename to be:

```
"print" -> base64 -> "cHJpbnQ="

style.media_cHJpbnQ=.css

becomes:

<link rel="stylesheet" src="style.media_cHJpbnQ=.css" media="print" />
```

A more complicated example:

```
"(min-width: 700px), handheld and (orientation: landscape)" -> base64 -> "KG1pbi13aWR0aDogNzAwcHgpLCBoYW5kaGVsZCBhbmQgKG9yaWVudGF0aW9uOiBsYW5kc2NhcGUp"

style.media_KG1pbi13aWR0aDogNzAwcHgpLCBoYW5kaGVsZCBhbmQgKG9yaWVudGF0aW9uOiBsYW5kc2NhcGUp.css

becomes:

<link rel="stylesheet" src="style.media_KG1pbi13aWR0aDogNzAwcHgpLCBoYW5kaGVsZCBhbmQgKG9yaWVudGF0aW9uOiBsYW5kc2NhcGUp.css" media="(min-width: 700px), handheld and (orientation: landscape)" />
```

Base64 is good for this use case because of the parens and other characters that could break the filename in different file systems.

yaycmyk:
This allows for things like print stylesheets and desktop or mobile-only styles to be isolated and conditionally loaded by the browser.

Closes #582

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /node_modules/
 /dist/
 npm-debug.log
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -75,6 +75,16 @@ If you have any CSS assets in webpack's output (for example, CSS extracted
 with the [ExtractTextPlugin](https://github.com/webpack/extract-text-webpack-plugin))
 then these will be included with `<link>` tags in the HTML head.
 
+If you want to automatically set the `<link>` element [`media` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#Attributes), ensure the extracted CSS file name includes "media_{base64MediaString}":
+
+```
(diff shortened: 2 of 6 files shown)


jantimon:
Wow cool 👍 but this is a little bit away from the standard features - please move it to a plugin for the html-webpack-plugin.

Take https://github.com/jantimon/resource-hints-webpack-plugin as example it does almost exactly the same.

yaycmyk:
@jantimon done :) https://github.com/yaycmyk/link-media-html-webpack-plugin

PR #584, opened 2017-02-05
Upgrade to webpack 2 syntax

quarthex:


jantimon:
This branch will be for webpack 1 - but you could upgrade the following branch:

https://github.com/jantimon/html-webpack-plugin/tree/version-3.x

PR #601, opened 2017-02-24
Adds type attribute to the generated link elements for CSS

dimitarivanov:
Although the rel element is set to stylesheet the created link element does not specify the MIME type of the linked resource. With this change the MIME type is set to text/css.
There are tools/frameworks that rely on MIME type for processing static content prior delivering the resources. In this particular case the href tag is not updated correctly because the type attribute is missing.


jantimon:
Could you please move that feature it into a plugin?

PR #634, opened 2017-04-02
Inject with custom placeholder

ccqgithub:
html:
```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>title</title>
</head>
<body>
  <div id="app"></div>

  <!-- html-webpack-plugin-body -->

  <div></div>
</body>
</html>

```

webpack.conf.js
```
new HtmlWebpackPlugin({
      template: './src/html/index.html',
      filename: 'index.html',
      bodyReplaceExp: /<!-- html-webpack-plugin-body -->/,
    }),
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -521,8 +521,8 @@ HtmlWebpackPlugin.prototype.generateAssetTags = function (assets) {
  */
 HtmlWebpackPlugin.prototype.injectAssetsIntoHtml = function (html, assets, assetTags) {
   var htmlRegExp = /(<html[^>]*>)/i;
-  var headRegExp = /(<\/head>)/i;
-  var bodyRegExp = /(<\/body>)/i;
+  var headRegExp = this.options.headReplaceExp || /(<\/head>)/i;
+  var bodyRegExp = this.options.bodyReplaceExp || /(<\/body>)/i;
   var body = assetTags.body.map(this.createHtmlTag);
   var head = assetTags.head.map(this.createHtmlTag);
 


ccqgithub:
❓

jantimon:
Thanks for your pr but this can already be achieved with `inject:false`.

ccqgithub:
@jantimon if i use `inject:false` and not use `js template like ejs` but with `html-loader`, how can i inject the bundles at expected positions?

i think it is much better to have a `placeholder` tell plugin where to inject css and where to inject js,  so i don't need to use ejs template but `html-loader`..

like: 
```html
<body>
  <script src="http://cdn.com/es5-shim.js"></script>

  <!-- html-webpack-plugin-css -->

  <script>
    console.log('complete')
  </script>
</body>
</html>
```

ccqgithub:
@jantimon  and some times, i don not want to use `ejs` in build process, because i will use the output html for `ejs  render` in node server, there wil confilict....

jantimon:
You can use any template language - just use another loader :)

ccqgithub:
oh...

but can u add a new feature like this for the plugin ? i think it is very usefull feature in many cases ...

now, i use a fork plugin to do this, but i very except that i can use the feature in this plugin, because the plugin has much more other usefull feature for me ~~~~~

ccqgithub:
in many multiple pages app cases, we do not want to use template, because every page is so diffrent, we just want inject bundles auto ..

ccqgithub:
@jantimon i look the source code quickly, i found i can right another plugin like `html-webpack-include-assets-plugin` to work with `html-webpack-plugin`.

is it the right away ?

PR #639, opened 2017-04-08
Option to inject variables to template context

smokku:
When using `html-loader` with `interpolate` option, the returned template is a function, not a plain string. This function is then run by html-webpack-plugin, but there is no way of passing any useful values to it. `require` is the only useful function injected currently.

This change adds `variables` option to configuration, which provides variables and values that the template function will have access to and can for example iterate over generating template code.


jantimon:
Sorry I would rather not build a work around for a missing feature in the html-loader.
It's not possible to use the html-loader that way in your webpack apps is it?

PR #643, opened 2017-04-17
doc(addons): add modules-cdn-webpack-plugin

mastilver:


jantimon:
Can you please rebase and merge it?

jantimon:
Just merge it directly if it still works with webpack 4

PR #644, opened 2017-04-18
Allow disabling DOCTYPE and title completely

vp2177:


mastilver:
I don't think this is a good idea, as the default should not contains much logic (`template` option is here if you want to provide a custom template)

Anyway if you want your pr to have more chance to be merged, test need to be green:
`jasmine spec/BasicSpec.js spec/CachingSpec.js` has 35 failures on your pr

PR #661, opened 2017-05-04
Add template params at evaluation time

### LINKED ISSUE #615: How does it know the hashed file names to use?
Given that the output file names are going to be different on every compilation, now does this thing know what names to inject into the resulting html file?

### LINKED ISSUE #617: Assets are not available in template processor context when using custom template processor
When you choose not to use `inject:true`, you'd opt to iterate over `htmlWebpackPlugin.assets` and manually insert them into your template.

However, when your templates are `swig` this is not possible because that template variable is not available.

I tried making a plugin that responds to the `htmlWebpackPlugin` events, but the html generated from the initial build resulted in an empty `htmlWebpackPlugin` variable.

JSteunou:
This fixes missing template params for loader like `html-loader` which return a String and not a callable.

fixes #639 
fixes #617 
fixes #615

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,7 +144,6 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       .then(function (result) {
         var html = result.html;
         var assets = result.assets;
-        var chunks = result.chunks;
         // Prepare script and link tags
         var assetTags = self.generateAssetTags(assets);
         var pluginArgs = {head: assetTags.head, body: assetTags.body, plugin: self, chunks: chunks, outputName: self.childCompilationOutputName};
--- a/index.js
+++ b/index.js
@@ -114,7 +114,6 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
         }
         // Once everything is compiled evaluate the html factory
         // and replace it with its content
-        console.log('evaluateCompilationResult', chunks, assets);
         return self.evaluateCompilationResult(compilation, compiledTemplate, assets);
       })
       // Allow plugins to make changes to the assets before invoking the template
@@ -131,7 +130,6 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       })
       // Execute the template
       .then(function (compilationResult) {
-        console.log('executeTemplate', chunks, assets);
         // If the loader result is a function execute it to retrieve the html
         // otherwise use the returned html
         return typeof compilationResult !== 'function'
--- a/index.js
+++ b/index.js
@@ -114,7 +114,8 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
         }
         // Once everything is compiled evaluate the html factory
         // and replace it with its content
-        return self.evaluateCompilationResult(compilation, compiledTemplate);
+        console.log('evaluateCompilationResult', chunks, assets);
+        return self.evaluateCompilationResult(compilation, compiledTemplate, assets);
       })
       // Allow plugins to make changes to the assets before invoking the template
       // This only makes sense to use if `inject` is `false`
@@ -130,6 +131,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       })
       // Execute the template
       .then(function (compilationResult) {
+        console.log('executeTemplate', chunks, assets);
         // If the loader result is a function execute it to retrieve the html
         // otherwise use the returned html
         return typeof compilationResult !== 'function'
@@ -144,6 +146,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
       .then(function (result) {
         var html = result.html;
         var assets = result.assets;
+        var chunks = result.chunks;
         // Prepare script and link tags
         var assetTags = self.generateAssetTags(assets);
         var pluginArgs = {head: assetTags.head, body: assetTags.body, plugin: self, chunks: chunks, outputName: self.childCompilationOutputName};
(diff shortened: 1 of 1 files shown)


JSteunou:
@jantimon I think I'll need some help with CI tests. Is my PR broken or just all tests?

mastilver:
The test are broken right now

As long as `jasmine spec/BasicSpec.js spec/CachingSpec.js` passes it's fine

JSteunou:
Nice, could this PR be validated then?

jantimon:
Could you please explain why this is necessary? 

The code already allows to access the assets here:

https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L137

JSteunou:
I dont like to play captain obvious but the line of code you quoted is a
ternary conditional : this method is called only if the template is a
function that can be called with those parameters

jantimon:
Yes that's how webpack and template loaders work - a template is transformed into a template function and this function can be called by your script (or in this case by the html-webpack-plugin).

Can you please give any code example where this pr might be useful?

JSteunou:
I disagree, it is not always a template function to be called. Look at here https://github.com/JSteunou/html-webpack-plugin/blob/d7217edc0b9a3d0591d5ce3678f100695e4e80f2/index.js#L117
This promise resolve either on a String or a Function. (that's why if have a condition at loc 137)

It is the case for example with the `html-loader`. It returns directly a string, that's why the only way is to give in parameters as a context for the vm here https://github.com/JSteunou/html-webpack-plugin/blob/d7217edc0b9a3d0591d5ce3678f100695e4e80f2/index.js#L239

As for the use case:
1. like I said, `html-loader`, simple usage, like describe in the README https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates
2. if you look at the related issues I linked above I might have more use cases.

mastilver:
Maybe I didn't understand it well, but I don't see the real benefice of that PR either

If the default loader that `html-loader`, I would completely understand your issue, but here the default loader is `ejs-loader` which allow parameters

What is your use case for using `html-loader`?

Can you for example, give us the `index.html` you would use?

JSteunou:
It is quite the same, you can actually use the `html-loader` as a template engine as it can evaluate ES6 template literals. For example

```html
${htmlWebpackPlugin.files.js.map(path => `<link rel="preload" type="text/javascript" href="${path}" as="script">`).join('\n')}
```

This allow me to write my own `link` nodes, but I need to have access to `htmlWebpackPlugin` but with the actual version you can access it only when using template engine that return a function.

That's the issue also raised by #639 & #617 you did not understand.

JSteunou:
So, what do you think guys?

jantimon:
Sorry we won't mess with the globals just to support this edge case.
Please use a `.js` file as entry point to get your case working:

```
template.js

module.exports = function(templateParams) {
  global.templateParams = templateParams;
  return require('index.html');
}
```

JSteunou:
I see what you mean, but I cannot get it works. I tried 

```js
            new HtmlWebpackPlugin({
                template: 'src/index.html.js',
                chunks: *'polyfill', 'common', 'app'],
                inject: false
            }),
```
with `index.html.js` a wrapper around the template file, but I go this error

```
/home/jerome/project/node_modules/webpack/lib/Compilation.js:269
                if(_this.profile) {
                        ^

TypeError: Cannot read property 'profile' of null
    at factoryCallback (/home/jerome/project/node_modules/webpack/lib/Compilation.js:269:13)
    at factory (/home/jerome/project/node_modules/webpack/lib/NormalModuleFactory.js:253:5)
    at applyPluginsAsyncWaterfall (/home/jerome/project/node_modules/webpack/lib/NormalModuleFactory.js:99:14)
    at /home/jerome/project/node_modules/tapable/lib/Tapable.js:204:11
    at NormalModuleFactory.params.normalModuleFactory.plugin (/home/jerome/project/node_modules/webpack/lib/CompatibilityPlugin.js:52:5)
    at NormalModuleFactory.applyPluginsAsyncWaterfall (/home/jerome/project/node_modules/tapable/lib/Tapable.js:208:13)
    at resolver (/home/jerome/project/node_modules/webpack/lib/NormalModuleFactory.js:74:11)
    at process.nextTick (/home/jerome/project/node_modules/webpack/lib/NormalModuleFactory.js:205:8)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

JSteunou:
Ok I have been on this for too long, sorry but that's a huge waist of time for me and my team and I dont see how can this work. So 

1. it does not pollute global, it's just the context of execution of the template see https://github.com/JSteunou/html-webpack-plugin/blob/d7217edc0b9a3d0591d5ce3678f100695e4e80f2/index.js#L239
2. if you do not accept this solution, maybe you or someone could come with a better solution to automatically wrap all sort of template that do not return a function but a string into a wrapper function. I do not think this sort of code should be inside each developer project, otherwise what's the point to write plugin over cookbooks.
3. if still dont want to solve this issue then you can close all related issue (I'm not the only one impacted) and update your readme, because as of today the statement about being able to write it's own template thanks to `htmlWebpackPlugin` variable being present is not true for all type of template engine.

mastilver:
A few things:
- A project maintainer has every rights to reject a feature into its code-base as they will be the one maintaining that feature on the long term
- Sorry for wasting your time, I won't spend any more of my own time  to try to find a workaround for you then
- Nothing is preventing you from using this PR on your project

JSteunou:
* sure, and I accept it very well as long as the explanation is not clunky. I'm also a maintainer, I also reject PR.
* I consider this a fix to an issue, a bug, a real one, not a workaround
* We are already using an internal fork as this was a serious bug for us

jantimon:
Sorry JSteunou you tried to use this plugin on a way it was not intended.
Your change is not a bug but rather a feature I would rather not like to maintain.
The way you explained would also fail when you use the html-loader in your project as it has only access on global objects. If you insist on using template literals over ejs/hbs/twig... you could easily write a loader which returns a function.

JSteunou:
template string is a feature from [html-loader](https://github.com/webpack-contrib/html-loader#interpolation) I won't write a loader over a loader, I prefer make this one works for all cases.

PR #666, opened 2017-05-10
docs(README): clearer explanation about minify

kud:
As read like this, I thought I had to import html-minifier and give the object as parameter like:

```javascript
import htmlMinifier from 'html-minifier'

new HtmlWebpackPlugin({
        template: 'src/index.html',
        minify: htmlMinifier,
      }),
```


kud:
Thank you :)

PR #674, opened 2017-05-15
Add meta initial-scale=1 for mobile device

lwdgit:


jantimon:
Sorry but this would break many projects. Please use a custom template :)

Issue #675, opened 2017-05-15
silently ignores "chunks" option when provided a string


stephenmathieson:
the outputted HTML given:

```
new HtmlWebpackPlugin({
  chunks: 'options',
  filename: 'options.html'
}),
```

contains all chunks, but this only contains "options":

```
new HtmlWebpackPlugin({
  chunks: [ 'options' ],
  filename: 'options.html'
}),
```

imo the plugin should warn (or something?) when provided arguments it doesn't understand

mastilver:
@jantimon Should we add a schema?

`webpack` is doing it that way: https://github.com/webpack/webpack/blob/83ad4b0f356c290f8ad1575563b033ee2831b23c/lib/webpack.js#L17-L20

I don't mind working on it if it's something you want

jantimon:
That's really a smart idea :)

You might want to take a look at https://www.npmjs.com/package/ajv and https://jsonschema.net/#/editor

Issue #481, opened 2016-11-02
Wrong chunks order also with chunksSortMode: 'none'


Ks89:
I want to create two html pages based on ejs templates, injecting chunks into the body tag.

Pages are created and everything seems to be ok, but the order of chunks is wrong.

If I use chunksSortMode:'auto' for both files -> the first is ok (polyfills->vendor->app), but the second one no (admin->polyfills->vendor)

If I use chunksSortMode:'none' for both files -> both of them are wrong. The first one is (vendor->app->polyfills), the second one (vendor->admin->polyfills)

For the first file (with app) I can use 'auto', but for the second one, I'm obliged to create a crazy-custom-order function [p..-v...-a... (not alphabetical order)] (that i don't want to post here, because it's something the no one should see :( ).

```
new HtmlWebpackPlugin({
      title: TITLE,
      inject: true,
      chunksSortMode: 'auto', //ok only with auto
      chunks: ['polyfills', 'vendor', 'app'],
      template: TEMPLATE_PATH,
      filename: TEMPLATE_HTML
    }),
    new HtmlWebpackPlugin({
      title: TITLE_ADMIN,
      inject: true,
      chunksSortMode: 'none' //wrong with every value -> I'm using a crazy order function (very bad)
      chunks: ['polyfills', 'vendor', 'admin'],
      template: TEMPLATE_ADMIN_PATH,
      filename: TEMPLATE_ADMIN_HTML
    }),
```

The expected behaviour should be this:
- chunksSortMode: 'none', chunks: ['polyfills', 'vendor', 'admin']
<body>
....polyfills...
....vendor...
....admin...
</body>
- chunksSortMode: 'none', chunks: ['polyfills', 'vendor', 'app']
<body>
....polyfills...
....vendor...
....app...
</body>

Every other behaviour with chunksSortMode: 'none' it is weird.

PS: I created my chunks in this way (probably, the cause of this issue is here...I really don't know):
```
entry: {
    polyfills: './src/polyfills.ts',
    vendor: './src/vendor.ts',
    app: './src/main.ts',
    admin: './src/admin.ts'
  },
........
new CommonsChunkPlugin({
      name: ['admin', 'app', 'vendor', 'polyfills'],
      minChunks: Infinity
    }),
``` 

I used: Nodejs 7.0.0, macOS Sierra, Webpack 2.1.0 beta 25, HtmlWebpackPlugin: 2.24.1

Thank u,
Ks89

[...]

PR #684, opened 2017-05-23
Update index.js

skiny2015:
Return includedChunks(In order of the includedChunks) from the compilation result


mastilver:
Can you explain the purpose of this code?
did you run into an issue?

skiny2015:
Sorry, I didn't explain it before。
When I create multiple pages, the JS order does not follow the specified chunks，I didn't do it with chunksSortMode，So I added a piece of code, and once you specify chunks, you output it in the order chunks。

mastilver:
I don't get why you can't use `chunksSortMode`
You should rely on `filter` to order something

skiny2015:
**My configuration:**
{
    new HtmlWebpackPlugin({
        filename: config.build.assetsRoot + '/pages/console.html',
        template: path.join(__dirname, '.', '../webapp/pages/console.html'),
        inject: true,
        chunks: ['manifest', 'public', 'react', 'console']
    }),
    new HtmlWebpackPlugin({
        filename: config.build.assetsRoot + '/pages/console2.html',
        template: path.join(__dirname, '.', '../webapp/pages/console2.html'),
        inject: true,
        chunks: ['manifest', 'public', 'vue', 'console2']
    })
}
**The result:**
page1:
<script type=text/javascript src=/static/js/console.02dbdc8063b95b346fd6.js></script>
<script type=text/javascript src=/static/js/vue.70601a24c5d766eff054.js></script>
<script type=text/javascript src=/static/js/manifest.78756d18087a5643b1eb.js></script>
<script type=text/javascript src=/static/js/public.588b4a534ea8d7df4c3a.js></script>
page2:
<script type=text/javascript src=/static/js/manifest.78756d18087a5643b1eb.js></script>
<script type=text/javascript src=/static/js/console2.a006b793ea9b9ed33756.js></script>
<script type=text/javascript src=/static/js/react.0351cf0b6ac2c7836187.js></script>
<script type=text/javascript src=/static/js/public.588b4a534ea8d7df4c3a.js></script>
The correct order should be the chunks I configured.
I can't sort correctly using chunksSortMode.  So I added a piece of code, as long as the configuration of the chunks, filter out in accordance with the results of the chunks order.  I don't know if I have a solution. If so, please help me

mastilver:
@skiny2015 Maybe this can help: https://github.com/jantimon/html-webpack-plugin/issues/481#issuecomment-262414169

skiny2015:
I know this method, I used, but only one page is valid, many pages do not work.  My goal is to have pages configured with chunks output in the order of chunks, and I don't know if there are any other ways.

jantimon:
This feature was requested several times - maybe we can modify it a bit to work in a nonbreaking way and introduce it for 3.x

skiny2015:
I expect you to provide the function of specifying the chunks sort.

skiny2015:
I think the code I added is able to solve the order of the specified chunks. I look for it in the order of chunks, and then output the correct result. Of course, you may think more carefully, and my purpose is only to solve my current problems, I hope you can improve this function.

mastilver:
@skiny2015 I may solve it, but I don't think it's the right way

You should edit: https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L332

Add a new `sortMode`, maybe `manual` ( @jantimon , does that sound good?)

When in that mode, just return the array as is

-----
TODO:
- [ ] add tests
- [ ] update documentation

skiny2015:
I did have the option of adding sortMode before, but I found that none had been achieved, so there was no added option

jantimon:
Manuel sounds quite nice 👍 

Also please don't add it to a function which is called filter as your code is not filter releated at all

skiny2015:
Ok.

mastilver:
> I did have the option of adding sortMode before, but I found that none had been achieved, so there was no added option

Yeah, sorry I edited my comment, I initially though `none` would do the job, but it only take webpack output (which has already been altered) and returns it

`manual` need to looks more deeply into webpack stats to make `html-webpack-plugin`' `chunk`, map with either `entries` or `entrypoints` (Those are just pointers, I need to dig more when I have more time)

skiny2015:
I modified my code,  and add a new sortMode. I've tested it on my project.

mastilver:
Would you mind adding some tests like: https://github.com/jantimon/html-webpack-plugin/blob/9b272cfb2bccee98bbe8438c817c4207a0a3cb71/spec/BasicSpec.js#L1380

Also can you update the readme?

I will do a code review later

@jantimon Do you want this to land on `v3.x`?

skiny2015:
Ok, I will

skiny2015:
I failed to add a example,  please tell me how to add the example in webpack2.x.

mastilver:
run:
- `npm i --no-save webpack@2`
- `npm run build-examples`
- `npm i`

jantimon:
What do you think - should we throw an error if chunks is empty and sort mode is manuel?

skiny2015:
Should output empty if  chunks is empty when the sort mode is manual, We can output a warning, but not an exception. What do you think?

mastilver:
@jantimon Sounds like something that should be handle by: https://github.com/jantimon/html-webpack-plugin/issues/675

I still need to work on it, I traveling to France this w-e so I will have plenty of spare time on the boat, I will give it a go

jantimon:
Okay cool

kbulis:
When is the next release? and will it include #481 (new manual chunksSortMode)?

PR #693, opened 2017-06-02
updated syntax

EsrefDurna:
passed all semistandard rules, 
loading lodash fully is unnecessary, its better to use by module, and its better to switch to es6 in the feature.
+no var,
+no extra space,
+no else return,

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,17 +1,17 @@
 'use strict';
-var vm = require('vm');
-var fs = require('fs');
-var _ = require('lodash');
-var Promise = require('bluebird');
-var path = require('path');
-var childCompiler = require('./lib/compiler.js');
-var prettyError = require('./lib/errors.js');
-var chunkSorter = require('./lib/chunksorter.js');
+let vm = require('vm');
+let fs = require('fs');
+let extend = require('lodash/extend');
+let uniq = require('lodash/uniq');
+let Promise = require('bluebird');
+let path = require('path');
+let childCompiler = require('./lib/compiler.js');
+let prettyError = require('./lib/errors.js');
+let chunkSorter = require('./lib/chunksorter.js');
 Promise.promisifyAll(fs);
-
 function HtmlWebpackPlugin (options) {
   // Default options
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,17 +1,17 @@
...
+let path = require('path');
+let childCompiler = require('./lib/compiler.js');
+let prettyError = require('./lib/errors.js');
+let chunkSorter = require('./lib/chunksorter.js');
 Promise.promisifyAll(fs);
-
 function HtmlWebpackPlugin (options) {
   // Default options
-  this.options = _.extend({
+  this.options = extend({
```
mastilver: use `Object.assign` here

index.js
```diff
@@ -616,36 +576,33 @@ HtmlWebpackPlugin.prototype.getFullTemplatePath = function (template, context) {
...
-  var promisedApplyPluginsAsyncWaterfall = Promise.promisify(compilation.applyPluginsAsyncWaterfall, {context: compilation});
+  let promisedApplyPluginsAsyncWaterfall = Promise.promisify(compilation.applyPluginsAsyncWaterfall, { context: compilation });
   return function (eventName, requiresResult, pluginArgs) {
     return promisedApplyPluginsAsyncWaterfall(eventName, pluginArgs)
       .then(function (result) {
         if (requiresResult && !result) {
           compilation.warnings.push(new Error('Using ' + eventName + ' without returning a result is deprecated.'));
         }
-        return _.extend(pluginArgs, result);
+        return extend(pluginArgs, result);
```
mastilver: You **probably** can use `Object.assign` here as well

mastilver:
Thank you for your work, I added a couple of comments that will remove `lodash` entirely on that file

Could you also go through the other files?

Anyway I can't guaranty this will go in as it's a lot of churn change that could easily lead to merge conflicts, will need to check with @jantimon

jantimon:
Sorry this won't work with node 0.10 the 3.x branch is already using es6

PR #698, opened 2017-06-07
Adding option injectCss

Mjesson:
If false css is not added to head element, user must add it by another way in ejs template for example


jantimon:
Wouldn't it be easier just to turn off inject?

Mjesson:
No, because turn off inject, stop all injections include js, the purpose here is just for the css <link>

mastilver:
@jantimon I keep thinking that having `inject` as a function will solve a lot of issues for many users

https://github.com/jantimon/html-webpack-plugin/pull/692

jantimon:
#692 adds a lot of complexity - maybe we can get around with a simpler solution which was proposed a while ago.

Right now we provide the generated `<script>` and `<link>` tags only if inject is activated but we should provide them also to the templates so it would be a lot of simpler to write the templates with inject off

AndrewCraswell:
If we provide an option for `injectCSS` to be `false`, shouldn't there also be an option for `injectJs: false`? I actually have the opposite problem -- there's a chunk with JS and CSS and I want PageA to render bother the JS and CSS, while Page2 only requires the CSS.

Issue #626, opened 2017-03-28
using CDN URL to bundle


sarkistlt:
Right now module pre-populate template like:
~~~js
<script type="text/javascript" src="/cms-frontend/common-code_90a8a39d1bf80ad3c9bb.js">
~~~

is it anyway to do something like:
~~~js
<script type="text/javascript" src="<%= htmlWebpackPlugin.options.CDN %>/cms-frontend/common-code_90a8a39d1bf80ad3c9bb.js">
~~~
?

sarkistlt:
solved with `<base href="<%= htmlWebpackPlugin.options.CDN %>">`

PR #701, opened 2017-06-10
add prefix to make things easy for s3 stuff

pronan:
If you upload your js and css files to the cloud after compilation and you will wish the html file `src` and `href` to be prefixed with the given string. e.g. `<script type="text/javascript" src="http://cdn.example.com/js/main.js">` or `<link href="http://cdn.example.com/css/main.css" rel="stylesheet">`. These lines enable this by an option `prefix`. It's totally harmless if you don't set this option.


jantimon:
Couldn't you use public path for that?

pronan:
Embarassing. Thanks.

ysyfff:
there is another way to sovle this, see #626

PR #703, opened 2017-06-16
Add the option "templateArgs" in order to pass custom arguments to the template.

yveskaufmann:
This PR add the option `templateArgs` which enables users to pass custom arguments to the template.

Currently, this is only possible when you pass these arguments inside the options of `HtmlWebpackPlugin` but this leads to long illegible template argument references inside the templates.  In addition, this behavior is not documented, so I think it is a not intended feature.  Therefore, I think that a dedicated options for this use case is desirable.


jantimon:
This was already possible before

Issue #708, opened 2017-06-20
Incorrect generated html 


mzbac:
## Description
Bug, when inject to head or body the generated html incorrect.
example:
when index.html valid template shown below : 
```html
<!DOCTYPE html>
<html lang="en" >
<head >
     <title>test</title >
</head >

<body >
<div id="main" ></div >
</body >
</html >
```
It has space in the closing tag, it's still a valid html. but when we tried to generate index.html by that template. we got generated html as below: 
```html 
 <!DOCTYPE html>
<html lang="en" ><head><link href="/bundle-ad4e0941ef4abe82cd73.css" rel="stylesheet"></head>
<head >
     <title>test</title >
</head >

<body >
<div id="main" ></div >
</body >
</html ><script type="text/javascript" src="/bundle-ad4e0941ef4abe82cd73.js"></script>
```
which is incorrecrt html. the reason it injected assets in wrong position I believe is the head and body regex which doesn't conside about spaces in the html tags:

https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L524

jantimon:
Do you really think it is worth fixing and maintaining such an edge case?

pekaaw:
There is a regex looking for `/(<\/body>)/i`. If no such expression is found, the script tag is added at the end.  
The problem seems to be the last space in your tags (`</body >` instead of `</body>`).  
The same goes true for your `</head >` tag. It is not found because of the unexpected space and thus it is created right after the `<html>` tag and the `<link>` tag is inserted there.  
@mzbac : Remove your unnecessary spaces and it will work as expected =)

mzbac:
@jantimon @pekaaw  the reason I got this edge case because one of eslint rule which suggest adding space before closing tag.
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-space-before-closing.md

so my webstorm IDE refomat hmtl template with space before closing tag. I spent almost a day try to find out why html-webpack-plugin doesn't work..

PR #709, opened 2017-06-20
Change headReg and bodyReg to allow spaces

mzbac:
This PR aim to solve the issue:https://github.com/jantimon/html-webpack-plugin/issues/708


jantimon:
Code looks good - thanks @mzbac  👍 

@mastilver What do you think should we merge and maintain this edge case or rather accept that it won't work for `</      body      >`?

If we merge it we should also add a test.

mastilver:
We shouldn't allow `</ body>` as it's invalid html, only `</body >`

@mzbac Could you add a test please

mzbac:
@mastilver you are right, space should be only added before the closing tag. I will try to fix the regex and add the test.

PR #712, opened 2017-06-20
Minor change to html output to keep body end tag at a separate line.

pekaaw:
Instead of the output 
```html
  <body>
  <script type="text/javascript" src="main.js"></script></body>
```
we get
```html
  <body>
  <script type="text/javascript" src="main.js"></script>
  </body>
```


jantimon:
Maybe we could run html tidy in a plugin

PR #723, opened 2017-06-29
Uses a singleton cache to store the compilation stats

scinos:
This PR introduces a singleton cache for the compilation stats.

In large Webpack builds generating the stats takes a while and each generated html pays that cost three times. Moreover, if the build uses multiple instances of HtmlWebpackPlugin to generate more than one html file, this cost is payed for each template.

In my local case, I have ~2000 modules and 18 html templates, which normally takes around 175 seconds. With this change, the time goes down to 148 seconds, a 15% improvement.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,14 @@ var prettyError = require('./lib/errors.js');
 var chunkSorter = require('./lib/chunksorter.js');
 Promise.promisifyAll(fs);
 
+var getStats = (function () {
+  var cachedStats = null;
+  return function (compilation) {
+    cachedStats = cachedStats || compilation.getStats().toJson();
+    return cachedStats;
+  };
+}());
+
 function HtmlWebpackPlugin (options) {
   // Default options
   this.options = _.extend({
@@ -65,7 +73,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   compiler.plugin('emit', function (compilation, callback) {
     var applyPluginsAsyncWaterfall = self.applyPluginsAsyncWaterfall(compilation);
     // Get all chunks
-    var allChunks = compilation.getStats().toJson().chunks;
+    var allChunks = getStats(compilation).chunks;
     // Filter chunks (options.chunks and options.excludeCHunks)
     var chunks = self.filterChunks(allChunks, self.options.chunks, self.options.excludeChunks);
     // Sort chunks
@@ -252,7 +260,7 @@ HtmlWebpackPlugin.prototype.executeTemplate = function (templateFunction, chunks
     .then(function () {
       var templateParams = {
         compilation: compilation,
-        webpack: compilation.getStats().toJson(),
+        webpack: getStats(compilation),
         webpackConfig: compilation.options,
         htmlWebpackPlugin: {
           files: assets,
@@ -384,7 +392,7 @@ HtmlWebpackPlugin.prototype.isHotUpdateCompilation = function (assets) {
 
 HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chunks) {
   var self = this;
-  var webpackStatsJson = compilation.getStats().toJson();
+  var webpackStatsJson = getStats(compilation);
 
   // Use the configured public path or build a relative path
   var publicPath = typeof compilation.options.output.publicPath !== 'undefined'
--- a/index.js
+++ b/index.js
@@ -360,7 +360,11 @@ HtmlWebpackPlugin.prototype.filterChunks = function (chunks, includedChunks, exc
       return false;
     }
     // Skip if the chunk should be lazy loaded
-    if (!chunk.initial) {
+    if (typeof chunk.isInitial === 'function') {
+      if (!chunk.isInitial()) {
+        return false;
+      }
+    } else if (!chunk.initial) {
       return false;
     }
     // Skip if the chunks should be filtered and the given chunk was not added explicity
--- a/index.js
+++ b/index.js
@@ -66,7 +66,6 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
     var applyPluginsAsyncWaterfall = self.applyPluginsAsyncWaterfall(compilation);
     // Get all chunks
     var allChunks = compilation.chunks;
-    debugger;
     // Filter chunks (options.chunks and options.excludeCHunks)
     var chunks = self.filterChunks(allChunks, self.options.chunks, self.options.excludeChunks);
     // Sort chunks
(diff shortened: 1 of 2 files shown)


jantimon:
Interesting- if to json is that expensive we might also check if we need it at all

scinos:
For reference, this is what `toJSON()` does (https://github.com/webpack/webpack/blob/master/lib/Stats.js#L71). tldr: a lot.

index.js
```diff
@@ -65,7 +75,7 @@ HtmlWebpackPlugin.prototype.apply = function (compiler) {
   compiler.plugin('emit', function (compilation, callback) {
     var applyPluginsAsyncWaterfall = self.applyPluginsAsyncWaterfall(compilation);
     // Get all chunks
-    var allChunks = compilation.getStats().toJson().chunks;
+    var allChunks = getStats(compilation).chunks;
```
mastilver: Can be replaced by `compilation.chunks`

index.js
```diff
@@ -384,7 +394,7 @@ HtmlWebpackPlugin.prototype.isHotUpdateCompilation = function (assets) {
 
 HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chunks) {
   var self = this;
-  var webpackStatsJson = compilation.getStats().toJson();
+  var webpackStatsJson = getStats(compilation);
```
mastilver: only need to get the hash, which we can get by doing `compilation.hash`

index.js
```diff
@@ -252,7 +262,7 @@ HtmlWebpackPlugin.prototype.executeTemplate = function (templateFunction, chunks
     .then(function () {
       var templateParams = {
         compilation: compilation,
-        webpack: compilation.getStats().toJson(),
+        webpack: getStats(compilation),
```
mastilver: I **think** this is equivalent to just passing `compilation`

I believe, the only difference between `compilation.getStats().toJson()` and `compilation` is that `compilation` have webpack types while `compilation.getStats().toJson()` gets us js native types

Not 100% sure on this one thought

mastilver:
@jantimon I think you are right here, we don't need `toJSON()`, see my comments on the code

index.js
scinos: From the doc, this should be `compilation.getStats()`, right?

> The following variables are available in the template:
>    * webpack: the webpack stats object.
(source https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates)

index.js
scinos: Changing this will change the api of `chunksSortMode`. Before it received two objects with an array property `names`. Now they will have a string property `name`. Example:

```diff
diff --git a/spec/BasicSpec.js b/spec/BasicSpec.js
index 27ea085..093f85d 100644
--- a/spec/BasicSpec.js
+++ b/spec/BasicSpec.js
@@ -1414,10 +1414,10 @@ describe('HtmlWebpackPlugin', function () {
       plugins: [
         new HtmlWebpackPlugin({
           chunksSortMode: function (a, b) {
-            if (a.names[0] < b.names[0]) {
+            if (a.name < b.name) {
               return 1;
             }
-            if (a.names[0] > b.names[0]) {
+            if (a.name > b.name) {
               return -1;
             }
             return 0;
```

Is not documented in the docs (https://github.com/jantimon/html-webpack-plugin#writing-your-own-templates), but it could break some uses. Are you ok with this change?

mastilver: Yeah probably not :D

I was wrong here. So is everything still works with `getStats()` then without the need for `toJSON()`?

index.js
mastilver: Yep you are right :+1:

index.js
scinos: Actually a few properties will change:

Before
```
{ id: 0,
  rendered: true,
  initial: true,
  entry: true,
  extraAsync: false,
  size: 165,
  names: [ 'main' ],
  files: [ 'bundle.js', 'styles.css' ],
  hash: '6db016c6e99e5efcf9e6',
  parents: [],
  origins: [ [Object] ] }
```

After
```
Chunk {
  id: 0,
  ids: [ 0 ],
  name: 'main',
  modules: [ [Object], [Object] ],
  chunks: [],
  parents: [],
  blocks: [],
  origins: [ [Object] ],
  files: [ 'bundle.js', 'styles.css' ],
  rendered: true,
  entry: true,
  initial: true,
  hash: '6db016c6e99e5efcf9e636c868282a89',
  renderedHash: '6db016c6e99e5efcf9e6' }
```

scinos:
I have removed the calls to `toJSON()` and everything seems to work (i.e. all tests are passing).

The only problem I see is that using `compilation.chunks` instead of `compilation.getStats().toJson().chunks` will change the parameters passed to the function `chunksSortMode` (see above comments for details).

scinos:
@mastilver Sorry, I missed your reply above (I blame github notifications).

So the problem is the change

```diff
   compiler.plugin('emit', function (compilation, callback) {
     var applyPluginsAsyncWaterfall = self.applyPluginsAsyncWaterfall(compilation);
     // Get all chunks
-    var allChunks = compilation.getStats().toJson().chunks;
+    var allChunks = getStats(compilation).chunks;
```

If we replace it with `compilation.chunks` it will change the API of the method `chunksSortMode`, and `compilation.getStats().chunks` doesn't exist. 

So our options are:
  * Use a cached `compilation.getStats().toJson()` (i.e. the original change)
  * Use `compilation.chunks` and break `chunksSortMode` API. By exposing a 
  * Do not change it. After all, the plugin is already 3x faster because we removed two calls to `compilation.getStats().toJson()`

Thoughts?

jantimon:
I would prefer not to introduce the cache as it is really hard to understand for other contributors.

scinos:
@jantimon so this is ready to be merged for 2.x, or do you want me to change something else?

jantimon:
The template option is a breaking change so we would rather merge it in 3.x

scinos:
@jantimon I removed the change that affected the `chunksSortMode` option, is that the breaking change?

What about merging now the current change (removes 2 out of 3 usages of `toJSON()`, no breaking changes), and merge the other one in 3.x?

Alternatively, what is the ETA for 3.x? Is there anything I can help with?

scinos:
What is the ETA for release a version of the plugin with this change?

cyqresig:
When will the fixed version be published?

gregjacobs:
Has this changed been published yet? I'm experiencing ~25 second rebuild times for 6 html files (project is quite large: ~5200 modules and 10 entry modules)

asolove:
I'm not the most knowledgable, but I'm running into this same problem with performance issues so I tried spelunking through the code history.

I think the performance improvements made by this PR were erased by the change in https://github.com/jantimon/html-webpack-plugin/commit/0cf580c628a495d77b85e35ed1b9d244e2ec5ae0 to fix a bug that this PR introduced. But the bugfix goes back to generating the full stats for every html file, when I think it _might_ only need a small subset. Unclear. But in any event, I wonder if someone can speak to whether it's possible to not generate the full stats json every time while still generating the data needed for templates.

PR #736, opened 2017-07-16
[WIP] Improve performance by reducing unnecessary calls to stats.

asolove:
Getting the full build stats can be very expensive, and has to happen for each child compiler. So let's avoid it in cases where we just need specific data from the compilation.

- Get hash straight off compilation
- Get chunks off compilation (and update some differences between actual chunk objects and chunk stats)

We have a [webpack stress test](https://github.com/TheLarkInn/webpack-stress-test/pull/3) using the multipage plugin (which uses this plugin). In that test, removing these stats serializations reduced the runtime from 38s to 26s on my machine.


asolove:
Not sure why the tests pass locally and fail on CI, but I apparently opened this too hastily and need to look into it more.

mastilver:
#723 duplicate

PR #737, opened 2017-07-16
default sortMode should be 'dependency' on webpack3

lxjwlt:
Default sortMode should be 'dependency' on webpack3, otherwise the order of chunks would be mess In some cases.


jantimon:
We will drop a default sorting entirely now that webpack 4 supports unsorted chunks:

PR #744, opened 2017-07-19
Allow CSS links to be placed in the head or body using an `injectCss` property

### LINKED ISSUE #741: Is there a way to inject stylesheets into the `body` tag?
Recently, I have met a confusing problem in my project when I have used a Webpack plugin named `extract-text-webpack-plugin` to extract my stylesheet into one file. But with using `html-webpack-plugin`, I can not find out a proper way to avoid blocking in rendering DOM elements by injecting such a stylesheet file into the body tag, rather than the head. Can you tell me how can I do this?

chrisdeely:
This is a slight enhancement to #698 which allows users to configure CSS ```<link>``` tags to be placed in the ```<head>```, ```<body>``` or ignored altogether if desired.

Fixes #741


mastilver:
Hi @chrisdeely 

Would you mind adding some tests please

jantimon:
Wouldn't this cause a fouc?

chrisdeely:
@jantimon it certainly could if used improperly. In my use case the critical styles are loaded into the head and CSS which is not needed at startup is placed in the body. In any case, the default setting is to place the styles in the head.

PR #746, opened 2017-07-20
add dll-link-webpack-plugin to readme

clinyong:


mastilver:
I would expect Dlls to be added to your output html file, no?

Maybe I don't understand what this plugin is really doing

clinyong:
Hi @mastilver 

I added an [example](https://github.com/clinyong/dll-link-webpack-plugin/tree/master/examples/html-webpack-plugin) of using with html-webpack-plugin. 

This plugin helps you to create dll file. And you can use it with or without html-webpack-plugin. But if you uses html-webpack-plugin, it can inject the dll file to the output html file.

mastilver:
So you are saying that it doesn't work with ` webpack.DllReferencePlugin` ?

clinyong:
Yes, the `DllReferencePlugin` must be replaced.

PR #769, opened 2017-08-26
Test for the js extension before creating the script tag

aeyoll:
I think it can be a good thing to check if the generated script tag is containing a js file (it is done for css files a few lines up).

### DIFF
--- a/index.js
+++ b/index.js
@@ -471,7 +471,10 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
  */
 HtmlWebpackPlugin.prototype.generateAssetTags = function (assets) {
   // Turn script files into script tags
-  var scripts = assets.js.map(function (scriptPath) {
+  var js = assets.js.filter(function (scriptPath) {
+    return /.js($|\?)/.test(scriptPath);
+  });
+  var scripts = js.map(function (scriptPath) {
     return {
       tagName: 'script',
       closeTag: true,


jantimon:
I guess that's not necessary - the css is part of a collection which might contain also images but the js is part of a collection which contains only javascript.

aeyoll:
In fact, I had the case in one of my projects where I had a css file present into the javascript collection (and the css collection too actually), generating a script tag with the css file. I haven't figured out why it's happening yet, but I thought it would be a good idea to do an extra test in this plugin.

jantimon:
We have millions of downloads every month and this is the first time I hear about that.  
If you can add a valid show case which causes such bugs I am open to merge it :)

aeyoll:
Here is my test case: https://github.com/aeyoll/html-webpack-plugin-pull-769-test-case. Clone the repo, and launch a ```npm run prod ``` to see the problem. You should get a ```web/build/frontend/index.html``` file looking like this:

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>Webpack App</title>
  <link href="/build/frontend/global.3107a352f88fb560b7985d1a4e01f7d5.css" rel="stylesheet"></head>
  <body>
  <script type="text/javascript" src="/build/frontend/manifest.d41d8cd98f00b204e980.js">
</script>
  <script type="text/javascript" src="/build/frontend/vendor.798e377370bd8f125984.js"></script>
  <script type="text/javascript" src="/build/frontend/app.c408edaf21151fe898e3.js"></script>
  <script type="text/javascript" src="/build/frontend/global.3107a352f88fb560b7985d1a4e01f7d5.css"></script></body>
</html>
```

My webpack setup is using symfony/webpack-encore, which is a wrapper around webpack. The use case provided will dump the generated webpack configuration when you'll run the ```npm run prod``` command, so you can see what's going on under the hood.

PR #776, opened 2017-08-31
refactor: replace bluebird with native promise

### LINKED ISSUE #772: Refactor request: replace Bluebird promises with native promises.
## Description

This is a feature/refactor request.

`html-webpack-plugin` currently uses Bluebird promises, which use `try...catch`. Such blocks are a known de-optimization (https://github.com/GoogleChrome/devtools-docs/issues/53#issuecomment-175904780) in some versions of V8, like the one present in Node 6 (current LTS). Node 6 also has native promise support though.

Thus using Bluebird promises can cause decreased performance. Below are a few screenshots of a Webpack rebuild, showing the deopt is occurring (all within the `innerArrayForEach` group):

image
image

I tried replacing this specific use of Promise (it's the one in `html-webpack-plugin/lib/compiler.js (compileTemplate)` with a native promise and I observed roughly 200ms savings on a rebuild in that function group:
image

This doesn't mean that all builds would benefit from the same reduction, but since rebuild performance is of high value perhaps this is a worthwhile refactor.

It's worth mentioning that in the V8 engine 6.0 this isn't relevant anymore: https://www.nearform.com/blog/node-js-is-getting-a-new-v8-with-turbofan. But that might be a ways off for a lot of users.

## Environment
```
Node.js v6.11.1
win32 10.0.15063
npm 5.3.0
webpack@3.5.5
html-webpack-plugin@2.30.1
```

filipesilva:
Fix #772


index.js
```diff
@@ -304,22 +318,24 @@ HtmlWebpackPlugin.prototype.postProcessHtml = function (html, assets, assetTags)
  */
 HtmlWebpackPlugin.prototype.addFileToAssets = function (filename, compilation) {
   filename = path.resolve(compilation.compiler.context, filename);
-  return Promise.props({
-    size: fs.statAsync(filename),
-    source: fs.readFileAsync(filename)
-  })
+  return Promise.all([
```
mastilver: I'm fine this that, but **if** we want to keep the same, we could use: https://github.com/sindresorhus/p-props

index.js
```diff
@@ -2,12 +2,28 @@
...
 var fs = require('fs');
 var _ = require('lodash');
-var Promise = require('bluebird');
 var path = require('path');
 var childCompiler = require('./lib/compiler.js');
 var prettyError = require('./lib/errors.js');
 var chunkSorter = require('./lib/chunksorter.js');
-Promise.promisifyAll(fs);
+
+function statAsync (filename) {
```
mastilver: Should we use: https://github.com/normalize/mz ?

filipesilva: I considered a couple of ways of doing it, the most straightforward I found was the native `util.promisify` (http://2ality.com/2017/05/util-promisify.html) and it's polyfill. 

I settled for just using basic promise functionality because it didn't introduce other dependencies and it seemed the most explicit way of doing it (at the cost of a bit of repetition).

index.js
filipesilva: That seems nice and small, and doesn't have any deps itself.

index.js
mastilver: > I settled for just using basic promise functionality because it didn't introduce other dependencies and it seemed the most explicit way of doing it (at the cost of a bit of repetition).

yep, I kinds of agree with you :) 
We just need to make a choice: more code or more dependencies
I will let Jan answer that

jantimon: Imho we should rather use a promisify util or mz as they provide already unit tested code and we don't have to maintain that part of code in our project. 
It looks like `mz` is not using native promises: https://github.com/normalize/mz#promise-engine

filipesilva: Added `util.promisify` and used it for `statAsync` and `readFileAsync`, but did not use it for `applyPluginsAsyncWaterfall` since it requires being called as a method of `compilation`.

jantimon:
As this is breaking we would have to merge it in 3.x

filipesilva:
What's the minimum version of node supported by `html-webpack-plugin`? Promise support (as used in the PR) seems to be available since node 0.12.18 (http://node.green/#ES2015-built-ins-Promise).

jantimon:
The plugin works with 0.10 - this is pure legacy.. we should create a 3.x beta release soon but right now I am extrem busy at work

ngyikp:
v3.0.0 no longer uses bluebird, this PR could be closed
https://github.com/jantimon/html-webpack-plugin/pull/861/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L46

PR #779, opened 2017-09-09
Test stub for webpackDevServer

BurkovBA:
This PR started with a bug that I encountered a week ago.

I observed, that when I tried to set `publicPath=/static/`, both `index.html` and app.js were successfully generated by html-webpack-plugin and webpack and put to a proper folder. However, when I was trying to serve them with webpackDevServer, index.html was served, while `/static/app.js` was not.

So, I added a small test stub to make sure that html-webpack-plugin works correctly not only with webpack, but with webpackDevServer as well.

Unfortunately, in this test callback inside `server.listen()` does not get executed (and test always passes successfully). Something wrong with Jasmine. Could you suggest an edit that would fix that?

Also, I have some bug with formatting of output file for Default example in Webpack 2 (probably some indents or something).

Also, note that I updated webpack to version 2, cause I used webackDevServer 2 as well.

Thanks!


BurkovBA:
Hi, @jantimon! 

Ugh, currently I can't guarantee that. I'll have to update this for webpackDevServer 3 that is compatible with webpack 4, I suppose.

Thanks for reminder!

jantimon:
Cool thanks 👍

BurkovBA:
@jantimon Hi, jantimon. 

Sorry about delays - I got stuck, learning how to work this tappable and new webpack-cli - I don't quite understand how to use them yet. Should have notified you earlier.

PR #786, opened 2017-09-18
Removed full lodash library import

aduryagin:


mastilver:
Thank you for you PR
Although I don't think I'm going to merge it

But I'm sure we can get rid of `lodash` all together on version 3 (as we will only be supporting node 4 or 6)

aiduryagin:
Well, we will wait for the 3rd version)

lukeed:
This could easily be a patch release! So tired of seeing all these messages in my build:

```
[BABEL] Note: The code generator has deoptimised the styling of "~/node_modules/lodash/lodash.js" as it exceeds the max of "500KB".
```

PR #787, opened 2017-09-21
Add compress-webpack-plugin

genu:


mastilver:
Hi @genu 

Can you explain me, that's the difference between your plugin and https://github.com/webpack-contrib/compression-webpack-plugin ?

genu:
I believe that one is unmaintained, and it I believe is more general as it doesn't hook into this plugin's events.

I essentially forked that plugin, and made it specifically listen to the event fired from this plugin so that it also updates the asset references in the target html file.

d3viant0ne:
It is & has been maintained by Webpack

mastilver:
@genu I'm closing as I think we should support the official plugin

If you think things should be improved, I would suggest you submit your ideas to the compression-webpack-plugin and contribute there :)

PR #789, opened 2017-09-28
Reference lodash loader by default

nphoff:
Hello!

It came to my attention that the default loader was being referred to as a ejs loader, when in fact it is a loadsh loader. The syntaxes for both look almost identical, which made it harder to realize what the difference was.

I'm not super fond of the wording I came up with to modify the docs, so I'd be open to other suggestions.

Let me know if this looks good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,12 +156,12 @@ and favicon files into the markup.
 plugins: [
   new HtmlWebpackPlugin({
     title: 'Custom template',
-    template: 'my-index.ejs', // Load a custom template (ejs by default see the FAQ for details)
+    template: 'my-index.html', // Load a custom template (lodash by default see the FAQ for details)
   })
 ]
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -156,12 +156,12 @@ and favicon files into the markup.
 plugins: [
   new HtmlWebpackPlugin({
     title: 'Custom template',
-    template: 'my-index.ejs', // Load a custom template (ejs by default see the FAQ for details)
+    template: 'my-index.html', // Load a custom template (lodash by default see the FAQ for details)
```
mastilver: I think that would be better, we get so many issues just because they are using the `.html` ext and don't understand what's happening (and tbw most shouldn't care)

mastilver:
Great changes, thank you! :)

Feel free to submit more of those goodies, we need more people focusing on docs!

PR #794, opened 2017-10-04
Add a Git .mailmap with merged repeated occurrences of various people

mgol:
In this way my past contribution is mapped correctly.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,8 @@
+Chris Greening <chris@cmgresearch.com>
+Jan Nicklas <j.nicklas@me.com>
+Jan Nicklas <j.nicklas@me.com> <j.nicklas@sinnerschrader.com>
+Ma Cheng <mc@mc-zone.me>
+Michał Gołębiowski-Owczarek <m.goleb@gmail.com>
+Thomas Grainger <tom.grainger@procensus.com>
+Thomas Grainger <tom.grainger@procensus.com> <tagrain@gmail.com>
+Zhanwei Wu <zhanwei.wu@dbjtech.com>


mastilver:
Hi @mgol 

Thank you for your previous contribution

Would you mind if there is other peoples that could be included in that file, like multiple contributors that have different commit profile

mgol:
@mastilver PR updated. I added entries for all the people for which I found repeated entries (I searched by last name & email). I sorted the entries alphabetically.

Does that look good now?

jantimon:
I am not sure if we really want to maintain such a file - is it worth the effort?

mastilver:
As a maintainer, I'm not really a big fan of it
As a contributor, I can see the (limited) appeal but I wouldn't care about it myself

mgol:
It doesn't require extensive maintenance; an entry in the file is not required for anyone and doesn't do anything for contributors with only one signature & email in the commit history. All it allows is to optionally squash some author entries into one. I'd expect people that want to have their entries squashed to submit PRs with their own entries (though I don't think this would happen very often).

In any case, if you don't want to add the file then fine, it's not a critical thing for me. Some people (e.g. trans folks) may care about it more (which is probably why my casual observation is projects maintained by trans people have `.mailmap` more often), it also affects women more than men as they change their last name more often.

mastilver:
> Some people (e.g. trans folks) may care about it more (which is probably why my casual observation is projects maintained by trans people have .mailmap more often), it also affects women more than men as they change their last name more often.

Thank you for bringing those up. We definitely want everybody to feel comfortable contributing to this project and if you feel it could be an obstacle for some people than yes I agree we should merge that

jantimon:
So is it up to date for merging?

mgol:
@mastilver 
> if you feel it could be an obstacle

It *might* be but that's just my impression based on why heard. I just wanted to raise such a possibility but I don't want to speak for others in absolute terms.

That said, just to reiterate, it's absolutely not critical to me personally. If you'd rather not maintain this file you may as well reject the PR & wait for someone for whom it's important to submit such a PR (at which point we could add entries I added in this PR as well).

mgol:
@jantimon I was just about to reply to you that there is one new entry for @mastilver that I added (he used a different mail for his last commit, I mapped it to the previous one which he still has on his GitHub page), but I wasn't on time before your merge. :) See #903.

Issue #726, opened 2017-06-30
No "type" attribute in generated "link" elements for stylesheets


jarkkorantala:
## Description

When generating the link elements for CSS stylesheets, the "type" attribute is omitted.

Although not mandatory as per HTML 4.01 or 5 specification, it provides valuable information for the browser in case the HTTP server serves the CSS files without a Content-Type header.

At least with Internet Explorer 11, a link element with no type and a reference to a file served without a Content-Type header raises error "SEC7113: CSS was ignored due to mime type mismatch".

### Error Message & Stack Trace

Internet Explorer 11:
[SEC7113](https://msdn.microsoft.com/query/dev12.query?appId=Dev12IDEF1&l=EN-GB&k=k(VS.WebClient.Help.SEC7113)): CSS was ignored due to mime type mismatch

### Config

Copy the relevant section from `webpack.config.js`:

```js
module.exports = {
    entry: {
        app: __dirname + '/src/app/core/index.module.js',
        vendor: __dirname + '/src/vendor.lib.js',
    },
    output: {
        path: __dirname + "/dist",
        filename: '[name].bundle.js',
        chunkFilename: '[name].bundle.js',
        sourceMapFilename :'[name].bundle.js.map'
    },
    plugins: [
        new ExtractTextPlugin(
            "[name].css"
        ),
        new HtmlWebpackPlugin({
            filename: 'index.html',
            template: 'src/index.html',
            chunksSortMode: orderByList(['vendor', 'app']),
            title: 'Tree-shaking'
        })
    ],
    module: {
        loaders: [{
            test: /\.html$/,
            loader: "html-loader",
            exclude: ['src/index.html']
        }, {
            test: /\.css$/,
            loader: ExtractTextPlugin.extract({fallback: "style-loader", use: "css-loader"})
        }]
    },
    devtool: 'source-map',
    devServer: {
        contentBase: __dirname + "/src"
    }
}
```

Copy your template file:

```html
<!doctype html>
<html ng-app="myApp">
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Site title</title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width">
  </head>
  <body>
    <div top-bar></div>
    <div nav-bar></div>
    <div ui-view></div>
  </body>
</html>
```

## Environment

[...]

PR #795, opened 2017-10-08
fix ie11 css caching

seva-ask:
fix for #726

### DIFF
--- a/index.js
+++ b/index.js
@@ -490,7 +490,8 @@ HtmlWebpackPlugin.prototype.generateAssetTags = function (assets) {
       selfClosingTag: selfClosingTag,
       attributes: {
         href: stylePath,
-        rel: 'stylesheet'
+        rel: 'stylesheet',
+        type: 'text/css'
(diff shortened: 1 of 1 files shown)


mastilver:
Hi @seva-ask thank you for the PR

Just need to check with @jantimon if it's really something we need... For two reason:
- it's fixable by adding `Content-Type` on the server side
- it's not on the specs

mastilver:
Oh, and also would you mind adding test ;)

richriscunha:
@mastilver,

I'm curious to know where this stands, if you could please advise.

jantimon:
Please move this feature into a plugin.
The problem happens for js

vudduu:
The commit also will need changes on unit tests to be applied.

For example next unit test will fail.
https://github.com/jantimon/html-webpack-plugin/blob/efbd3525a18bee03a0c8c3899d264a14f2945d85/spec/BasicSpec.js#L479

Issue #113, opened 2015-11-07
config option for adding `async` attribute to script tag?


jeberly:
Hi, 

I didn't see anything in the source or docs about a config option to add the `async` attribute to the script tags that get injected, am I missing something?

http://www.w3schools.com/tags/att_script_async.asp

Making the script load async is usually recommended after running https://developers.google.com/speed/pagespeed/insights/

I am happy to submit a PR, but just want to check first.

Thanks for the lib!

vyorkin:
:+1: pls don't forget about `defer`.
for now I just use custom template

tvervest:
:+1: Loving the plugin! And I'd love it even more if this was a feature ;)

duncanbeevers:
I don't think it's safe to add the `async` attribute to these script tags as their execution order matters.

For example, when I generate a bundle with two entry points, only one of them provides the `window.webpackJsonp` global callback. Subsequent chunks depend on that function in order to register their modules.

That could be overcome by only omitting the `defer` attribute on the first script tag but including it on subsequent tags, but this seems to be a path toward some complexity, rather than just turing a `defer` option on or off for the whole plugin.

alextreppass:
I'd really love to see `async` and `defer` options for those who know what they're doing, and want to fine-tune their page load.

I imagine the config could map to `config.entry` keys, e.g. given the following webpack config:
```javascript
  entry: {
    app: ['./app/bootstrap.js'],
    vendor: ['./app/vendor.js'],
    thirdparty: ['./app/third-party.js']
  },
```

The plugin config could look something like:
```javascript
  new HtmlWebpackPlugin({
    filename: 'app.html',
    template: 'html!./src/app.html',
    inject: 'body',
    async: ['app', 'vendor'],
    defer: ['third-party']
  })
```

jantimon:
@alextreppass Unfortunately there are more then just asnyc and defer..
![Async](https://pbs.twimg.com/media/CZMJBnDWwAA0VVW.jpg:large)

Maybe a more future proof solution could be

```js
new HtmlWebpackPlugin({
  inject: {
    default: 'body',
    entries: {
       app: 'defer',
       vendor: 'module',
       thirdparty: 'head'
    }
  }
});
```

what do you think?

numical:
I have gone down a slightly different route with [script-ext-html-webpack-plugin](https://github.com/numical/script-ext-html-webpack-plugin).
No `module` support yet but I will look at that next...

[...]

PR #796, opened 2017-10-12
to improve load asynchrony on the screen

billyrtc13:
based on issue #113
async is not feasible
but module greatly improves performance


mastilver:
Hi @billyrtc13 

I'm sorry, unless I don't understand your issue this can already be resolved by using https://github.com/numical/script-ext-html-webpack-plugin

Let me know if I'm wrong in any way

Issue #724, opened 2017-06-29
Multiple entry points -> multiple html outputs webpack rebuild  very slow


CharlieLau:
## my  question  is:

- in my project, i has 70's entries and htmls page. when one entry changed, html-webpack-plugin
rebuild  very slowly, Has anyone ever met ？

cyqresig:
I met this, too. Rebuilding is rather slow using html-webpack-plugin@2+ on webpack 2 or 3.
I try to downgrade it to 1.7.0, then everything is ok.
But I think it is a serious bug that need to be fixed as soon as possible.

mastilver:
Could you try if it gets better this: https://github.com/jantimon/html-webpack-plugin/pull/723

cyqresig:
I try the pr code and it works, but I found that:
1. Rebuilding of this has returned to normal, but it is still a little slower than it of version 1.7.0,
2. Initial building of this is almost 30% slower than it of version 1.7.0, and it is almost 25% faster than it of version 2.X.

mastilver:
@cyqresig thank you for you feedback, we will make sure to improve that!

CharlieLau:
@mastilver  I try the pr code,but  it  didn't work .
   I used "webpack multicompile " to resolve it

simon-yxl:
I met this, too.

githoniel:
me two.
when there is 10 entry, rebuild take about 8000ms. i think it's acceptable
when I add to 24 entry, rebuild take about 40000ms.......

jantimon:
Is there a difference between webpack 1 and webpack 3? 
Because we have similar issues here: https://github.com/webpack/webpack/issues/5505#issuecomment-322979832

githoniel:
yes i think it's because of webpack3.
I switch to webpack@2, rebuild is only 300ms.
```
[0] webpack: Compiling...
[0] webpack: wait until bundle finished: /chunk.2.js
[0] webpack building...
[0]  DONE  Compiled successfully in 3001ms14:45:37
[0]
[0] webpack built 0ee20b12b0698f93c3a9 in 3001ms
[0] Hash: 0ee20b12b0698f93c3a9
[0] Version: webpack 2.7.0
[0] Time: 3001ms
```
notice that [0] webpack: wait until bundle finished: /chunk.2.js.
It seem that webpack2 only rebuild chunk2.
but in webpack3，there is a full rebuild.

M-torch99:
[...]

Issue #680, opened 2017-05-18
HTML files are always emitted HMR


moimael:
Hi,

Using this plugin with HMR cause html files to be emitted all the time, even if there are no content change.
I'm using the cache option.
It adds ~3.5s to recompile in HMR mode. If I remove the plugin, HMR is much quicker. (1.9s vs 5.5s)

EDIT: Formatting change were caused by my atom editor. If I diff the files nothing changed but they are still emitted again with caching enabled.

Here is the relevant conf:

```
    ...Object.keys(apps())
      .map(app => new HtmlWebpackPlugin({
        title: app,
        filename: `${app}.html`,
        chunks: ['manifest', 'vendor', `${app}_vendor`, app],
        cache: true,
        minify: {
          collapseWhitespace: true,
        },
        inject: false,
        template: HtmlWebpackTemplate,
        appMountId: 'app',
        mobile: true,
        meta: [
          {
            title: `${app} | Universal Avenue`,
          },
          {
            'http-equiv': 'Accept-CH',
            content: 'DPR, Viewport-Width, Width',
          },
        ],
        scripts: [
          '/sw-registration.js',
        ],
        window: {
          NODE_ENV: process.env.NODE_ENV,
          API_BASE_PATH: process.env.API_BASE_PATH,
          CLIENT_API_BASE_PATH: process.env.CLIENT_API_BASE_PATH,
          SENTRY_CLIENT_DSN: process.env.SENTRY_CLIENT_DSN,
        },
        alwaysWriteToDisk: true,
      })),
    new HtmlWebpackHarddiskPlugin(),
```

Seems related to #244

mastilver:
Yes, we need to improve the caching mechanism, PR welcome :)

moimael:
In #263 it seems like there i no interest in fixing it. Otherwise we could just use PR #244 ?

jantimon:
@moimael is that a underlying webpack bug?

moimael:
Not sure, I mean all other asset works well with hot reload and recompile only what's changed. I only have issue with this plugin. The cache does not seems to work as intended.

nkoterba:
Any update to this issue since it's been 5 months now?  

I've noticed that my index.html is *always* emitted.  Is this normal behavior if my other assets change and they are injected into my index.html?

My `favicon.ico` is also always emitted.  I'm trying to reduce emitted assets to only those items that change to reduce development time.

gkatsanos:
Any updates?

[...]

PR #797, opened 2017-10-12
fix "Multiple entry points -> multiple html outputs webpack rebuild very slow #724"

daifee:
# issues 724
https://github.com/jantimon/html-webpack-plugin/issues/724

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 HTML Webpack Plugin
 ===================
+
+add a configuration options (`multihtmlCache`) to solve multihtml recompile slow
+
 [![npm version](https://badge.fury.io/js/html-webpack-plugin.svg)](http://badge.fury.io/js/html-webpack-plugin) [![Dependency Status](https://david-dm.org/jantimon/html-webpack-plugin.svg)](https://david-dm.org/jantimon/html-webpack-plugin) [![Build status](https://travis-ci.org/jantimon/html-webpack-plugin.svg)](https://travis-ci.org/jantimon/html-webpack-plugin) [![Windows build status](https://ci.appveyor.com/api/projects/status/github/jantimon/html-webpack-plugin?svg=true&branch=master)](https://ci.appveyor.com/project/jantimon/html-webpack-plugin) [![js-semistandard-style](https://img.shields.io/badge/code%20style-semistandard-brightgreen.svg?style=flat-square)](https://github.com/Flet/semistandard) [![bitHound Dependencies](https://www.bithound.io/github/jantimon/html-webpack-plugin/badges/dependencies.svg)](https://www.bithound.io/github/jantimon/html-webpack-plugin/master/dependencies/npm) [![license](https://img.shields.io/github/license/mashape/apistatus.svg?maxAge=2592000)]()
 
 [![NPM](https://nodei.co/npm/html-webpack-plugin.png?downloads=true&downloadRank=true&stars=true)](https://nodei.co/npm/html-webpack-plugin/)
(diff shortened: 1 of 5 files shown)


mastilver:
Thank you for having a go at this issue :)

But I think the best way to solve that issue would be that `html-webpack-plugin` takes an array of config

So rather that instantiate multiple plugins, we would only have one. How does that sounds to you?

Also tests would be awesome! ;)

daifee:
"html-webpack-plugin takes an array of config" can not solve this issue, The root of this issue is cache.

when a file has been changed, `compiler` will recompile and trigger `plugin('make', cb)`, then `html-webpack-plugin` will create new `childCompiler`, and run `compile` for all html. it is slow.

you should avoid create new `childCompiler`, when the html has not been changed.

daifee:
can you merge the pr, and publish a feature fix version?

loopArray:
这个确实减少了很多的编译时间，之前11个页面更新其中的一个就需要全部编译，花40s左右，现在基本上6s就行了，但是还有一个问题，就是我更新了index.pug仍然会产生很多次的编译。不知道是不是配置的问题。

This is really a lot of compilation time, before 11 pages to update one of them need to compile all, spend 40s or so, now basically 6s on the line, but there is a problem that I updated the index.pug will still produce a lot Compile time. Do not know if the configuration is not the problem.

image

kidneyleung:
@loopArray It is recommended to send configuration, my project test did not appear this situation

loopArray:
My fault, just debugging again, I want to find a place to cry for a while! Traversed again when the local template file, nested loops two times, and this bug actually always existed, I did not find and I was looking for reasons why html-webpack-plugin multiple html so slow.
image

jinker:
@loopArray 我也遇到类似问题，你可以看看这个pr可不可以解决 #802

mastilver:
> @loopArray 我也遇到类似问题，你可以看看这个pr可不可以解决 #802

@jinker I don't mind a few words in a foreign language (just to say hi ;) ), but please keep the conversation in English please

mastilver:
@daifee I'm closing this in flavor of #802 as I find the code simpler and it doesn't add a new variable

Let me know if you disagree! :)

daifee:
@jinker @mastilver  i also tried to do so, and it only reduce 50% time(at my project, recompilation takes 20s).

at my project, the task of compile template is very complex, so `childCompiler.compileTemplate()` is slow. multiple pages will take long time.

the template has not been changed, it should not be recompilated.

mastilver:
@daifee Correct me if I'm wrong but this doesn't to do with multiple pages... It's a caching issue (#680) and it's just because you have multiple pages you get hit by that more heavily than others

daifee:
@mastilver I think so, my project is more heavily. i has forked html-webpack-plugin, and published html-webpack-plugin-for-multihtml to solve my issue.

daifee:
@mastilver Thanks your feedback.

limerickgds:
@mastilver can you publish new version? i also have this issue

xilin:
@daifee Thanks for your work! Using your fork make our project's build time reduce from 10s -> 0.8s. 

Could you share the profile experience on how to locate the time consuming part in webpack building? I tried to use `--progress` and ` --profile`, but the only thing I can tell is that `asset optimization` phase is slow.

qiming-kooboo:
@daifee 
OMG... I really want to say love you so much...
This issue distracting our dev for a long time. 
Finally, it is resolved by your Repo.
Our project rebuild time reduced from 30+s => 1s..

daifee:
@xilin view the source code of html-webpack-plugin.

PR #802, opened 2017-10-27
Cache chunks result between multiple entry point html page

jinker:
In some time the method `compilation.getStats().toJson()`  take long.
At my project, it takes about 1 to 3s, and I have 155 pages, it will take 155 to 460s total, too slow.
Cache the chunks result between entry points, it will reduce a lot of time.
``` js
    //...
    // Get all chunks
    var allChunks = compilation.getStats().toJson().chunks;
    //....
}
```

``` js
/**
 * Cache chunks result between multiple entry point html page
 */
var hashToChunksMap = {};
//...
    //...
    // Get all chunks
    var allChunks = self.getAllChunks(compilation);
    //....
}
//....
/**
 * Get all chunks from compilation
 * @param compilation
 * @return {Array}
 */
HtmlWebpackPlugin.prototype.getAllChunks = function (compilation) {
  var allChunks = hashToChunksMap[compilation.hash];
  if (!allChunks) {
    allChunks = compilation.getStats().toJson().chunks;
    hashToChunksMap[compilation.hash] = allChunks;
  }
  return allChunks;
};
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -665,7 +665,7 @@ HtmlWebpackPlugin.prototype.applyPluginsAsyncWaterfall = function (compilation) 
 HtmlWebpackPlugin.prototype.getAllChunks = function (compilation) {
   var allChunks = hashToChunksMap[compilation.hash];
   if (!allChunks) {
-    allChunks = compilation.getStats().toJSON().chunks;
+    allChunks = compilation.getStats().toJson().chunks;
     hashToChunksMap[compilation.hash] = allChunks;
   }
   return allChunks;
--- a/index.js
+++ b/index.js
@@ -668,7 +668,7 @@ HtmlWebpackPlugin.prototype.getAllChunks = function (compilation) {
     allChunks = compilation.getStats().toJSON().chunks;
     hashToChunksMap[compilation.hash] = allChunks;
   }
-  return allChunks
+  return allChunks;
 };
 
 module.exports = HtmlWebpackPlugin;
--- a/index.js
+++ b/index.js
@@ -664,7 +664,7 @@ HtmlWebpackPlugin.prototype.applyPluginsAsyncWaterfall = function (compilation) 
  */
 HtmlWebpackPlugin.prototype.getAllChunks = function (compilation) {
   var allChunks = hashToChunksMap[compilation.hash];
-  if(!allChunks){
+  if (!allChunks) {
     allChunks = compilation.getStats().toJSON().chunks;
     hashToChunksMap[compilation.hash] = allChunks;
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
```
mastilver: Can you revert the changes on this file please

index.js
```diff
@@ -652,4 +657,18 @@ HtmlWebpackPlugin.prototype.applyPluginsAsyncWaterfall = function (compilation)
   };
 };
 
+/**
+ * Get all chunks from compilation
+ * @param compilation
+ * @return {Array}
+ */
+HtmlWebpackPlugin.prototype.getAllChunks = function (compilation) {
```
mastilver: I don't think it should returns only the `chunks`, it should return the whole JSON so we can use it for https://github.com/jantimon/html-webpack-plugin/blob/aedc623c770bce51f460da6a154690064fba2417/index.js#L255

jinker: Push new opt commit, these two places are handled uniformly

package.json
jinker: Has been reverted

sdoomz:
@mastilver what is missing to merge this PR?

githoniel:
well I really need this PR, My project have 50 entries now..

gregjacobs:
Hey guys, what's the status on this? Failing test?

I'm also wondering (and I mentioned this on another issue before): How come we need to use `getStats().toJson()` at all? How come the regular object returned by `getStats()` can't be used?

index.js
```diff
@@ -652,4 +657,16 @@ HtmlWebpackPlugin.prototype.applyPluginsAsyncWaterfall = function (compilation)
   };
 };
 
+/**
+ * Get stats from compilation
+ */
+HtmlWebpackPlugin.prototype.getCompilationStats = function (compilation) {
+  var stats = compilationHashToStatsMap[compilation.hash];
+  if (!stats) {
+    stats = compilation.getStats().toJson();
```
gerges: This `toJson` call can be passed a number of `false` options for specific keys to save more time, and limit generation to only the chunk information (which is all that's used by this plugin). You can see an example of that here in another perf PR: https://github.com/jantimon/html-webpack-plugin/pull/825/files#diff-168726dbe96b3ce427e7fedce31bb0bcL67

gerges:
`getStats().toJSON()` is also explicitly called again in the template renderer, so that it could be provided to the template if needed. This causes another pretty large performance hit. There's another PR that allows a plugin parameter which prevents that call: https://github.com/jantimon/html-webpack-plugin/pull/825

jinker:
i don't know why the test failed, what can i do for this pr to merge.@mastilver

yingye:
My project have 100+ entries, it's very slow. I really need this PR, to improve pref.

PR #809, opened 2017-11-11
Refactor manual chunksSortMode

cvetanov:
Refactor manual chunksSortMode by creating a lookup object for the chunks by their name to later avoid iterating through the chunks array in a nested for loop.

Add an additional comment to the README section for manual chunkSortMode.


jantimon:
Hey - thanks for your merge request.

Names like `util` or `object` don't help much if someone tries to understand what the code is doing.

Actually I don't get what the advantage of that change would be.  
The amount of chunks is always a number smaller than 100 - so there won't be any benefits from performance optimizations.

cvetanov:
When I come to think about it again, you are right.
I got a little bit ahead of myself and this looks like premature optimization so I will close the PR.

PR #831, opened 2018-01-05
Fix build when the first chunk file is a JSON manifest

tomekwi:
Hi, thanks for the awesome plugin!

Recently our build broke after a stack upgrade. It turns out one of the chunks comes with the `files` property looking like this:

```json
[
  "manifest.json",
  "vendor.8de0754649cf8ee08d2bcefea840020d52c818d2.b971ea180336280f6e10.js",
  "vendor.8de0754649cf8ee08d2bcefea840020d52c818d2.b971ea180336280f6e10.js.map"
]
```

which leads to a wonky `<script type="test/javascript" src=".../manifest.json"></script>` being built into the HTML. This PR fixes it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -443,7 +443,11 @@ HtmlWebpackPlugin.prototype.htmlWebpackPluginAssets = function (compilation, chu
 
     // Webpack outputs an array for each chunk when using sourcemaps
     // But we need only the entry file
-    var entry = chunkFiles[0];
+    var entry = chunkFiles.filter(function (chunkFile) {
+      // Sometimes the first chunk isn’t the entry file but a JSON manifest.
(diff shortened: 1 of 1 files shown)


tomekwi:
I’ll try to come up with a minimal configuration that reproduces this so I can add it to the test suite. If anyone has a tip about how to track it down, I’ll appreciate it!

Our build is pretty complex and the change appeared out of the blue. I’m digging through to find the cause of this right now.

tomekwi:
Hey @jantimon, I’ve found the reason why `manifest.json` is being passed as a chunk and resolved it in our build. So the fix isn’t required for us anymore. Closing the PR, because our original problem was elsewhere – but if you think it’s worth keeping the updated logic anyway, feel free to reopen 🙂

PR #842, opened 2018-01-26
Provide complete WebPack config example and link WP docs

TomasHubelbauer:
I think this basic example is better because it shows exactly what minimal WebPack configuration should work with this plugin.


jantimon:
@TomasHubelbauer  could you please rebase and rename `Module` to `module` ?

TomasHubelbauer:
Someone already made basically the changes I was going for here. :-)

PR #846, opened 2018-02-02
fixes #835

### LINKED ISSUE #835: Feature Request: type="text/javascript" is not needed on script tag
Version: 2.30.1

I wasn't sure if this was a feature request or a bug, it could go either way.

On line 479 of index.js

```
// Turn script files into script tags
  var scripts = assets.js.map(function (scriptPath) {
    return {
      tagName: 'script',
      closeTag: true,
      attributes: {
479        type: 'text/javascript',
        src: scriptPath
      }
    };
  });
```
The type attribute should not be set.  This is an obsolete attribute and HTML linters will complain about it, and rightly so.  I could not find any documentation that indicated a way to prevent this.

It appears that all that is required to fix it is to remove line 479.  I guess alternatively there could be an option to disable this or it could be tied to the doctype.

labsvisual:
Changed a little bit of code to fix issue #835. Added the `scriptType` option to make it easy for the user to toggle the `type="text/javascript"` attribute.


jdanford:
Nice work, I was going to make a PR for this myself! Make sure to update the tests in the [`spec`](https://github.com/jantimon/html-webpack-plugin/tree/master/spec) directory, as they're still expecting the attribute to be present in the generated HTML.

labsvisual:
Oh, I completely forgot that! Will send in the PR by today! 😄

labsvisual:
@jdanford, quick question -- I tried to understand how the tests work, but I was really unable to. Can you guide me a little bit? I was able to make out that it dynamically generates and validates HTML based on the output of the plugin, but changing them was a little difficult. Even the smallest of the change created a failing-avalanche. 😕

jdanford:
First off, are all the tests passing for you on `upstream/master`, i.e. without your changes? When I do so, I get a bunch of failures from `CachingSpec.js` and `ExampleSpec.js`. Assuming you're having the same problem, disable/delete those files for now.

With that out of the way, you can run the tests on your branch, resulting in a lot of similar failures. Let's take a look at the simplest one:
`Expected '<script type="text/javascript"></script>' to match /^<script type="text\/javascript" src="app_bundle\.js"><\/script>$/.`

Contrary to the intention of your changes, it looks it's keeping the `type` attribute and dropping the `src` attribute, which suggests that the following chunk of code is at fault:
```js
tagAttr = _.extend(tagAttr, {
  attributes: {
    type: 'text/javascript'
  }
});
```

Since `tagAttr.attributes` is already defined, you can do away with `_.extend` altogether and just assign the `type` property directly:
```js
tagAttr.attributes.type = 'text/javascript';
```

However, if you make this change and run the tests again, the failure has not gone away but changed:
`Expected '<script src="app_bundle.js" type="text/javascript"></script>' to match /^<script type="text\/javascript" src="app_bundle\.js"><\/script>$/.`

The `src` and `type` attributes appear in the wrong order, which matches how they are defined in the code. I'll leave it up to you to solve that part, but let me know if you're still having trouble.

There's also the problem that the `scriptType` option defaults to `true`, which means that the tests will behave exactly the same as before the option was added. Given that HTML5 is the norm these days, it seems more useful for the option to default to `false`, and then you could add one or more tests with the option enabled.

Finally, I think it would be clearer to rename the option to `includeScriptType`, as `scriptType` suggests that it would be the value used in the `<script>` element's `type` attribute. Does that sound reasonable to you?

jdanford:
@labsvisual I hope I didn't write too much and scare you off! Is there anything I can clarify?

labsvisual:
@jdanford, don't worry! I just got a little caught up with work. 

Yeah, I was thinking of renaming it to `includeScriptType` but since I was in a hurry (sorry!), I forgot to change it.

I did fix the problem with the `_.extend(...)`, but thought that I will push it once the tests run fine. The problem I have with the tests is that I can't figure out what to amend  and where. The `tests` directory is a little confusing for me; I guess because I haven't really been exposed to that structure of tests (my bad!) 

I just needed a little help to figure out how the tests worked; I did read the code but got incredibly confused. Can you just put me on the right track? I will figure the rest out.

jdanford:
No worries! [BasicSpec.js](https://github.com/jantimon/html-webpack-plugin/blob/master/spec/BasicSpec.js) is pretty large, but all you need to look for are certain calls to `testHtmlPlugin(...)`. The first argument is a Webpack config, the second is an array of strings or regexes that should be present in the generated HTML, and the remaining arguments don't matter for our purposes. What you need to do is look for strings or regexes in that array that look like `<script type="text/javascript" ...` and remove the `type` attribute.

For example, take a look at the first call to `testHtmlPlugin(...)` on lines [108-115](https://github.com/jantimon/html-webpack-plugin/blob/master/spec/BasicSpec.js#L108-L115):
```javascript
testHtmlPlugin({
    entry: path.join(__dirname, 'fixtures/index.js'),
    output: {
    path: OUTPUT_DIR,
    filename: 'index_bundle.js'
    },
    plugins: [new HtmlWebpackPlugin()]
}, [/<body>[\s]*<script type="text\/javascript" src="index_bundle.js"><\/script>[\s]*<\/body>/], null, done);
```

Changing the regex from
```javascript
/<body>[\s]*<script type="text\/javascript" src="index_bundle.js"><\/script>[\s]*<\/body>/
```
to
```javascript
/<body>[\s]*<script src="index_bundle.js"><\/script>[\s]*<\/body>/
```
should do the trick – does that make sense?

labsvisual:
Ah, I understood! I did think it was somewhere there but wasn't really sure what to change. Turns out, I have to rebuild the examples with `npm run build-examples` and then run `npm run test` for the tests to pass. 

I included all the changes along with a test case for the new option. 

However, the Travis-CI build is still failing for some reason. It cleared every test on my machine. I just inspected it and it seems it's still adding the `type="text/javascript` attribute to the HTML generated (something which doesn't happen on my local machine). Any leads there?

labsvisual:
@jdanford, can we merge this?

labsvisual:
Yep, I have added that. Look at [this](https://github.com/jantimon/html-webpack-plugin/pull/846/commits/8867d3c3e22c81d4b7a21d3f5a56d6a5ad65f849) commit.

jdanford:
My bad, I should have looked harder. Anyway, this project has quite a few pending pull requests at the moment, so I'm not sure when this will actually get merged, but thanks for all your hard work!

labsvisual:
No worries! I will pick up something new from currently documented issues.

jantimon:
Can we please move the mime support to a plugin?
It’s only for servers which don’t add the mime type in combination with IE - and it happens also for css

PR #848, opened 2018-02-13
expected behavior: options.template could be a function

oschade:
Hello,

I'm expected, that the config property "template" of the html-webpack-plugin-config could also be a function delivering a string. For example the "entry"-config option of webpack can be either a string or a function.

So, to be coherent, I think this plugin script should support this too.
In my case, it enables me to make the webpack configuration be more dynamic.


jantimon:
Can you please rebase this change?

PR #856, opened 2018-02-23
Add support for injecting Normal chunks

math-nao:
This PR adds support for adding `Normal` chunks  (lazy loaded and asynchronous) as `Initial` chunks. It may improves performances at page load for some app.


math-nao:
Setting `inject` to `false` is not going to help for getting chunks hash.

PR #868, opened 2018-03-01
Speed up build

jantimon:
Based on #825 but without a new option (by now)


gerges:
Hey @jantimon thanks for this. This doesn't include the `toJson()`  provided to the template. Any plans to add that improvement as well?

jantimon:
Yes the plan is to remove stats.toJSON entirely and switch over to `compilation.entrypoints` `Entrypoint.getChunks()` and `Entrypoint.getFiles()`

PR #881, opened 2018-03-05
allow to specify regexp as included or excluded chunks

mike1808:
I write this feature because of new `splitChunks` naming in webpack 4. It makes common chunks for several chunks and names them as `chun1~chunk2`.  So, if I want to inject only chunks which have code for my `chunk1` I added this change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -369,6 +369,21 @@ class HtmlWebpackPlugin {
   }
 
   /**
+   * Helper to filter chunk
+   */
+  testChunk (chunkName, test, shouldInclude) {
+    if (Array.isArray(test) && shouldInclude === (test.indexOf(chunkName) !== -1)) {
+      return true;
+    }
+
+    if (test instanceof RegExp && shouldInclude === test.test(chunkName)) {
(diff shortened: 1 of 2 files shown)


mike1808:
also, I don't know what to do with `manual` sort when `chunks` is regexp

jantimon:
Thanks for your pull request - however we will try to solve it with the new webpack 4 api `compilation.entrypoints` `Entrypoint.getChunks()` or `Entrypoint.getFiles()`

askoretskiy:
@jantimon Is this fix you mention like https://gist.github.com/gribnoysup/68c0a9713901eced85d4d95857d6e787 ?

mike1808:
@jantimon @askoretskiy I think, he meant this https://gist.github.com/mike1808/3d2ec798ab1e734b906d4e2fd34e540c

jantimon:
Yes but we need to get rid of `const allChunks = compilation.getStats().toJson(chunkOnlyConfig).chunks;` if possible

PR #884, opened 2018-03-05
Add a note about ordering first

jacobroufa:
Per @thelarkinn's suggestion, it is necessary to order `html-webpack-plugin` first before other plugins that make use of it.


TheLarkInn:
Thank you so much!!!!!

PR #897, opened 2018-03-09
Mobile-friendly default

jakearchibald:
I guess the default template is deliberately bare, but it feels like mobile-friendly should be the default these days.

### DIFF
--- a/default_index.ejs
+++ b/default_index.ejs
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, minimum-scale=1.0">
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>


NekR:
Unfortunately I know zillion of folks who doesn't care about mobile and just ship desktop version to it. So this at least would need to have option to not include the viewport tag (also, yeah, many people don't write HTML manually, so won't just use custom template to remove the viewport ¯\_(ツ)_/¯). Not sure if those 2 groups of people "overlap" though.

jakearchibald:
Would this harm users who don't care about mobile?

NekR:
Not "don't care", but rather "let people on mobile use desktop version and zoom in an out, because we don't have money and time for mobile dev and who cares anyway when there is React Native" lol.

That minimum zoom value won't allow users to scale down 900px wide website, right?

jakearchibald:
> That minimum zoom value won't allow users to scale down 900px wide website, right?

Correct. If this is still prevalent enough that creating a custom template is too high a bar, then I guess there could be a simpler opt-out.

jantimon:
Some background info: 
https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag

We also allow to inject viewports into the html file. 
So it would be quite easy to have a plugin similar to https://github.com/GoogleChromeLabs/preload-webpack-plugin which would add the viewport tag.

We could also provide multiple base templates.

jantimon:
I would like to close it as this could be solved with: #419  
Please let me know if you see any issues here.

PR #903, opened 2018-03-22
Add a Git .mailmap entry for Thomas Sileghem

mgol:


mgol:
Re-submitted as #1067.

PR #905, opened 2018-03-22
fix(loader): Prevent foreign loaders from compile lodash

jantimon:
Fixes #786


jantimon:
Unfortunately I would only help with only remove the following warning:

```
[BABEL] Note: The code generator has deoptimised the styling of "~/node_modules/lodash/lodash.js" as it exceeds the max of "500KB".
```

PR #913, opened 2018-03-28
Add entry config to get chunks of entrypoint.

zhanfang:
The feature is to resolve  the problem of the common chunks  inserting HTML. 
For example:
webpack config
```
module.exports = {
      entry: {
          index:  './src/index/main.js'
      },
      optimization: {
        splitChunks: {
            cacheGroups: {
                commons: {
                    chunks: 'initial',
                    minChunks: 2,
                    maxInitialRequests: 5, // The default limit is too small to showcase the effect
                    minSize: 0 // This is example is too small to create commons chunks
                },
                vendor: {
                    test: /node_modules/,
                    chunks: 'initial',
                    name: 'vendor',
                    priority: 10,
                    enforce: true
                }
            }
        }
    },
    plugins: [
        new HtmlWebpackPlugin({
          // Required
          entry: 'index'
          inject: true,
          template: './src/index/index.ejs',
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ Allowed values are as follows
 
 |Name|Type|Default|Description|
 |:--:|:--:|:-----:|:----------|
+|**[`entry`](#)**|`{String}`|``|The entry to get chunks for entrypoint|
 |**[`title`](#)**|`{String}`|``|The title to use for the generated HTML document|
 |**[`filename`](#)**|`{String}`|`'index.html'`|The file to write the HTML to. Defaults to `index.html`. You can specify a subdirectory here too (eg: `assets/admin.html`)|
 |**[`template`](#)**|`{String}`|``|`webpack` require path to the template. Please see the [docs](https://github.com/jantimon/html-webpack-plugin/blob/master/docs/template-option.md) for details|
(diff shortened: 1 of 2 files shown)


jantimon:
Hi @zhanfang 

this is exactly what we are working on for the upcoming release 👍👍
We could probably use some help but it's a quite time intense refactoring.

Let us know if you would like to contribute.

zhanfang:
Hi @jantimon 
Yes, I want to contribute. The feature is very important for multipage project.

jantimon:
The current WIP branch is this one:

What needs to be done?

1. Replace the current way the assets are generated:

```js
      // Get all entry point names for this html file
      const entryNames = Array.from(compilation.entrypoints.keys());
      const filteredEntryNames = self.filterChunks(entryNames, self.options.chunks, self.options.excludeChunks);
      const sortedEntryNames = self.sortChunks(filteredEntryNames, compilation, self.options);
      // Turn the entry point names into file paths
      const assets = self.htmlWebpackPluginAssets(compilation, sortedEntryNames);
```

2. Change the sorter to sort entryPoint names instead of chunks (we should also pass the compilation here) e.g.:

```
      return chunkSorter[sortMode](entryPointNames, compilation, htmlWebpackPluginOptions);
```

3. Refactor `htmlWebpackPluginAssets`

4. Rename all hooks from `chunk` to `entry`


Especially point *3* is difficult as we have to come up with a clean API which works for all use cases which already exist but remove the data duplication.

zhanfang:
I read you the proposal you suggest, But I don't understand why you want replace chunks with entryPoint. I suppose that you want to complate multi entry once？

jantimon:
This will fix the aggressive splitting plugin issues and improve the performance

PR #926, opened 2018-04-15
Fix confusing indentation

felixrabe:


jantimon:
😄great pr - I believe we can merge it right away.  
Could you explain why it is now less confusing?

felixrabe:
It took me a while when I was reading the passage to figure out where `compilationResult` in line 170 was coming from.

Issue #543, opened 2017-01-10
Appending chunkhash to the file name instead of a hash


tomitrescak:
Hi, thank for all the great work!
I wanted to ask if it is possible to append chunkhash to the filename instead of a hash?
Currently the { hash: true } can append `file?#hash` to the file, but I'd really like to add this for all automatically generated chunks.

The idea is to generate chunks only with filename (e.g. mychunk.js) and inserting it to HTML with (mychunk.js?chunkhash).

This way I can store the builds easily in git and when deployed, users will get only recently changed files.

damianobarbati:
Yeah, absolutely +1!!!
Use case: 2 chunks, 1 for the app (frequently changed), 2 for the vendor (almost never change).
Since we append the "bundle hash", the vendor gets invalidated but it didn't change at all!

We should append the hash relative to the chunk, this way only the chunks which really changed will be reloaded by browsers.

damianobarbati:
any news on this?

damianobarbati:
Solved.
Don't use the plugin hash option because it only use the bundle [hash].
Just append [chunkhash] to your outputs.

damianobarbati:
up!

damianobarbati:
@tomitrescak did you find a solution to use chunkhash in the different chunk/assets in the html generated?

damianobarbati:
@jantimon any chance on this? :)

caub:
The only option I can see is to read the output files on disk, hash them with md5 but they are not written on disk yet at that time right?

mastilver:
I personally don't use `hash`, I use `chunkhash` on `output.filename`

caub:
Yes I tried that at the same time `filename: '[name].js?v=[chunkhash:4]'`

html-webpack-plugin generates a 


```
<script type="text/javascript" src="app.js?v=e2b6">
```

but keep the css like:

```
<link href="app.css" rel="stylesheet">
```


webpack config:

```js
module.exports = {
    entry: {
        app: [
            './src/scripts/main.js',
            './src/style/style.scss'
        ],
        'pdf.worker': './node_modules/gravit/lib/import/pdf/pdf.worker.js',
    },
    output: {
        path: process.cwd()+'/build',
        filename: '[name].js?v=[chunkhash:4]',
        // chunkFilename: '[chunkhash].js'
    },
    node: {
        fs: 'empty',
        child_process: 'empty'
    },
    resolve: {
[...]

PR #928, opened 2018-04-18
generate a different hash for each asset

cnlevy:
### Before: all assets share the same hash
When one chunk is modified, all the emitted assets are cache-busted.

<pre>
<script type="text/javascript" src="styles-bundle.js? <b>f623fb66ba22aa937565</b>"></script>
<script type="text/javascript" src="vendors~admin~app-bundle.js?<b>f623fb66ba22aa937565</b>"></script>
<script type="text/javascript" src="app-bundle.js?<b>f623fb66ba22aa937565</b>"></script>
<script type="text/javascript" src="admin-bundle.js?<b>f623fb66ba22aa937565</b>"></script>
</pre>

### After: each emitted asset has its own hash:
When one chunk is modified, only the corresponding asset hash changes and will be cache-busted, other hashes stay the same.

<pre>
<script type="text/javascript" src="styles-bundle.js?<b>e5d2e7b1e4ffbe9bdbb1</b>"></script>
<script type="text/javascript" src="vendors~admin~app-bundle.js?<b>4e81e0a04fe7db3fe400</b>"></script>
<script type="text/javascript" src="app-bundle.js?<b>3227c67dd9b69fcfb7f6</b>"></script>
<script type="text/javascript" src="admin-bundle.js?<b>81e7e4d745085f4de14d</b>"></script>
</pre>

### DIFF
--- a/index.js
+++ b/index.js
@@ -449,7 +449,7 @@ class HtmlWebpackPlugin {
 
       // Append a hash for cache busting
       if (this.options.hash) {
-        chunkFiles = chunkFiles.map(chunkFile => self.appendHash(chunkFile, compilationHash));
+        chunkFiles = chunkFiles.map(chunkFile => self.appendHash(chunkFile, chunk.hash));
       }
 
       // Webpack outputs an array for each chunk when using sourcemaps


cnlevy:
The same result can be achieved by appending ?[chunkhash] to the output filename, like
```
output: {
    path: process.cwd()+'/build',
    filename: '[name].js?[chunkhash]',
},
```
See [issue #543 ](https://github.com/jantimon/html-webpack-plugin/issues/543#issuecomment-316978461) for details

jantimon:
I really like the idea - it would be a breaking change and therefore I would move it to our `@next branch`.

For our next branch we will restructure the way chunks are detected so I am not 100% sure that we can takeover but hopefully we can.

mastilver:
Shouldn't we drop `hash` all together? and let the users deal with that by themselves?

With this PR, I believe we will still have the same issue with Css assets

Issue #218, opened 2016-02-11
Multiple entry points -> multiple html outputs?


kirbysayshi:
Hello! I have kind of a specific use case, and was hoping to get some feedback on if html-webpack-plugin is the right tool for this job. Also, I'm relatively new to webpack so I might be missing something.

I have a webpack config similar to this:

```
module.exports = {
  entry: {
    'page1: './apps/page1/scripts/main.js',
    'page2': './apps/page2/src/main.js',
  },
  output: {
    path: __dirname,
    filename: "apps/[name]/build/bundle.js"
  }
}
```

This generates a `bundle.js` in a build folder for each "app". I tried using html-webpack-plugin to also generate an html page for each app (notice the filename placeholder similar to the `output` key):

```
plugins: [
  new HtmlWebpackPlugin({
    filename: 'apps/[name]/build/index.html'
  })
]
```

But that fails with the following error:

```
node_modules/.bin/webpack --progress -v
 70% 11/11 build modulesnode_modules/html-webpack-plugin/lib/compiler.js:86
          content: childCompilation.assets[outputOptions.filename].source()
                                                                  ^
TypeError: Cannot read property 'source' of undefined
    at node_modules/html-webpack-plugin/lib/compiler.js:86:67
    at Compiler.<anonymous> (node_modules/webpack/lib/Compiler.js:214:10)
    at node_modules/webpack/lib/Compiler.js:403:12
    at Compiler.next (node_modules/webpack/node_modules/tapable/lib/Tapable.js:67:11)
    at Compiler.<anonymous> (node_modules/webpack/lib/CachePlugin.js:40:4)
    at Compiler.applyPluginsAsync (node_modules/webpack/node_modules/tapable/lib/Tapable.js:71:13)
    at Compiler.<anonymous> (node_modules/webpack/lib/Compiler.js:400:9)
    at Compilation.<anonymous> (node_modules/webpack/lib/Compilation.js:577:13)
    at Compilation.applyPluginsAsync (node_modules/webpack/node_modules/tapable/lib/Tapable.js:60:69)
    at Compilation.<anonymous> (node_modules/webpack/lib/Compilation.js:572:10)
    at Compilation.applyPluginsAsync (node_modules/webpack/node_modules/tapable/lib/Tapable.js:60:69)
    at Compilation.<anonymous> (node_modules/webpack/lib/Compilation.js:567:9)
[...]

PR #944, opened 2018-05-04
added entryPoint option to help make multiple html pages / helps with…

umeboshi2:
… issue #218 

All this does is gets "allChunks" from the entrypoints part of stats.
"entryPoint" property is added to options, and if set to name of entry point, 
code to retrieve "allChunks" only looks at entrypoints in stats.

The idea behind this PR should work on the webpack-4 branch as well.


jantimon:
We are trying to make this the default for the next major version

Issue #878, opened 2018-03-03
🐛 Fail to inject code-splitting script files when webpack 4 optimization.splitChunks.name: false


dwiyatci:
## Description

As @sokra suggested in [one of webpack 4 posts](https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366) to preserve long term caching:

> or switch off name generation via optimization.splitChunks.name: false

`html-webpack-plugin` somehow fails to inject the generated "splitted" files when I apply this option in combination with `output: { filename: '[name].[chunkhash].js' }`.

### Config

```js
module.exports = {
  mode: 'production',
  output: {
    filename: '[name].[chunkhash].js',
  },
  plugins: [
    new HtmlWebpackPlugin({
      template: './src/index.ejs',
      favicon: './src/favicon.ico',
    }),
  ],
  optimization: {
    splitChunks: {
      chunks: 'all',
      name: false,
    },
    runtimeChunk: true,
  },
};
```

Template file:

```html
<!DOCTYPE html>
<html>
    <head>
        <title>My App</title>
    </head>
    <body>
        <div id="app"></div>
    </body>
</html>
```

## Relevant Links

- https://github.com/dwiyatci/minimalistic-devserver/blob/webpack-4/webpack.config.js

## Environment

```
Node.js v8.9.0
darwin 15.6.0
npm 5.5.1
webpack 4.0.1
html-webpack-plugin 3.0.4
```

atilkan:
+1

prencher:
This happens for us with names on as well, dependent chunks are not included.

loveky:
any update on this issue? is there a walk around or plan to fix ?

pspgbhu:
I met this problem too.

softdays:
+1

mqliutie:
+1

boraturant:
+1

dwiyatci:
@jantimon Works like a charm. :trumpet::kissing::tada: Thanks!

PR #945, opened 2018-05-05
added includeSiblingChunks for webpack4's optimization.splitChunks

zayarwinttun:
This is a quick fix for webpack4's `optimization.splitChunks` issues like https://github.com/jantimon/html-webpack-plugin/issues/878 before the webpack-4 branch is ready.

I've also added `includeSiblingChunks` option as a flag for activating this feature to avoid regression.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ Allowed values are as follows
 |**[`cache`](#)**|`{Boolean}`|`true`|Emit the file only if it was changed|
 |**[`showErrors`](#)**|`{Boolean}`|`true`|Errors details will be written into the HTML page|
 |**[`chunks`](#)**|`{?}`|`?`|Allows you to add only some chunks (e.g only the unit-test chunk)|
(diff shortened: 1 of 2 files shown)


daweilv:
When will this change release?

neo7official:
I have tried this fix but still issue remains. When devServer runs first time all css chunks converted into separate style block and works OK. But then if we made modifications then page got refreshed but style block did not appended into html on rutime.

I am pasting parts of my configuration here



`
{
				test:/\.(css|sass|scss)$/,
				use:[function(){
						if(mode === "production")
							{
								return {
									loader:MiniCssExtractPlugin.loader,
									options:{
										publicPath:"../"
									}
								}
							}else{
								return "style-loader";
							}
					}(),
					"css-loader", "fast-sass-loader", "postcss-loader"],
			}

new HtmlWebpackPlugin({
				template: path.resolve(src_path, `${ent}.html`),
				filename:path.resolve(dist_path, `${ent}.html`),
				inject:true,
				includeSiblingChunks :true,
			    chunks:[ent]
			})

Optimization.splitChunks={
	name:Public.options.splitChunks.name,
	chunks:"all",
};
`

PR #958, opened 2018-05-29
addFileToAssets allow output name override

WearyMonkey:
Allows for e.g. including a hash in the output name.


jantimon:
Could you add an example how this might be useful?

WearyMonkey:
@jantimon I've added `[contentHash]` and `[name]` placeholder support and a test which should make the usage more clear.

We are using the `add-asset-html-webpack-plugin` to add assets to the compilation and html file. I've raised a PR there to propagate the option: https://github.com/SimenB/add-asset-html-webpack-plugin/pull/124

WearyMonkey:
@jantimon any thoughts on this?

Issue #962, opened 2018-05-31
Options.cache does not prevent unchanged HTML pages from being rebuilt in watch mode


dwoznicki:
### Background
I'm using `html-webpack-plugin` to generate around 25 HTML files for my project. These pages are basically static and rarely need to be rebuilt.

### Problem
When running `webpack` in watch mode, changes to any file causes all HTML files to be rebuilt, even when `options.cache` is set to `true`. This increases the build time from ~500ms to ~3000ms.

### Cause
The cache check [on line 132 of `index.js`](https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L132) is never reached because `isCompilationCached` always comes back false. This is because `compilationResult.hash` [(line 81)](https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L81) changes on every recompile.

### Possible fix
I can fix this behavior by removing the check for `isCompilationCached` on line 132.
```javascript
//if (isCompilationCached && self.options.cache && assetJson === self.assetJson) {
if (self.options.cache && assetJson === self.assetJson) {
```
This solves the issue for my needs, but I'm not sure what the implications of making this change are.

**Edit**: This change causes the "HtmlWebpackPluginCaching should compile the webpack html if the template file was changed" test to fail. This never comes up for my specific use case as the template file does not change, but I'm thinking my use case might be more of a feature request.
Related: #963 

### Environment
```
Node.js v8.11.1
linux 4.12.0-041200rc1-generic
npm version 5.6.0
webpack@4.5.0
html-webpack-plugin@3.2.0
```

wxlworkhard:
`isCompilationCached` is to mark if the template file is changed.
A HtmlWebpackPlugin instance correspond to a webpack chunk.
`compilationResult.hash` is  this chunk's hash.

However after webpack upgrade to version 4，`compilationResult.hash` is always changes even I do not change template file, Why?

`compilationResult.hash`(chunk's hash) is generate by `chunk.id` `chunk.ids` `chunk.name` `chunk._modules`. You can see `webpack/lib/Chunk.js` file at line 407 the `updateHash` method.

In webpack@4, the id is change between 1 and 0 every time  when you change any file. I know that id is always 0 in webpack@2.3.3. The same chunk's id should not change.

The `chunk.id` change lead to `compilationResult.hash` change and `isCompilationCached` is always `false`.

[...]

PR #963, opened 2018-05-31
Option to only build HTML on initial compile

dwoznicki:
Adds option to only build HTML files on initial compilation. For my project, html pages are basically static and only need to be recreated for the first build. Setting this option reduces my recompile time during watch mode for 25 html files from ~3000ms to ~500ms.

This addresses the long recompile time for issue #962

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,8 +33,10 @@ class HtmlWebpackPlugin {
       chunksSortMode: 'auto',
       meta: {},
       title: 'Webpack App',
-      xhtml: false
+      xhtml: false,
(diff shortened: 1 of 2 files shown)


jantimon:
Is this still required if we merge #972 ?

dwoznicki:
Probably not, but let me test build times again. If this option provides a significant decrease in compilation time, I may still push for it. Otherwise I'll close this PR.

dwoznicki:
@jantimon I don't see any difference between this options and #972, so I'm going to close this PR now.

PR #965, opened 2018-06-03
Timestamp based check for changes to HTML template

dwoznicki:
As of webpack version 4, `isCompilationCached` no longer appears to be a reliable method of checking whether a template file has been changed. Instead, we can check `compilation.fileTimestamps` to see when the template file was last modified.
For a 25 html page project, I've found this change reduces the rebuild time from ~3000ms to ~700ms in `--watch` mode.

**NOTE**: this commit fails the build test because `WebpackRecompilationSimulator.simulateFileChange`  does not change the `compilation.fileTimestamps` object.
One way to solve this might be to have `simulateFileChange` write to the source file and then immediately revert the changes. Another might be to initialize the `HtmlWebpackPlugin` with the temp file as the template instead of the source file. Both require making changes to [webpack-recompilation-simulator](https://github.com/jantimon/webpack-recompilation-simulator), which I'm not sure how to build.
In practice, I've found this commit to work as expected with `webpack --watch`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,8 +39,8 @@ class HtmlWebpackPlugin {
 
   apply (compiler) {
     const self = this;
-    let isCompilationCached = false;
     let compilationPromise;
+    this.templateModifiedTimestamp = Date.now();
 
     this.options.template = this.getFullTemplatePath(this.options.template, compiler.context);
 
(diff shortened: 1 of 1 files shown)


jantimon:
Right now I am fiddling around with a new child compiler but I need some more time to get it working.

dwoznicki:
Cool! That should work quite well. I'm going to grab your changes in `feature/single-compiler` and make another pull request actually since it strikes me that I'm attempting to merge into `master` instead of `webpack-4` here.

PR #967, opened 2018-06-03
feat(compiler): Use a single compiler for multiple plugin instances

jantimon:
This is a wip proposal to reuse a single compiler for multiple HtmlWebpackPlugin instances.

#962
#965

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,12 +84,19 @@ class HtmlWebpackPlugin {
       this.options.filename = path.relative(compiler.options.output.path, filename);
     }
 
+    // Clear the cache once a new HtmlWebpackPlugin is added
+    childCompiler.clearCache(compiler);
+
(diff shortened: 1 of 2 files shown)


vhmth:
@jantimon what's the best way I can incorporate this into my repo for testing? Would love to test build times and get back to you with any weird discrepancies.

jantimon:
```
cd node_modules
git clone https://github.com/jantimon/html-webpack-plugin.git
git checkout feature/cache-file-timestamps
```

vhmth:
@jantimon dopeness!

Went through and it seems like this might be the last thing to land on `webpack-4`? The `webpack-4` branch I take it updates `html-webpack-plugin@next` automatically? If so, can pull from that and try this immediately.

jantimon:
It's not published yet because of #972 but we might ship a new alpha soon 👍

vhmth:
Ack ack!

PR #972, opened 2018-06-08
feat(compiler): Use timestamps to verify cache validity

jantimon:
Based on #967 -  try to im prove the performance by caching the entire child compilation.

This chang verifies the cache by comparing the child compilations file dependencies with the main compilation timestamps.

PR #979, opened 2018-06-14
Add fix for failing tests and other bugs for file timestamp based cache

dwoznicki:
### Test fixes
- WebpackRecompilationSimulator does not appear to modify the original file, and therefore does not change the timestamp in `compilation.fileTimestamps` (or at least it doesn't on my machine). I made a temporary monkey-patch fix in [ spec/CachingSpec.js ](spec/CachingSpec.js), but really it belongs in WebpackRecompilationSimulator.

### Bug fixes
- Caching whether the template cache was valid in `HtmlWebpackChildCompiler` was causing html files to rebuild every compilation for me, so I moved it out.
- I found that if I don't clear the child compiler cache every compilation, webpack stops watching my template files. I really have no idea what to make of it, and it's difficult to create tests for since it only seems to happen in `--watch` mode.

All checks run in the following environment.
```
Node.js v8.11.1
linux 4.12.0-041200rc1-generic
npm version 5.6.0
webpack@4.5.0
html-webpack-plugin@3.2.0
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -87,7 +87,10 @@ class HtmlWebpackPlugin {
     // Clear the cache once a new HtmlWebpackPlugin is added
     childCompiler.clearCache(compiler);
 
-    compiler.hooks.compile.tap('HtmlWebpackPlugin', () => {
+    compiler.hooks.compilation.tap('HtmlWebpackPlugin', (compilation) => {
+      if (childCompiler.hasOutDatedTemplateCache(compilation)) {
+        childCompiler.clearCache(compiler);
+      }
       childCompiler.addTemplateToCompiler(compiler, this.options.template);
     });
 
@@ -101,12 +104,13 @@ class HtmlWebpackPlugin {
           compilation.errors.push(prettyError(err, compiler.context).toString());
           return {
             content: self.options.showErrors ? prettyError(err, compiler.context).toJsonHtml() : 'ERROR',
-            outputName: self.options.filename
+            outputName: self.options.filename,
+            hash: ''
           };
         })
         .then(compilationResult => {
           // If the compilation change didnt change the cache is valid
-          isCompilationCached = compilationResult.hash && self.childCompilerHash === compilationResult.hash;
+          isCompilationCached = Boolean(compilationResult.hash) && self.childCompilerHash === compilationResult.hash;
           self.childCompilerHash = compilationResult.hash;
           self.childCompilationOutputName = compilationResult.outputName;
           callback();
@@ -121,9 +125,6 @@ class HtmlWebpackPlugin {
      * @param {() => void} callback
     */
       (compilation, callback) => {
-        // Clear the childCompilerCache
-        childCompiler.clearCache(compiler);
-
         // Get all entry point names for this html file
         const entryNames = Array.from(compilation.entrypoints.keys());
         const filteredEntryNames = self.filterChunks(entryNames, self.options.chunks, self.options.excludeChunks);
--- a/index.js
+++ b/index.js
@@ -89,6 +89,11 @@ class HtmlWebpackPlugin {
 
     compiler.hooks.thisCompilation.tap('HtmlWebpackPlugin', (compilation) => {
       isTemplateCacheOutdated = childCompiler.hasOutdatedTemplateCache(compilation);
+      if (isTemplateCacheOutdated) {
+        childCompiler.clearCachedPromise(compiler);
+      } else {
+        childCompiler.injectFileDependencies(compilation);
+      }
       childCompiler.addTemplateToCompiler(compiler, this.options.template);
     });
 
@@ -108,7 +113,6 @@ class HtmlWebpackPlugin {
         })
         .then(compilationResult => {
           // If the compilation change didnt change the cache is valid
-          isCompilationCached = Boolean(compilationResult.hash) && self.childCompilerHash === compilationResult.hash;
           self.childCompilerHash = compilationResult.hash;
           self.childCompilationOutputName = compilationResult.outputName;
           callback();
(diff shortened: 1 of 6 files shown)


jantimon:
Wow thanks could you fix it directly inside https://github.com/jantimon/webpack-recompilation-simulator instead extending a broken version?

Otherwise we would have to maintain the fixes inside the html-webpack-plugin.

Is there any reason for renaming all variables? Imho `needsRebuild` is less accurate than `hasOutDatedTemplateCache`

I having troubles to understand all the changes - why would you introduce `previousCompilationTimestamp` which is now a singleton although there might be multiple child compiler instances?

`needsRebuild` has an sideEffect of clearing the cache. 
I would prefer if a function does only one thing.

I am not sure but did you also remove the vm eval cache?

dwoznicki:
> Wow thanks could you fix it directly inside https://github.com/jantimon/webpack-recompilation-simulator instead extending a broken version?

I can do that, but I'm not sure how to build it, so I put it here to get the tests to pass.

> Is there any reason for renaming all variables? Imho needsRebuild is less accurate than hasOutDatedTemplateCache

Yeah that's also fine. I had just been working on this before you released `feature/cache-file-timestamps`, so I ended up using the variable names I had originally been using on my other branch.

> I having troubles to understand all the changes - why would you introduce previousCompilationTimestamp which is now a singleton although there might be multiple child compiler instances?

I'm not sure I totally understand all the ways this plugin works yet. What would cause there to be multiple child compilers? In any case, I can change it to use a WeakMap instead of a singleton. I do need to track previous compilation timestamps in some way.

> needsRebuild has an sideEffect of clearing the cache.

I can restructure it to be called somewhere in `index.js` probably.

> I am not sure but did you also remove the vm eval cache?

I don't think so, but I'm not totally sure what that is/used for.

Anyway, it's kinda late here, so I'm gonna go to sleep now. Just wanted to get this PR in tonight :]

jantimon:
@dwoznicki I tried to fix the webpack compilation simulator but it seems that the caching logic was changed or is broken:

webpack/webpack/issues/7533

dwoznicki:
@jantimon Ahh, that sucks.
I made some changes to bring things back more in line with the origin branch.

### Notes
- Renamed `needsRebuild` to `hasOutdatedTemplateCache`
  - previously `hasOutdatedTemplateCache` returned **false** if a file dependency was changed (to determine if the child compiler cache needed to be cleared)
  - now it returns **true** if a file dependency was changed (to determine if the html template needs to emitted). 
- Child compiler cache is cleared at the end of every compilation
  - added `childCompiler.clearCachedPromise` to unset compilation promise without unsetting compilation started timestamp
- Moved `hasOutdatedTemplateCache` check into `thisCompilation` hook because `compilation` hook was being called twice per compilation for some reason

jantimon:
Cool I'll look into your changes very soon :)

I have also opened a new pr here https://github.com/jantimon/webpack-recompilation-simulator/pull/2

jantimon:
@sokra told me that caching works only for development - so I adjusted the tests for the webpack-recompilation-simulator and published 2.0.0:

https://www.npmjs.com/package/webpack-recompilation-simulator

jantimon:
@denis-sokolov "Child compiler cache is cleared at the end of every compilation" would require more compilations. Why do we need that?

denis-sokolov:
Wrong mention, @jantimon? I have no idea what you’re talking about.

dwoznicki:
Failing to clear the compiler cache at the end of every compilation causes a strange bug where the file watcher for for HTML template files is removed on subsequent compilations for me. I.e., the compiler stops watching template files.
I'm having trouble creating tests to reproduce this issue since it only occurs in `--watch` mode. Let me try pulling this version of the branch and see if the issue persists.

jantimon:
So webpack keeps track during watching of all files here:

https://github.com/webpack/webpack/blob/4de3ce057d5ce64055fc258505e004d2eccb2cb7/lib/Compiler.js#L146-L147

So we might have to extend the webpack-recompilation-simulator to create reliable unit tests for that case.

dwoznicki:
I'll go ahead and see if I can put together a test in that case.

dwoznicki:
Hey, so I found out what the issue is, but I'm having trouble solving it without clearing the cache at the end of every recompilation. Basically here's what's happening:

1. At the end of every compilation, webpack takes the file dependencies from the compilation object and uses them to determine the watched files. https://github.com/webpack/webpack/blob/8a7597aa6eb2eef66a8f9db3a0c49bcb96022a94/lib/Watching.js#L102-L106
2. The file dependencies of child compilations are also added to the main compilation's file dependencies here. https://github.com/webpack/webpack/blob/e3678aa3767aaeb26ebcf5e3028412bbebdb4f9f/lib/Compilation.js#L1735-L1745
3. Child compilations only get added when `childCompiler.runAsChild` is called. https://github.com/webpack/webpack/blob/e3678aa3767aaeb26ebcf5e3028412bbebdb4f9f/lib/Compiler.js#L227-L245
4. With our current setup, `runAsChild` is not called for cached promises, so the child compilation dependencies (such as the template file) never get added to the file watcher.

I tried caching and the compilation object per child compiler, but it's causing basically all the build stats to be wrong for testing.
I'm still working on this, but it's taking a while atm.

jantimon:
@dwoznicki what about copying the file dependencies (which we store anyway) to the main compilation after each build?

dwoznicki:
@jantimon Hey, couple more changes
### Notes
- No longer clears child compilation each round
- Injects child compilation file dependencies into main compilation file dependencies if template has not changed
  - https://github.com/jantimon/html-webpack-plugin/pull/979/files#diff-168726dbe96b3ce427e7fedce31bb0bcR92
- removed the temp version of webpack-recompilation-simulator, so this branch is now dependent on webpack/loader-utils#115 as well

I've found that these changes work as expected for `--watch` mode in a small test project.

jantimon:
@dwoznicki The webpack-recompilation-simulator does now support watch mode tests: https://github.com/jantimon/webpack-recompilation-simulator#watch-mode

I will release a new version of the webpack-recompilation-simulator once https://github.com/webpack/loader-utils/issues/115 is solved

dwoznicki:
These recent changes to the `feature/cache-file-timesamps` branch cause a ton of conflicts with my version and appears to add a lot of the same functionality. Is there any reason to keep this PR open?

jantimon:
I am not 100% happy with feature/cache-file-timestamps so I might take over some more of your ideas 👍

jantimon:
Thanks again @dwoznicki for all your inputs and work - I merged a lot of it into #972 and will close this for now.

PR #986, opened 2018-06-22
feat: Allow URL or Data URI favicon

wuwb:
feat: Allow URL or Data URI favicon.


mastilver:
Did you have a look at: https://github.com/jantimon/favicons-webpack-plugin ?

wuwb:
@mastilver yes, i had looked through favicons-webpack-plugin. but in my project, i use the configs extracted from vue-route and a custom config with favicon to generate html-webpack-plugin configs, if use favicons-webpack-plugin, i have to generate a mapping favicons-webpack-plugin config to each html-webpack-plugin config, and favicons-webpack-plugin didn't providate this feature, it inject the same config to every html-webpack-plugin config.

edmorley:
Might another way to do this, be to have favicon use whatever webpack loader has been declared for the file type is it? ie: Then if `.(ico|...)` have been configured to use url-loader, the favicon will either be inlined as a data URI or as a path to the copied asset, automatically?

edmorley:
(That approach would also solve the "have hash in favicon filename" use-case)

PR #988, opened 2018-06-26
Inner links added to options

atilkan:
I hope i did it correctly. Can't test.


atilkan:
Nope, it is not correct. I will test before doing again. Just reject.

PR #992, opened 2018-07-03
Add placeholder for custom inject position.

mudoo:


jantimon:
If you need more control of the template please disable the injection mode

mudoo:
@jantimon Sometimes, we just need to keep something below the head tag, if disable the injection mode, and there are many templates in the project, it's too tedious.

jantimon:
I see - but still it doesn't feel right to invent a second template language on top of another.

Right now we are focussing on performance improvements for the next major release.  
However one feature we might also add is to pass the rendered html of the head and body parts.

That way you could place them anywhere in your template.
Would that help?

mudoo:
That will be great. tks.

PR #1014, opened 2018-08-01
Add support for `stats.entrypoints[].childAssets`

### LINKED ISSUE #934: Add support for `stats.entrypoints[].childAssets`
Webpack 4.6 allows to prefetch/preload dependencies using dynamic imports:

https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c

The webpack core api allows to consume those information also for html generation.
So we could turn the following data into `<link />` tags:

```
console.log(stats.entrypoints[].childAssets);

{ 
 main: 
 { 
   assets: ["a.js", "a.css"], 
   childAssets: { 
     preload: [ assets: ["b.js", "b.css"] }, 
     prefetch: { assets: ["c.js", "c.css"] 
    } 
 } 
}
```

--->

```
head
  link(rel=preload, href=a.js, as=script)
  link(rel=preload, href=a.css, as=stylesheet)
  link(rel=preload, href=b.js, as=script)
  link(rel=preload, href=b.css, as=stylesheet)
  link(rel=prefetch, href=c.js, as=script)
  link(rel=prefetch, href=c.css, as=stylesheet)
  ...
  link(rel=stylesheet, href=a.css)
body
  ...
  script(src=a.js)
```

@mastilver do you think we should add this to the html-webpack-plugin itself or rather as a plugin?

loveky:
fix #934


loveky:
travis ci is failing becaus the webpack version used in CI is forced to 4.0.0

jantimon:
I couldn't find a way to get around the `stats` object for this.  
And we just removed the `stats` object because the generation was very slow.  

Any ideas how we might get around that?

jimmydief:
@jantimon Is the call to `getStats().toJson()` still a concern if we were to make use of the optional `options` argument to pick and choose only the parts of the stats that we need? I've seen big perf improvements from doing so in our fairly large project (~6000 modules) as it can avoid serializing information about each individual asset.

`toJson` is documented [here](https://webpack.js.org/api/node/#statstojsonoptions) and its options are documented [here](https://webpack.js.org/configuration/stats/). I think that we could get by with `getStats().toJson({all: false, entrypoints: true})` in this case.

If that's acceptable, I might ask to tack `chunks: true` onto that list as well as I have some use cases for having that information available in the template.

Issue #556, opened 2017-01-22
`[hash]` in filename uses template file hash, not hash of built html


jenan-stripe:
When using `new HtmlWebpackPlugin({ filename: 'index-[hash].html', ...})`, the hash value seems to be based on the text content of the template file, not on what it produces. This means that changing dependent chunks will not update the output filename, even though the contents have changed. This appears to be the case with both `inject: true` and explicitly using `ejs` templating to include `<script src="..."></script>` tags. 

E.g., if I have a template `index.html`:

```html
<html>
<body>
</body>
</html>
```

and use a webpack config with:

```js
entry: {index: 'index.js'},
output: {path: 'build/', filename: '[name]-[hash].js'},
plugins: [
  new HtmlWebpackPlugin({
    filename: 'index-[hash].html',
    template: 'index.html',
    inject: true,
    chunks: ['index'],
  }),
]

````

and then get a `build/index-aaaaaaaa.html` of:

```html
<html>
<body>
<script type="text/javascript" src="index-bbbbbbb.js"></script>
</body>
</html>
```

and then change the javascript for `index.js`, I'll get a `build/index-aaaaaaaa.html` (same file name as before) of:

```html
<html>
<body>
<script type="text/javascript" src="index-cccccccc.js"></script>
</body>
</html>
```

jenan-stripe:
[Here](https://github.com/jenan-stripe/minimal-html-webpack-plugin-issue-556) is a minimal reproduction.

JuoCode:
+1

jantimon:
@jenan-stripe the reason is that it uses the hashing algorithm by webpack - otherwise we would have issues with caching while using the webpack middleware or the webpack dev server

jenan-stripe:
@jantimon Hm does anything else in webpack exhibit this behavior, where the contents may change but the fingerprint does not?

lbalceda:
@jenan-stripe I'm running into the same problem with a similar configuration. were you able to get it to work?

jenan-stripe:
@lbalceda Nope. I'm manually fingerprinting in a post-build step.

moimael:
It makes it very hard to cache the html in a service worker for offline use. This is definitely something that should be fixed.

Measy:
@jenan-stripe Did u still manually fingerprinting in a post-build step or find a new way to get it work?

jenan-stripe:
@Measy Still manually fingerprinting.

[...]

PR #1021, opened 2018-08-07
feat: Add support for the [contenthash] placeholder inside htm file names

jantimon:
Solves #556


jantimon:
@jenan-stripe could you please take a look if you this could solve your problem?

PR #1024, opened 2018-08-14
docs: add 'templateContent' option

lightpohl:
Hello!

Noticed this option was missing from docs. Let me know if there are any issues with adding it, or if you'd like the wording changed a bit. 

This plugin is extremely handy by the way! Thanks for maintaining it. 😃


jantimon:
Cool :)

could you please keep the spaces you removed?

lightpohl:
Of course! Updated. 👍

PR #1027, opened 2018-08-16
add media-query-plugin to the docs

SassNinja:
This PR adds a link to the [media-query-plugin](https://github.com/SassNinja/media-query-plugin) to the docs.
It plays together well with the html-webpack-plugin and provides the extracted files to the template with the query data to let you use something as `<link media='query'>`

I'm aware there's already a similar plugin in the list (link-media-html-webpack-plugin). But its description is wrong (the browser downloads the files always, not conditionally) and it provides less features.

Thus I hope it's ok to add the media-query-plugin, too.


jantimon:
Hey @SassNinja

I am very sorry for postponing this issue for soooo long.  
The main focus was the major 4 release of the html-webpack-plugin.

I really like your idea of splitting css into different files by media query.
Could you please try to add support for the new html-webpack-plugin hooks api and rebase this pull request?

PR #1030, opened 2018-08-22
Improve migration docs in regards to changes in template syntax

samloibl:
Minor improvement to the migration docs that highlights the fact that unescaped output has an unexpected syntax change.


jantimon:
writing good documentation is hard - thank you so much for helping out 👍

PR #1034, opened 2018-08-27
Rename contenthash to hashByContent to resolve internal webpack conflict

### LINKED ISSUE #1033: [v4] Fingerprinting the output html file name with a hash determined by the html contents conflicts with webpack's `contenthash` variable
## Expected behaviour

It should fingerprint the output html file name with a hash determined by the html contents after the scripts have been injected

## Current behaviour

Webpack throws error:
```
UnhandledPromiseRejectionWarning: Error: Path variable [contenthash] not implemented in this context: html/app.[contenthash].html
```
## Environment

```
node: v8.11.2
npm: v5.6
yarn: v1.9.2
webpack@4.17.1
html-webpack-plugin@4.0.0-alpha
```

### Config

```js
module.exports = {
  ...,
  plugins: [
    new HtmlWebpackPlugin(
      template: './app/assets/html/app.html',
      chunks: ['app-bundle', 'vendor~app-bundle'],
      filename: IS_DEV ? 'index.html' : 'html/app.[contenthash].html'
    ...
  ]
}
```

### Similar issues
There was a similar behaviour from extract-text-webpack-plugin in the past when `webpack-4` introduced this internally in version `4.3.0`:

https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/763

pegiadise:
This occurs with webpack >= 4.3 because they internally added a [contentHash] variable. The simplest way around it is to rename the key to something else which is what I did here. I needed a quick fix for it, but I also stand by the choice of choosing a different key there because:

1. more declarative as to what it is doing
2. treating webpack internal variables as a black box

Closes #1033

### DIFF
--- a/index.js
+++ b/index.js
@@ -239,9 +239,9 @@ class HtmlWebpackPlugin {
             return self.options.showErrors ? prettyError(err, compiler.context).toHtml() : 'ERROR';
           })
           .then(html => {
-            // Allow to use [contenthash] as placeholder for the html-webpack-plugin name
+            // Allow to use [hashByContent] as placeholder for the html-webpack-plugin name
             // From https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/8de6558e33487e7606e7cd7cb2adc2cccafef272/src/index.js#L212-L214
-            const finalOutputName = self.childCompilationOutputName.replace(/\[(?:(\w+):)?contenthash(?::([a-z]+\d*))?(?::(\d+))?\]/ig, function () {
+            const finalOutputName = self.childCompilationOutputName.replace(/\[(?:(\w+):)?hashByContent(?::([a-z]+\d*))?(?::(\d+))?\]/ig, function () {
               return loaderUtils.getHashDigest(html, arguments[1], arguments[2], parseInt(arguments[3], 10));
             });
             // Replace the compilation result with the evaluated html code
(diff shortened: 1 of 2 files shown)


jantimon:
Wouldn't it be better to keep [contenthash] and tell webpack to ignore it instead of coming up with our own new syntax?

pegiadise:
Maybe you are right. I took a look at `mini-css-extract-plugin` way of [doing it](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/a78b301aa373d78a457c7224a3523604a4a4ff12/src/index.js#L229), but you might be more familiar with tapping into compilation hooks. Also it would probably require bumping webpack to a higher version and I don't know if there are any dependencies right now holding you there.

jantimon:
Would it work to replace `[contenthash]` to `[html-webpack-plugin-contenthash]` during startup and change https://github.com/jantimon/html-webpack-plugin/blob/fb448dd6b6e09613af1ec974925e1bcf714bcbce/index.js#L268 to `html-webpack-plugin-contenthash`?

sibiraj-s:
how about `templatehash`?

pegiadise:
@sibiraj-s `[templatehash]` sounds like a good alternative but I think the maintainer wants to keep the `[contenthash]` in order to generally comply with webpack API. @jantimon I 'll give your proposal a try tonight and get back to you.

sibiraj-s:
> but I think the maintainer wants to keep the [contenthash] in order to generally comply with webpack API.

hi @pegiadise . You are correct. Can you check #1065 before you work on something? I tried to preserve both `contenthash` and `templatehash`

jantimon:
Moved to #1065

PR #953, opened 2018-05-21
html-webpack-plugin 4.x - Switching to Webpack 4 api

jantimon:
### Bug Fixes

* **chunksorter:** Don't sort chunks by default ([22fb03f](https://github.com/jantimon/html-webpack-plugin/commit/22fb03f))
* **loader:** switch to loaderUtils.getOptions ([a0a0f0d](https://github.com/jantimon/html-webpack-plugin/commit/a0a0f0d))
* **tests:** Upgrade webpack-recompilation-simulator ([dfe1d10](https://github.com/jantimon/html-webpack-plugin/commit/dfe1d10))
* Add dependencies from the child compilation to the main compilation ([27c3e72](https://github.com/jantimon/html-webpack-plugin/commit/27c3e72))
* Ignore foreign child compilers ([1422664](https://github.com/jantimon/html-webpack-plugin/commit/1422664))
* Improve perfomance for appcache files ([b94e043](https://github.com/jantimon/html-webpack-plugin/commit/b94e043))
* Prevent chunks from beeing added multiple times ([d65b37d](https://github.com/jantimon/html-webpack-plugin/commit/d65b37d))
* Remove compilation.getStats() call for performance reasons ([7005a55](https://github.com/jantimon/html-webpack-plugin/commit/7005a55))
* Update lodash to 4.17.10 ([cc3bf49](https://github.com/jantimon/html-webpack-plugin/commit/cc3bf49))

### Code Refactoring

* Change the structure of the internal assets object ([37db086](https://github.com/jantimon/html-webpack-plugin/commit/37db086))
* Changed hook names and arguments - the hook order is 'beforeAssetTagGeneration', 'alterAssetTags', 'alterAssetTagGroups', 'afterTemplateExecution', 'beforeEmit', 'afterEmit' ([14b4456](https://github.com/jantimon/html-webpack-plugin/commit/14b4456))
* Drop support for Webpack 1-3 to use Webpack 4 apis ([47efdea](https://github.com/jantimon/html-webpack-plugin/commit/47efdea))

### Features

* Add default viewport meta tag for default template ([302e39e](https://github.com/jantimon/html-webpack-plugin/commit/302e39e)), closes [#897](https://github.com/jantimon/html-webpack-plugin/issues/897) [#978](https://github.com/jantimon/html-webpack-plugin/issues/978)
* **hooks:** Add a helper for easier hook access ([b6dec4b](https://github.com/jantimon/html-webpack-plugin/commit/b6dec4b))
[...]

PR #1038, opened 2018-08-28
Example/chunk optimization

jantimon:


edmorley:
Hi! If I'm reading the diff here correctly, it looks like it might resolve some of the incompatibilities with `html-webpack-template` (https://github.com/jantimon/html-webpack-plugin/pull/953#issuecomment-416320338) - is that correct? :-)

jantimon:
Yes that's correct

PR #1048, opened 2018-09-16
Enable minification by default when 'mode' is production

### LINKED ISSUE #1036: Decide what to do with the HTML minification feature
**Is your feature request related to a problem? Please describe.**
`html-webpack-plugin` has a `minify` option that provides support for minifying the generated HTML. 

However:
- it currently defaults to off (`false`) so unless people manually enable it, isn't used
- it adds 14 transitive dependencies (according to [cost-of-modules](https://www.npmjs.com/package/cost-of-modules))
- there are other plugins out there for HTML minification (for example [html-minifier-webpack-plugin
](https://www.npmjs.com/package/html-minifier-webpack-plugin)) in case people do want to minify
- the win from minifying HTML is minimal, since most of the possible optimisations can (and IMO should) really be made to the original template - making it easier to track down any possible regressions that result
- the current implementation in `html-webpack-plugin` has some issues (eg #743)

**Describe the solution you'd like**
For the minify feature and the dependency on the `html-minifier` package to be removed in the next major version of `html-webpack-plugin` (ie v4 on the `webpack-4` branch).

This idea was previously mentioned in #558 and #452.

**Describe alternatives you've considered**
Leaving the `html-minifier` dependency as-is in v4.

**Additional context**
N/A

edmorley:
Previously minification was disabled by default. Now, if `minify` is `undefined` and `mode` is `'production'`, then it is enabled using the following options:

```js
{
  collapseWhitespace: true,
  removeComments: true,
  removeRedundantAttributes: true,
  removeScriptTypeAttributes: true,
  removeStyleLinkTypeAttributes: true,
  useShortDoctype: true
}
```

These options were based on the settings used by create-react-app, Neutrino and vue-cli, and are hopefully fairly conservative. See:
https://github.com/jantimon/html-webpack-plugin/issues/1036#issuecomment-421408841
https://github.com/kangax/html-minifier#options-quick-reference

These same defaults can enabled regardless of `mode`, by setting `minify` to `true` (which previously passed an empty object to html-minifier, meaning most minification features were disabled). Similarly, minification can be disabled even in production, by setting `minify` to `false`.

This change has no effect on users who pass an object to `minify`.

Fixes #1036.


index.js
```diff
@@ -99,6 +99,19 @@ class HtmlWebpackPlugin {
       this.options.filename = path.relative(compiler.options.output.path, filename);
     }
 
+    const minify = this.options.minify;
+    if (minify === true || (minify === undefined && compiler.options.mode === 'production')) {
```
jantimon: Could you please add the following code to make it easier to understand?

```js
// Check if webpack is running in production mode 
// @see https://github.com/webpack/webpack/blob/3366421f1784c449f415cda5930a8e445086f688/lib/WebpackOptionsDefaulter.js#L12-L14
const isProductionLikeMode = compiler.options.mode === 'production' || !compiler.options.mode;
```

typings.d.ts
```diff
@@ -53,7 +53,7 @@ interface HtmlWebpackPluginOptions {
      * HTML Minification options
      * @https://github.com/kangax/html-minifier#options-quick-reference
      */
-    minify: boolean | {},
+    minify: undefined | boolean | {},
```
jantimon: All properties are optional so we don't need to add the undefined here :)

edmorley: Without this I got a type error. I can try changing it to optional instead?

index.js
edmorley: Good idea :-)

edmorley:
One of the Travis jobs failed due to an infra error - could you restart it? (Travis only permits those with commit access to restart jobs unfortunately)

typings.d.ts
jantimon: Oh sorry you are absolutely right - usually all options are non optional after the constructor.  
But as we don't know the compiler state at this time your solution is absolutely correct.

edmorley: Would you prefer optional or a union with undefined?

jantimon:
Thank you so much 👍

edmorley:
Thank you for the review/merge :-)

PR #1057, opened 2018-09-27
Add link to new html-webpack-link-type-plugin to README.md

swimmadude66:
add link to new html-webpack-link-type-plugin


jantimon:
Looks really cool :)

We are about to release the beta soon - unfortunately it won't work with your plugin anymore because we switched to a new hook system (proposed by the webpack core team)

Could you please try to support the current beta version as well?

flow

swimmadude66:
Seems pretty straightforward to migrate. Will the `htmlWebpackPluginAlterAssetTags` hook we are currently using be removed at the same time? Just trying to come up with a way to detect the new hooks, and use the correct ones

jantimon:
Yes they were removed for html-webpack-plugin 4

swimmadude66:
@jantimon I am looking in to this now, and on branch `4.0.0-beta.1` I cannot seem to find the appropriate hook, I see the `hooks.js` file with the same docs as you posted above, but when I print all the keys inside `compilation.hooks` I don't see any which directly match the hooks I expect. Am I missing some other change here?

PR #1060, opened 2018-09-28
Add csp-html-webpack-plugin

sibiraj-s:
csp-html-webpack-plugin adds CSP meta tags to the HTML output.


jantimon:
Thanks alot :)  
Please note that we will relase the new major 4 version soon - so maybe you could also provide a pr for the webpack-4 branch.

Did you already see the meta tag option which does also allow to use CSP?

https://github.com/jantimon/html-webpack-plugin/tree/webpack-4#meta-tags

sibiraj-s:
> Did you already see the meta tag option which does also allow to use CSP?

Cool. didn't see that.

I made a PR to master since it doesn't support `html-webpack-plugin-v4`.

There is still a PR on `csp-plugin` to support v4 which will be merged once html-webpack-plugin the final version is released . but that plugin does a little more than adding tags, like generating `SHA` based on the content.

sibiraj-s:
Do you want me to change this PR to webpack-4 branch?

jantimon:
Wow that's pretty pretty cool - so you can basically guarantee that no forged script can be load?

sibiraj-s:
Yes that is correct.

PR #1062, opened 2018-09-29
Migrate to mini-css-extract-plugin from extract-text-webpack-plugin

sibiraj-s:
**Migrated from `ETWP` to `mini-css-extract-plugin`**. Webpack discourages the use of ETWP since webpack 4.0

Since `mini-css-extract-plugin` requires webpack>4.4, hence webpack is updated.
Which inturn causes conflicts with `contenthash` so to fix tests. I renamed it to **`templatehash`**.. I can't think of more suitable name than this.

**Breaking**: `contenthash` is renamed to `templatehash`


jantimon:
Sorry this pr includes to many features and also some quite opinionated ones

sibiraj-s:
@jantimon there are no new features added this is more like a cleanup. I barely renamed few stuffs.

sibiraj-s:
What about migrating to `mini-css-extract-plugin`?

edmorley:
In general it's best to keep to one type of change per PR. Looking at the changes here, I see linting fixes, gitignore changes, test cleanups, dependency bumps, deleting dist folders and the switch to mini-css-extract-plugin. Each of those should likely be a separate PR - which then makes it simpler to review/discuss :-)

sibiraj-s:
Yes @edmorley My bad. I thought these changes will be good. but separate PR.

@jantimon can you reopen the PR. I will update this to the change with shifting from 

ETWP ->mini-css-extract-plugin

sibiraj-s:
@jantimon @edmorley Updated the changes. This includes `dist/` folder which is auto-generated

index.js
```diff
@@ -279,10 +279,10 @@ class HtmlWebpackPlugin {
               return self.options.showErrors ? prettyError(err, compiler.context).toHtml() : 'ERROR';
             })
             .then(html => {
-              // Allow to use [contenthash] as placeholder for the html-webpack-plugin name
+              // Allow to use [templatehash] as placeholder for the html-webpack-plugin name
```
edmorley: This change seems unrelated?

examples/custom-template/webpack.config.js
```diff
@@ -13,14 +13,14 @@ module.exports = {
...
+      { test: /\.css$/, use: [MiniCssExtractPlugin.loader, 'css-loader'] },
       { test: /\.png$/, loader: 'file-loader' }
     ]
   },
   plugins: [
     new HtmlWebpackPlugin({
       template: 'template.html'
     }),
-    new ExtractTextPlugin('styles.css')
+    new MiniCssExtractPlugin({ filename: 'style.css' })
```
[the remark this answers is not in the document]
sibiraj-s: yes, too many diff. I missed that. Updated now

Checkout #1064

index.js
sibiraj-s: That is unrelated yes. But tests will fail without that.

edmorley: Ah makes sense. I guess this change fixes #1033.

sibiraj-s: Yes. Just noticed there is also another PR #1034 does the same.

sibiraj-s:
What about updating webpack in peerDependency to `>= 4.4` as mini-css plugin. these plugins will be used together mostly

index.js
jantimon: Actually it's fine for me - maybe we could also replace `contenthash` to `templatehash` before we compile it.  
So users will be able to use `contenthash` and `templatehash`

jantimon:
Thanks alot 👍

index.js
sibiraj-s: Cool. So that no breaking change will be introduced. 👍

sibiraj-s: @jantimon checkout #1065

PR #1063, opened 2018-09-30
fix: the publicPath is set to relative path

Wind4:
## The Problem

Some `webpack.config.js` configuration :

```js
export default {
  entry: {
    'index': './pages/index.js',
    'about/index': './pages/about/index.js',
  },

  output: {
    path: './build',
    filename: '[name].js',
    chunkFilename: '[name].js',
    publicPath: process.env.PUBLIC_URL || '',
  },

  plugins: [
    new HtmlWebpackPlugin({
      filename: 'index.html',
      chunks: ['index'],
    }),

    new HtmlWebpackPlugin({
      filename: 'about/index.html',
      chunks: ['about/index'],
    }),
  ],
}
```

After the compilation is completed: 

image

The generated `build/index.html` looks good, but `build/about/index.html` like this:
```html
<script type="text/javascript" src="about/index.js"></script>
```
The correct path should be `../about/index.js` or `index.js`.

**So, I tried to solve the problem when publicPath was set to relative path.**

## Test Case

image


jantimon:
Hey @Wind4 thank you very much :)

Your change is a breaking change so it would be better to move it to the webpack-4 branch which is the current beta version.

Could you please add jsdoc typing comments and change the branch?

Wind4:
OK, I will rebase to `webpack-4` branch later.

PR #1064, opened 2018-10-01
[Cleanup] Don't Commit `dist/` folders, Generate it dynamically

sibiraj-s:
Having the `dist/` folder creates too many changes when related files are updated. which becomes very difficult to track and see changes

`dist/` folders are now generated dynamically before running tests. Added to gitignore and pretest step.


jantimon:
The `dist/` folders are a reference for users to understand what the plugin is capable of.  
Also the dist files are used for snapshot tests.

sibiraj-s:
Yes. I agree. Hence I added that here https://github.com/jantimon/html-webpack-plugin/pull/1064/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R14 to generate before test

Sometimes any code change or dependency updates leads tests to fail. When I first started working I totally had no idea why the test failed. And the errors message wasn't useful too. Then after meddling for sometime. I regenerated examples and all tests passed. 

It will confusing for contributors, especially for first time contributors (like me 😅 ). If you still want the `dist/` folder, may be I should add the pretest step only?

jantimon:
I agree snapshot tests are a little bit confusing especially if the messages are bad.  
But adding it to pretest it would make those tests useless (as they will always pass).

Most non-breaking changes won't cause any changes so we can be sure that the old generation is still possible.

sibiraj-s:
Thanks for the explanation. Let's keep it then 👍 . looking forward for the final release  😍 . 

Also a quick question. I made a dependency bump in previous PR, which was later reverted since i made a mess in it. Shall i raise it in a separate one now?

sibiraj-s:
I am not so familiar with this. but isn't the intention to run tests with current changes. 

however one will have to re run the build-example before they commit, for the tests to pass. one way or another few files will get updated and sometimes most the files will be updated. in such cases it increases the diff size only and difficult to track.

From my point i don't see much benefits of having them.

PR #1065, opened 2018-10-02
Allow use of contenthash along with templatehash

sibiraj-s:
This allows to use `contenthash` which was removed in #1062 . Now when filename is provided with `contenthash` it will be internally replaced to `templatehash`

Fixes: #1033

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,16 @@ class HtmlWebpackPlugin {
       this.options.filename = path.relative(compiler.options.output.path, filename);
     }
 
+    // `contenthash` is introduced in webpack v4.3
+    // which conflicts with the plugin's existing `contenthash` method, hence
+    // hence is renmaed to `templatehash` to avoid such conflicts
+    const REGEXP_CONTENTHASH = /\[(?:(\w+):)?contenthash(?::([a-z]+\d*))?(?::(\d+))?\]/ig;
+    if (REGEXP_CONTENTHASH.test(this.options.filename)) {
+      this.options.filename = this.options.filename.replace(REGEXP_CONTENTHASH, (match) => {
+        return match.replace('contenthash', 'templatehash');
+      });
+    }
+
     // Check if webpack is running in production mode
     // @see https://github.com/webpack/webpack/blob/3366421f1784c449f415cda5930a8e445086f688/lib/WebpackOptionsDefaulter.js#L12-L14
     const isProductionLikeMode = compiler.options.mode === 'production' || !compiler.options.mode;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -99,6 +99,16 @@ class HtmlWebpackPlugin {
       this.options.filename = path.relative(compiler.options.output.path, filename);
     }
 
+    // `contenthash` is introduced in webpack v4.3
+    // which conflicts with the plugin's existing `contenthash` method, hence
+    // hence is renmaed to `templatehash` to avoid such conflicts
+    const REGEXP_CONTENTHASH = /\[(?:(\w+):)?contenthash(?::([a-z]+\d*))?(?::(\d+))?\]/ig;
```
pegiadise: I think this could be simplified to:

```
      this.options.filename = this.options.filename.replace(REGEXP_CONTENTHASH, (match) => {
        return match.replace('contenthash', 'templatehash');
      });
```

No need for the verbose variable declaration and the if check.

index.js
```diff
@@ -99,6 +99,16 @@ class HtmlWebpackPlugin {
       this.options.filename = path.relative(compiler.options.output.path, filename);
     }
 
+    // `contenthash` is introduced in webpack v4.3
+    // which conflicts with the plugin's existing `contenthash` method, hence
+    // hence is renmaed to `templatehash` to avoid such conflicts
```
pegiadise: Typo and extra hence here 😸

index.js
```diff
@@ -279,10 +289,11 @@ class HtmlWebpackPlugin {
               return self.options.showErrors ? prettyError(err, compiler.context).toHtml() : 'ERROR';
             })
             .then(html => {
+              const REGEXP_TEMPLATEHASH = /\[(?:(\w+):)?templatehash(?::([a-z]+\d*))?(?::(\d+))?\]/ig;
               // Allow to use [templatehash] as placeholder for the html-webpack-plugin name
               // See also https://survivejs.com/webpack/optimizing/adding-hashes-to-filenames/
               // From https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/8de6558e33487e7606e7cd7cb2adc2cccafef272/src/index.js#L212-L214
-              const finalOutputName = childCompilationOutputName.replace(/\[(?:(\w+):)?templatehash(?::([a-z]+\d*))?(?::(\d+))?\]/ig, (_, hashType, digestType, maxLength) => {
+              const finalOutputName = childCompilationOutputName.replace(REGEXP_TEMPLATEHASH, (_, hashType, digestType, maxLength) => {
```
pegiadise: Again, maybe it's just me but I would prefer to see the regex directly here and not having to look for the verbose variable.

index.js
sibiraj-s: Yeah, I thought of moving them both to the very top. hence assigned to a variable. Which I thought may be reusable in future.

index.js
sibiraj-s: Will fix, thanks 👍

index.js
jantimon: I agree - it would be easier to read if we just copy the regexp into the replace

sibiraj-s: Okay. Updated 👍

sibiraj-s:
Hi @jantimon After beta.1 few important changes were added, can we have another `beta` to test these out .

jantimon:
Sure I'll try to release it this weekend :)

PR #1066, opened 2018-10-03
feat: Add publicPath support relative path

Wind4:
## The Problem

Some `webpack.config.js` configuration :

```js
export default {
  entry: {
    'index': './pages/index.js',
    'about/index': './pages/about/index.js',
  },

  output: {
    path: './build',
    filename: '[name].js',
    publicPath: process.env.PUBLIC_URL || '',
  },

  plugins: [
    new HtmlWebpackPlugin({
      filename: 'index.html',
      chunks: ['index'],
    }),

    new HtmlWebpackPlugin({
      filename: 'about/index.html',
      chunks: ['about/index'],
    }),
  ],
}
```

After the compilation is completed: 

image

The generated `build/index.html` looks good, but `build/about/index.html` like this:
```html
<script type="text/javascript" src="about/index.js"></script>
```
The correct path should be `../about/index.js` or `index.js`.

**So, I tried to solve the problem when publicPath was set to relative path.**

### DIFF
--- a/spec/basic.spec.js
+++ b/spec/basic.spec.js
@@ -485,6 +485,51 @@ describe('HtmlWebpackPlugin', () => {
     }, ['<link href="styles.css?%hash%"'], null, done);
   });
 
+  it('should allow to add cache hashes to with the css assets', done => {
+    testHtmlPlugin({
+      mode: 'production',
+      entry: path.join(__dirname, 'fixtures/theme.js'),
+      output: {
+        path: OUTPUT_DIR,
+        filename: 'index_bundle.js',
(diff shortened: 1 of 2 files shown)


spec/basic.spec.js
```diff
@@ -550,6 +595,19 @@ describe('HtmlWebpackPlugin', () => {
     }, ['<link href="styles.css" rel="stylesheet"/>'], null, done);
   });
 
+  it('prepends the webpack public path to relative path', done => {
```
habina: test for when the webpack public path is a relative path

PR #1067, opened 2018-10-03
Add a Git .mailmap entry for Thomas Sileghem

mgol:


mgol:
Re-submitted #903.

jantimon:
Why is this necessary?

mgol:
This was a followup to #794 shortly after it got merged.

PR #1076, opened 2018-10-13
Move @types/tapable to devDeps

### LINKED ISSUE #1075: [4.0.0-beta.1] @types/tapable installed as a dependency
## Expected behaviour

`html-webpack-plugin` should not install any `@types` packages in user's project.

## Current behaviour

`html-webpack-plugin` has `@types/tapable` as a dependency in its `package.json`, so package is installed in user's project.

## Environment

```
Node.js v8.12.0
linux 4.18.12-200.fc28.x86_64
6.4.1
stylable@ /home/avi/projects/stylable
└── webpack@4.20.2

stylable@ /home/avi/projects/stylable
└── html-webpack-plugin@4.0.0-beta.1```
```

### Config

Irrelevant.

AviVahl:
- types packages should not be installed for plugin users
- adjusted bundles to current webpack@4 behavior

closes #1075

### DIFF
--- a/examples/default/dist/webpack-4/bundle.js
+++ b/examples/default/dist/webpack-4/bundle.js
@@ -473,7 +473,9 @@ function addStyle (obj, options) {
 
 	// If a transform function was defined, run it on the css
 	if (options.transform && obj.css) {
-	    result = options.transform(obj.css);
+	    result = typeof options.transform === 'function'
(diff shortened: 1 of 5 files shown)


AviVahl:
@jantimon please also note the `.d.ts` file is not included in published package. It should be added to the "files" field of package.json.

PR #1078, opened 2018-10-15
Add long-term caching details to README

sibiraj-s:
Update long-term caching details to the README.

Additionally, I corrected few typos.


jantimon:
Thanks! 👍

README.md
```diff
@@ -383,6 +383,30 @@ plugins: [
...
+
+`contenthash/templatehash` is the hash of the content of the output file.
+
+Optionally, You can configure like `[<hashType>:contenthash:<digestType>:<length>]`
+
+* `hashType` - one of `sha1`, `md5`, `sha256`, `sha512`  or any other node.js supported hash type
+* `digestType` - one of `hex`, `base26`, `base32`, `base36`, `base49`, `base52`, `base58`, `base62`, `base64`
+* `length` - the length of the hash in chars
+
+**Defaults:** `[md5:contenthash:hex:9999]`
```
sibiraj-s: @jantimon I have mentioned `9999` as defaults. It corresponds to the max-length that is provided as input. but the length varies accordingly to `hashType`. Do You think It will create any confusion.

if so, `length` can be renamed to `maxLength`?.

jantimon: Probably - what exactly would 9999 mean?

sibiraj-s: It refers to the `maxLength` for the hash by loader-utils https://github.com/webpack/loader-utils/blob/master/lib/getHashDigest.js#L41

PR #1093, opened 2018-10-26
chore: No longer leave a comment when issues/PRs are locked

### LINKED ISSUE #1085: Lock bot generates a lot of noise when watching the repository
Hi!

Currently the lock bot is configured to lock both closed issues and pull requests after 30 days of inactivity, and to leave a comment when it does so:
https://github.com/jantimon/html-webpack-plugin/blob/65879ebe5061d821c9cbc9639915d855a612851b/.github/lock.yml

Example of locking + comment:
https://github.com/jantimon/html-webpack-plugin/issues/1039#issuecomment-431575466

This generates a lot of GitHub notification emails if watching the repository - enough that I've almost unwatched the repo on several occasions.

@jantimon, now that lock bot supports setting a label at time of locking - would you be open to using an informative label instead of a comment? (eg `locked-for-inactivity` or similar). The supported options are here:
https://github.com/dessant/lock-threads#configuration

Also, could PRs perhaps be excluded from locking? I would imagine it's much more common for unrelated comments to be added to old issues, than to old PRs - and for PRs it's more likely that there may be valid implementation questions at a later date.

Many thanks :-)

edmorley:
Since it creates too much email notification noise, and the reason for locking should be clear enough on its own (it's performed by a username of `lockbot` and the status of the issue/PR is set to `"resolved"`).

I decided against adding a label, since it will only end up adding noise to the labels list and issue/PR search results that adds little value other than as a workaround for inability to set a custom GitHub lock reason.

Fixes #1085.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -7,16 +7,8 @@ exemptLabels: []
 lockLabel: false
 
 # Comment to post before locking. Set to `false` to disable
-lockComment: >
-  This issue has been automatically locked since there has not been
-  any recent activity after it was closed. Please [open a new issue](https://github.com/jantimon/html-webpack-plugin/issues/new) for
-  related bugs.
+lockComment: false
 
 # Number of days of inactivity before a closed issue or pull request is locked
 daysUntilLock: 30
 
-pulls:
-  lockComment: >
-    This pull request has been automatically locked since there has not been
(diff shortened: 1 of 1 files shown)


jantimon:
Sorry was quite busy last week :)  
Thanks alot!

edmorley:
No worries - thank you :-)

Issue #1089, opened 2018-10-24
Overwriting compiler cache causes breaks webpack 5


niieani:
## Expected behaviour

html-webpack-plugin shouldn't crash when building/rebuilding in webpack 5

## Current behaviour

This part of `compiler.js` overwrites webpack's internal `cache` property with a plain object:

https://github.com/jantimon/html-webpack-plugin/blob/65879ebe5061d821c9cbc9639915d855a612851b/lib/compiler.js#L122-L132

In Webpack 5 `cache` is no longer an Object, but an instance of `Cache`:

https://github.com/webpack/webpack/blob/477f786781faed10d19dcd6582f83b3230d4a233/lib/Compiler.js#L132

This causes Webpack to crash, since it's expecting something else than a plain object.

## Suggested fix

EDIT: Update from @sokra:

> [This](https://github.com/jantimon/html-webpack-plugin/blob/65879ebe5061d821c9cbc9639915d855a612851b/lib/compiler.js#L122-L132) can be removed. It could also be removed for webpack 4.

-----

Original suggestion:

My suggestion would be to replace this line:

https://github.com/jantimon/html-webpack-plugin/blob/65879ebe5061d821c9cbc9639915d855a612851b/lib/compiler.js#L128

With:

```js
compilation.cache[compilerName] = new compilation.cache.constructor();
```

This will be both:
- backwards compatible, because `new {}.constructor()` is also an empty object
- forwards compatible, because `new compilation.cache.constructor()` will create a new instance of Webpack's Cache.

Ideally though, we probably shouldn't store the child compiletion's Caches as properties of the main Cache, but perhaps keep them in an internal Map?

## Environment

Tell us which operating system you are using, as well as which versions of Node.js, npm, webpack, and html-webpack-plugin. Run the following to get it quickly:

```
Node.js 10.11.0
webpack#next (from GitHub)
html-webpack-plugin#master (from GitHub)
```

### Config

Any webpack.config.js will reproduce the issue, e.g.:

```js
module.exports = {
  entry: 'app.js',
  output: {
    path: 'dist',
    filename: 'index_bundle.js'
  },
  plugins: [
    new HtmlWebpackPlugin(),
  ]
}
```

bbrzoska:
@jantimon yeah, let's do it that way 👍

jantimon:
Fixed with #1102

PR #1102, opened 2018-11-09
feat: drop workaround for "Uncaught TypeError: __webpack_require__(….) is not a function" to be compatible with webpack 5

jantimon:
See https://github.com/jantimon/html-webpack-plugin/issues/1089

@niieani could you please take a look if this would be the required change?


jantimon:
Thanks @sokra !

niieani:
Thanks for merging. It should work, I'll test it out on our app soon.

jantimon:
Released in html-webpack-plugin@4.0.0-beta.5

Issue #999, opened 2018-07-10
What is the FAQ?


chemoish:
## Expected behaviour

I would expect FAQ to be present or the comment to be improved (i.e. link to `_.template` or something?).

## Current behaviour

`README.md` states: `// Load a custom template (lodash by default see the FAQ for details)`

jantimon:
Looks like the link got lost somehow:
https://github.com/jantimon/html-webpack-plugin/blob/master/docs/template-option.md

PR #1107, opened 2018-11-11
fix(README): adds a link to template option documentation

panstav:
The readme file currently states that there's a FAQ document somewhere. Searching around I went to the issues tracker and found the link to the correct location in this [issue](https://github.com/jantimon/html-webpack-plugin/issues/999).


jantimon:
Thanks 👍

Issue #1098, opened 2018-11-06
Regression in 4.0.0-beta.2: script urls no longer point to publicPath


ahettlin:
I have upgraded from v3.2.0 to v4.0.0-beta.2 in order to get injection of bundles generated via splitChunks to work as suggested in other issues.  However, it appears that the behavior of the plugin has changed with regard to how the urls for injected scripts are generated.  Before, it seemed that the urls were just `publicPath + bundleName.js` (I have not looked at the source code of either version yet). With the new version, the url is now a relative path from the template location to the js file, which is incorrect when the file is served by our server.

Our urls changed from 
```html
<script type="text/javascript" src="app/main/ccf-app.js"></script>
```
to
```html
<script src="../../webapp/app/main/ccf-app.js"></script>
```

I was unable to find any documentation on how to customize the url, if there is any.  I realize this is a beta version. Is this intentional? Is there any way to get the old behavior back?

### Config

```js
module.exports = {
  entry: {
    'ccf-app':	'./src/main/javascript/main/main.ts'
  },
  output: {
    path: helpers.root('src/main/webapp/app/main'),
    publicPath: 'app/main/',
    filename: '[name].[hash].js',
    chunkFilename: '[name].[chunkhash].chunk.js'
  },
  module: {
    rules : [...]
  }
  plugins: [
    new HtmlWebpackPlugin({
      template : './src/main/templates/index.ejs',
      filename: '../../../resources/templates/index.ftl',
      inject: 'body',
      chunksSortMode: 'none',
      minify: false
    }),
    ...
  ],
  optimization: {
    removeAvailableModules: false,
    removeEmptyChunks: true,
    splitChunks: {
      chunks: 'all'
    }
  }
}
```

## Environment

Node.js: 10.13.0
npm: 6.4.1
webpack: 4.25.0
html-webpack-plugin: 4.0.0-beta.2
OS: win32 10.0.17758

Wind4:
In the v4 version, if `webpackConfig.output.publicPath` is set to a relative path, it will resolve to the relative path of entry, otherwise to `publicPath + bundleName.js`.

Can you set `webpackConfig.output.publicPath = '/app/main/'`?

jantimon:
Otherwise you would have to copy the result from the dist folder to the correct location after generating the html

ahettlin:
[...]

PR #1114, opened 2018-11-23
fixed: relative publicPath

jinasonlin:
#1098

### DIFF
--- a/index.js
+++ b/index.js
@@ -545,9 +545,11 @@ class HtmlWebpackPlugin {
      * if a path publicPath is set in the current webpack config use it otherwise
      * fallback to a realtive path
      */
-    let publicPath = compilation.mainTemplate.getPublicPath({hash: compilationHash}).trim() !== ''
+    const webpackPublicPath = compilation.mainTemplate.getPublicPath({hash: compilationHash});
+    const isPublicPathDefined = webpackPublicPath.trim() !== '';
+    let publicPath = isPublicPathDefined
       // If a hard coded public path exists use it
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -557,13 +542,13 @@ class HtmlWebpackPlugin {
 
     /**
      * @type {string} the configured public path to the asset root
-     * if the absolute path publicPath is set in the current webpack config use it otherwise
+     * if a path publicPath is set in the current webpack config use it otherwise
      * fallback to a realtive path
      */
-    let publicPath = this.isAbsolutePath(compilation.options.output.publicPath)
-      // If the absolute path is set in the publicPath use it
+    let publicPath = compilation.mainTemplate.getPublicPath({hash: compilationHash}).trim() !== ''
```
jantimon: Will `compilation.mainTemplate.getPublicPath({hash: compilationHash})` always return a string?

Maybe we could change it to:

```
const webpackPublicPath = compilation.mainTemplate.getPublicPath({hash: compilationHash});
const isPublicPathDefined = webpackPublicPath !== undefined && String(webpackPublicPath).trim() !== '';
```

That way we don't have to run `compilation.mainTemplate.getPublicPath({hash: compilationHash})` twice.

jinasonlin: `compilation.mainTemplate.getPublicPath({hash: compilationHash})` now always return a string. 

https://github.com/webpack/webpack/blob/8e841feda3ef32c64df1a175d0d9f71e47cfbe8e/lib/MainTemplate.js#L500

It will be right :)
```javascript
const isPublicPathDefined = webpackPublicPath.trim() !== '';
```

PR #1116, opened 2018-11-30
update to v4

pungggi:
Exmample of plugin ist still on old api


jantimon:
Actually that is the new api

pungggi:
HtmlWebpackPlugin.getHooks(compilation).beforeEmit

does not work with webpack 4 ..

pungggi:
This answers a lot of questions i had.. Thank you

PR #1132, opened 2019-01-01
Fix TypeScript types

### LINKED ISSUE #1108: Missings `typings` property in `package.json` in beta
## Expected behaviour

Types load correctly.

## Current behaviour

Types do not load.

## Environment

Latest TS version on node 10.12.

### Config

```ts
import HTMLPlugin from 'html-webpack-plugin';
```

## Relevant Links

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

wingrunr21:
This resolves #1108

* Adds `types` property to `package.json` (see [Publishing docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html))
* Revamps the `typings.d.ts` file to more accurately reflect both the internal and external API. Specifically, the `Options` interface should have every property marked optional (for external consumption) but most of those properties are assumed to be not null/undefined throughout `index.js`. I thus introduced an `InternalOptions` interface which matches this requirement.
* Moved the `Hooks` typedef into the `typings.d.ts` file vs defining it inline.
* Moved `HtmlTagObject` inside the `HtmlWebpackPlugin` namespace.

I did not fix compatibility with `noImplicitAny` in this PR. As you are shipping an annotated JS file, I believe TS users downstream with both `noImplicitAny` and `allowJs` enabled will get a bunch of compiler warnings.

cc @SimonSchick

### DIFF
--- a/index.js
+++ b/index.js
@@ -630,7 +630,7 @@ class HtmlWebpackPlugin {
    *
    * @param {string|false} faviconFilePath
    * @param {WebpackCompilation} compilation
-   * @parma {string} publicPath
+   * @param {string} publicPath
    * @returns {Promise<string|undefined>}
    */
   getFaviconPublicPath (faviconFilePath, compilation, publicPath) {
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 // Import types
 /** @typedef {import("./typings").HtmlTagObject} HtmlTagObject */
 /** @typedef {import("./typings").Options} HtmlWebpackOptions */
-/** @typedef {import("./typings").InternalOptions} HtmlWebpackInternalOptions */
+/** @typedef {import("./typings").ProcessedOptions} ProcessedHtmlWebpackOptions */
 /** @typedef {import("./typings").TemplateParameter} TemplateParameter */
 /** @typedef {import("webpack/lib/Compiler.js")} WebpackCompiler */
 /** @typedef {import("webpack/lib/Compilation.js")} WebpackCompilation */
@@ -36,7 +36,7 @@ class HtmlWebpackPlugin {
     const userOptions = options || {};
 
     // Default options
-    /** @type {HtmlWebpackInternalOptions} */
+    /** @type {ProcessedHtmlWebpackOptions} */
     const defaultOptions = {
       template: path.join(__dirname, 'default_index.ejs'),
       templateContent: false,
@@ -57,7 +57,7 @@ class HtmlWebpackPlugin {
       xhtml: false
     };
 
-    /** @type {HtmlWebpackInternalOptions} */
+    /** @type {ProcessedHtmlWebpackOptions} */
     this.options = Object.assign(defaultOptions, userOptions);
 
     // Default metaOptions if no template is provided
@@ -922,7 +922,7 @@ class HtmlWebpackPlugin {
      headTags: HtmlTagObject[],
      bodyTags: HtmlTagObject[]
    }} assetTags
- * @param {HtmlWebpackInternalOptions} options
+ * @param {ProcessedHtmlWebpackOptions} options
  * @returns {TemplateParameter}
  */
 function templateParametersGenerator (compilation, assets, assetTags, options) {
--- a/index.js
+++ b/index.js
@@ -1,8 +1,11 @@
 // @ts-check
 // Import types
-/* eslint-disable */
-/// <reference types="webpack" />
-/* eslint-enable */
+/** @typedef {import("./typings").HtmlTagObject} HtmlTagObject */
+/** @typedef {import("./typings").Options} HtmlWebpackOptions */
+/** @typedef {import("./typings").InternalOptions} HtmlWebpackInternalOptions */
+/** @typedef {import("./typings").TemplateParameter} TemplateParameter */
+/** @typedef {import("webpack/lib/Compiler.js")} WebpackCompiler */
+/** @typedef {import("webpack/lib/Compilation.js")} WebpackCompilation */
 'use strict';
 
 // use Polyfill for util.promisify in node versions < v8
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -28,14 +29,14 @@ const fsReadFileAsync = promisify(fs.readFile);
 
 class HtmlWebpackPlugin {
   /**
-   * @param {Partial<HtmlWebpackPluginOptions>} [options]
+   * @param {HtmlWebpackOptions} [options]
```
jantimon: Could we find a better name here?  
Maybe `HtmlWebpackPluginOptionsWithDefaults` or something similar?

index.js
```diff
@@ -1,8 +1,9 @@
 // @ts-check
 // Import types
-/* eslint-disable */
-/// <reference path="./typings.d.ts" />
```
jantimon: Really cool 👍

lib/hooks.js
```diff
@@ -12,67 +10,6 @@
 
 const AsyncSeriesWaterfallHook = require('tapable').AsyncSeriesWaterfallHook;
 
-// The following typedef holds the API definition for all available hooks
```
jantimon: The idea was to explain which kind of hooks can be used.  

Could we at least also move this comment to the typings file?

typings.d.ts
```diff
@@ -1,139 +1,247 @@
...
    */
-  attributes: {
-    [attributeName: string]: string|boolean
-  },
-  /**
-   * Wether this html must not contain innerHTML
-   * @see https://www.w3.org/TR/html5/syntax.html#void-elements
-   */
-  voidTag: boolean,
+  interface InternalOptions extends Required<Omit<Options, "minify">> {
```
jantimon: Could you please add a short comment explaining what you are doing here and especially why minify is not required?  
This might save the next one some time :)

index.js
wingrunr21: That's not what's getting passed in here though. This is the constructor argument which is the `Options` from the type definition file. Lines 60/61 is where the "options with defaults" part is setup which (which I named `HtmlWebpackInternalOptions`)

lib/hooks.js
wingrunr21: Ah. Since that also applies outside of the TS world probably shouldn't go in the typings file. I'll add some documentation back in.

typings.d.ts
wingrunr21: Sure. `Required` and `Omit` already have comments regarding their origins further up the file but I can add some additional context.

wingrunr21:
Sorry about the Options sorting thing. Should have dropped that in an independent commit. I'll push some changes!

index.js
jantimon: You are right - I meant `HtmlWebpackInternalOptions`

wingrunr21: `HtmlWebpackPluginConfiguration`?

I'm not a fan of `HtmlWebpackOptionsWithDefaults` as the interface doesn't actually have defaults.

lib/hooks.js
wingrunr21: Added back in via 0d0160f

index.js
jantimon: But actually it holds the `UserOptions` + `defaults` or is that wrong?

wingrunr21: It's just an interface that specifies that all fields are required + a change in how `minify` should look. The interface is actually used to describe both the `defaults` and the `user options` + `defaults` given both use cases need to have all fields as required.

jantimon: What about `ProcessedHtmlWebpackOptions` ?

wingrunr21: sure. I just made the change

jantimon:
Just realized that the `Require<Omit` does not work. All properties are still optional.

wingrunr21:
Hmm, I tested it locally. `Omit` is for sure working. Let me look at it.

wingrunr21:
Ok, scratch that. Looks like `Omit` is the one causing the problem. That's odd as that's the official definition of the type ([see here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html)). I'll keep looking.

wingrunr21:
Ok, so apparently I didn't need to redefine `Required` as in TS 2.8+ it is part of `lib.d.ts`.

The other part is apparently you can remove the optionality aspect of a type but you can't remove its ability to be set to undefined (see [here](https://github.com/Microsoft/TypeScript/issues/24067)).

So, this approach may not actually solve the use case I was trying to solve. As it's still internal it may not be that big of a deal but...

jantimon:
Well right now the types were used to guarantee the quality and to reduce bugs..  
Why would that be not a big deal?

wingrunr21:
Because they weren't accurate to begin with. If you're that worried about it feel free to revert this PR. Your TS support for downstream users is completely broken though. How are they supposed to guarantee quality and reduce bugs?

wingrunr21:
I'm not going to debate this, sorry. I think #1108  + the explanations in this PR cover the reasons.

So, looks like I spoke too soon. I poked at this some more and the problem is actually the wildcard `[option: string]: any` line at the bottom of the type. If you comment that out then the `Required<Omit` stuff works the way it is supposed to.

You can see a simple example [here on the TS playground](http://www.typescriptlang.org/play/index.html#src=interface%20Foo%20%7B%0D%0A%20%20%20%20bar%3F%3A%20string%0D%0A%20%20%20%20baz%3F%3A%20string%0D%0A%20%20%20%20%5Boption%3A%20string%5D%3A%20any%0D%0A%7D%0D%0A%0D%0Atype%20Omit%3CT%2C%20K%20extends%20keyof%20T%3E%20%3D%20Pick%3CT%2C%20Exclude%3Ckeyof%20T%2C%20K%3E%3E%3B%0D%0A%0D%0Ainterface%20Bar%20extends%20Required%3COmit%3CFoo%2C%20%22bar%22%3E%3E%20%7B%0D%0A%20%20%20%20bar%3A%20number%0D%0A%7D%0D%0A%0D%0Aconst%20f%3A%20Foo%20%3D%20%7B%7D%0D%0Aconst%20b%3A%20Bar%20%3D%20%7B%0D%0A%20%20%20%20bar%3A%205%0D%0A%7D). Comment out line 4 and you can see the errors change accordingly.

jantimon:
I guess the cleanest way would be to write two interfaces or to start with the ProcessedOption and use Partial again

jantimon:
Before we relied on the webpack source code - now we rely on @types/webpack

wingrunr21:
> Before we relied on the webpack source code - now we rely on @types/webpack

No it doesn't. See https://github.com/jantimon/html-webpack-plugin/blob/master/index.js#L7-L8

jantimon:
But why did we have to add @types/webpack ?

wingrunr21:
https://github.com/jantimon/html-webpack-plugin/blob/master/typings.d.ts#L1

jantimon:
Thanks good hint :) 

I’ll try to refactor those changes on a new or tomorrow

jantimon:
@wingrunr21 I fixed the remaining open points - would love to get your feedback:

https://github.com/jantimon/html-webpack-plugin/commit/f4cb241157a9a1fed4721b1abc1f390b09595494

PR #1160, opened 2019-02-07
introduce support for <base> (matching <title> and <meta>)

atstp:
this will let users inject a `base` tag in a manner similar to the current implementation for `title` and `meta`. 

It also includes [docs](https://github.com/atstp/html-webpack-plugin/tree/add-base-support#base-tag) along with two new tests (`adds a base tag with attributes` and `adds a base tag short syntax`) for the introduced features.

I didn't see a contributing guide, so if you have any other questions or requests just let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Allowed values are as follows
 |**`inject`**|`{Boolean\|String}`|`true`|`true \|\| 'head' \|\| 'body' \|\| false` Inject all assets into the given `template` or `templateContent`. When passing `true` or `'body'` all javascript resources will be placed at the bottom of the body element. `'head'` will place the scripts in the head element|
 |**`favicon`**|`{String}`|``|Adds the given favicon path to the output HTML|
 |**`meta`**|`{Object}`|`{}`|Allows to inject `meta`-tags. E.g. `meta: {viewport: 'width=device-width, initial-scale=1, shrink-to-fit=no'}`|
+|**`base`**|`{Object\|String\|false}`|`false`|Inject a [`base`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) tag. E.g. `base: "https://example.com/path/page.html`|
 |**`minify`**|`{Boolean\|Object}`|`true` if `mode` is `'production'`, otherwise `false`|Controls if and in what ways the output should be minified. See [minification](#minification) below for more details.|
 |**`hash`**|`{Boolean}`|`false`|If `true` then append a unique `webpack` compilation hash to all included scripts and CSS files. This is useful for cache busting|
 |**`cache`**|`{Boolean}`|`true`|Emit the file only if it was changed|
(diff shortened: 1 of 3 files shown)


jantimon:
Why wouldn’t that be possible with the webpack publicPath option?

atstp:
### `base` and `publicPath`

It's not possible with `publicPath` because `base`, aside from a small scenario, serves a different role.  They are similar in that they both can affect how paths are interpreted, but that's about it.  How they do it, what they can affect, and the context they work in are all different. Here's a few significant differences:

* Webpack's `publicPath` sets a prefix string only for assets that webpack manages.
* The `base` tag's `href` determines how all relative urls are resolved, including relative hyperlinks. 
* There's also `base`'s `target` attribute. If it's set, it specifies the default browsing context when navigating via links and forms (e.g. setting `target=_blank` in `base` will cause links to open in a new tab).

### first-class support vs. templates vs. plugins

It's is possible with templates, or plugins for that matter.  But, implementing it in html-webpack-plugin was cleaner and less obtuse than the alternatives.

Here's our use-case if you're curious:

* We didn't need templates for anything else
  * Introducing a file just to get a single line seemed heavy handed 
* We use `base` (and sometimes its `target`) conditionally, depending on where we're deploying, thanks to some legacy requirements.
  * so if we did add a template it would need a bit of logic and new template _plus_ template-logic seemed even less appealing
  * it would be nice to as much build configuration in `webpack.config.js`. moving decision making to templates isn't ideal.
* We're doing it with a plugin (for now)
  * it seems goofy to hook into webpack's compilation just to add a line that doesn't involve any assets being compiled

On the user side, for each alternative above, the implementation was about as long and significantly more obtuse than implementing the same thing in html-webpack-plugin. 

Alternatives aside, specifying `base` would fit well as an option to html-webpack-plugin in its own right. `title`, `meta`, and `base` are unique HTML elements: They're the only standardized `head` elements that are configured with a few short strings and don't load external assets. html-webpack-plugin already supports the first two (for obvious reasons); supporting `base` would make sense for the same reasons, and would make html-webpack-plugin's behavior consistent.

Hope it helps, and if you've got any further questions, let me know.

README.md
```diff
@@ -387,6 +388,39 @@ plugins: [
...
+})
+```
+
+The `target` can be specified with the corresponding key:
+
+```js
+new HtmlWebpackPlugin({
+  'base': {
+    'href': 'http://example.com/some/page.html',
+    'target': '_blank'
```
jantimon: Would this example be vulnerable to target blank hack attempts?

atstp: In the worst case, copy-pasting this example would only end in a few broken links. IANA [reserves example.com](https://www.iana.org/domains/reserved) for documentation. 

I'd trust IANA with a domain more than I'd trust myself, and if `example.com` were to ever act maliciously, `target="_blank"` would be near the bottom on a list of concerns.

-----

Also, I just ran through the most commonly described phishing scenario on firefox and chrome; both prevent setting `window.opener.location` when opened in a `_blank` context defined by `base`:

on firefox: 

```js
window.opener.location = 'https://phishing-origin.com'
// NS_ERROR_DOM_PROP_ACCESS_DENIED: Access to property denied
```

and on chrome: 

```js
window.opener.location = 'http://phishing-origin.com'
// Unsafe JavaScript attempt to initiate navigation for frame with origin
// 'https://www.original-safe-origin.com' from frame with URL 'https://www.compromised-origin.com/'. 
// The frame attempting navigation is neither same-origin with the target, nor is it the target's parent or opener.
```

jantimon: Okay looks like it is not vulnerable 👍 so that should be no concern

jantimon:
Right now this feature would not work if inject is false - I think that should be okay.. or do you think we should also add it to the default template?

atstp:
Yeah, I understand your caution: the real cost comes from maintenance.

For what it's worth, you're welcome to ping me here on github if this needs attention in the future. I'll be happy to write up a patch or field questions (i'll be doing that no matter what, this PR just affects where).

It's your project though, so we'll work with whatever your final decision is.

Either way, thanks for actively maintaining this project.

atstp:
I agree. The `inject: false` behavior seems fine as is.

PR #1167, opened 2019-02-17
fix: Use src/index.ejs by default if present

jantimon:
BREAKING CHANGE: Use `src/index.ejs` if no template option is set.

Webpack will pick up `src/` by default that means depending on your resolving settings it will use `src/index.js`, `src/index.jsx`, `src/index.ts` ... or similar by default.

This pull request adds the same feature for the html-webpack-plugin.

I would love to know what you think in general about this pull request and if we should support:

`src/index.ejs`
`src/index.html`
or
`public/index.html`

cc @sokra @edmorley

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,7 +38,7 @@ class HtmlWebpackPlugin {
     // Default options
     /** @type {ProcessedHtmlWebpackOptions} */
     const defaultOptions = {
-      template: path.join(__dirname, 'default_index.ejs'),
+      template: 'auto',
       templateContent: false,
       templateParameters: templateParametersGenerator,
       filename: 'index.html',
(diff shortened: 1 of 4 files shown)


sokra:
:+1:

index.js
```diff
@@ -908,6 +908,12 @@ class HtmlWebpackPlugin {
    * The webpack base resolution path for relative paths e.g. process.cwd()
    */
   getFullTemplatePath (template, context) {
+    if (template === 'auto') {
+      template = path.resolve(context, 'src/index.ejs');
```
jeremenichelli: I wouldn't override the value coming from the argument, I think is confusing in terms code readability. I suggest doing:

```js
getFullTemplatePath (template, context) {
  if (template === 'auto') {
    let templatePath = path.resolve(context, 'src/index.ejs');
    if (!fs.existsSync(templatePath)) {
     return path.join(__dirname, 'default_index.ejs');
  }
}
```

and later conusme the value return from the method, that way makes the function pure :)

jantimon: Thanks for your feedback 👍
We can't return early here because of the prefixing and path normalisation.

jeremenichelli: Oh ok got it, was pretty sure I was missing context from the rest of the code. But would be cool to refactor in the future. Thanks for the explanation.

Issue #1186, opened 2019-03-29
RangeError: Maximum call stack size exceeded


CT1994:
## Expected behaviour

html-webpack-plugin should build without any error

## Current behaviour

when building more then 245 html-webpack-plugins you will get a RangeError

## Environment

operating system: Windows 10
Node.js version: 11.13.0
npm version: 6.7.0
webpack 4.29.6
html-webpack-plugin 4.0.0-beta.5

## Relevant Links

I have created an example branch with the current issue

https://github.com/CT1994/html-webpack-plugin-range-error

[example as zip](https://github.com/jantimon/html-webpack-plugin/files/3025390/html-webpack-plugin-range-error-master.zip)

jantimon:
OH 😨 

This looks horrible
I got the same thing on mac with `355` plugin instances.

```
Node.js v10.13.0
darwin 17.7.0
6.4.1
```



<details><summary>Click to expand Stacktrace</summary>
<p>

```
Unhandled rejection RangeError: Maximum call stack size exceeded
    at new Function (<anonymous>)
    at AsyncSeriesHookCodeFactory.create (node_modules/tapable/lib/HookCodeFactory.js:32:10)
    at AsyncSeriesHook.compile (node_modules/tapable/lib/AsyncSeriesHook.js:24:18)
    at AsyncSeriesHook._createCall (node_modules/tapable/lib/Hook.js:24:15)
    at AsyncSeriesHook.lazyCompileHook [as callAsync] (node_modules/tapable/lib/Hook.js:153:21)
    at Compiler.emitAssets (node_modules/webpack/lib/Compiler.js:441:19)
    at onCompiled (node_modules/webpack/lib/Compiler.js:241:9)
    at hooks.afterCompile.callAsync.err (node_modules/webpack/lib/Compiler.js:630:14)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (node_modules/tapable/lib/HookCodeFactory.js:32:10), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook (node_modules/tapable/lib/Hook.js:154:20)
    at compilation.seal.err (node_modules/webpack/lib/Compiler.js:627:30)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (node_modules/tapable/lib/HookCodeFactory.js:32:10), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook (node_modules/tapable/lib/Hook.js:154:20)
[...]

PR #1191, opened 2019-04-05
Update html webpack plugin to process all instance-of HtmlWebpackPlugins as a single operation

CT1994:
The changes in this patch is to fix the following issue.
https://github.com/jantimon/html-webpack-plugin/issues/1186

this adds the ability to process all html webpack plugins as a single plugin this fixes the maximum rangeError


CT1994:
ok here is what has changed

index.js

- This class does no handling of tappable hooks anymore, this has been moved into [multiHtmlWebpackPlugin.js](https://github.com/jantimon/html-webpack-plugin/pull/1191/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-8396ea829480953ae6aa07684d795419)

- This is to be able to process all instanceof html-webpack-plugin as one operation, this will help resolve the issue with maximum call stack error.

[optionsHelper.js](https://github.com/jantimon/html-webpack-plugin/pull/1191/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ts#diff-41bd2331495fbd36f05d88c47c11a9fb)

- Anything to do with html-webpack-plugin options has been moved to optionsHelper.js this is just a refactor of structure

[templateParametersGenerator.js](https://github.com/jantimon/html-webpack-plugin/pull/1191/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.png#diff-390aabe7a8ec525f2ef10648790dd560)

- this is just cleaning up index.js

jantimon:
Maybe @sokra could take a look at this solution and let us know why tapable needs such a big workaround here.

Maybe we are only using webpack tapable in a wrong way.

PR #1198, opened 2019-04-22
feat: allow for templates resolved from modules

stevematney:
In cases where a module might export a template file, this allows for importing from those modules directly.

This could also be accomplished at the downstream level with something like:
```es6
// ...
new HtmlWebpackPlugin({
  template: require.resolve('my-module/template.html'),
  // ...
})
```

With the new template loading functionality in the PR, that simply looks like:
```es6
// ...
new HtmlWebpackPlugin({
  template: 'my-module/template.html',
  // ...
})
```

### DIFF
--- a/__mocks__/fake-module/plain.html
+++ b/__mocks__/fake-module/plain.html
@@ -1,0 +1,10 @@
+<!doctype html>
+<html lang="en" manifest="foo.appcache">
+<head>
+  <meta charset="utf-8">
+  <title>Example Plain file</title>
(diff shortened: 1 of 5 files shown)


jantimon:
Sorry I totally missed this pr!

You can already do that by using webpacks `context` property

jantimon:
CSS Loader is using the following syntax to enforce `node_modules` loading:

`~my-module/template.html`

What about using the same for html-webpack-plugin?

PR #1201, opened 2019-04-25
Upgrade all dependencies to latest stable versions

AviVahl:
adapted code to newer semi-standard (used --fix)
rebuilt examples using newest webpack


AviVahl:
Hey @jantimon, any chance you can review this? The upgrade to `html-minifier@4` will rid of some duplicates we have in our `node_modules`.

jantimon:
Hey @AviVahl thanks for all the work! :)
Looks like the only breaking change is html-minifier which dropped node < 6.  
As we don't support that node version anyway that's absolutely fine for us.

Issue #1092, opened 2018-10-25
[v4-beta.2] Alternative to htmlPluginData.assets.chunks?


jeffposnick:
I'm working on getting `preload-webpack-plugin` updated to work with `html-webpack-plugin` v4. The current WIP code is at https://github.com/GoogleChromeLabs/preload-webpack-plugin/tree/html-plugin-updates

One issue that I'm running into, which is alluded to in https://github.com/jantimon/html-webpack-plugin/pull/953#issue-189369685, is that previous versions of `html-webpack-plugin` would pass in plugin data to `htmlWebpackPluginAfterHtmlProcessing` that included an `assets` as a property, and in that `assets` object, `chunks`.

The release notes say:

> The assets object which is used for the template parameters and inside hooks was changed. The chunks property was removed and the js and css property was converted from a string into an object `{ entryName: string, path: string}`

However, from what I'm seeing, after moving from `htmlWebpackPluginAfterHtmlProcessing` to `beforeEmit`, the plugin data that's passed to my callback doesn't have `assets` set at all. (I'm being passed a plugin data object with 3 properties: `html`, `plugin`, and `outputName`.)

How can I get the equivalent of `pluginData.assets.chunks` in v4?

Or, more to the point, why isn't `assets` being set? Could it be related to the issues I'm having tapping into the right hook (see https://github.com/jantimon/html-webpack-plugin/issues/1091)?

(See https://github.com/GoogleChromeLabs/preload-webpack-plugin/blob/c57b0b3521ca31091e0a77784bdb56a8ae8ba98d/src/index.js#L156-L159 for the context of where I'm trying to get that data.)

jeffposnick:
Here's how that info was being used when working with `html-webpack-plugin` v3:

https://github.com/GoogleChromeLabs/preload-webpack-plugin/blob/798fc677f42e9ed2c1176065f733efa6b7fdc735/index.js#L172-L190

The main sticking point here is that the `preload-webpack-plugin` allowed white/blacklists to operate on the basis of `chunk` names, and in order to apply that (especially in a model where there might be multiple `html-webpack-plugin` instances for a given compilation), we needed to know which chunk names (not asset names) were associated with a given `html-webpack-plugin` instance.

jantimon:
For performance reasons we switched from the `chunks` to `entries` - which is very similar but a different webpack api.

The idea of restructuring the hooks is to provide only one truth but I understand your need.  
[...]

PR #1202, opened 2019-04-26
Restore chunk info to template parameters

### LINKED ISSUE #1158: Ability to indicate different inject position ('body' or 'head') to different chunk.
Feature request - I want to inject two chunks - one chunk in head and another in body.

```
module.exports = {
  entry: {
    index: "./index.ts",
    index_head: "./index.css" // adding "_head" at the end to inject in head. (injects in head)
  },
  output: {
    path: "./dist",
    filename: "[name].bundle.js"
  },
  plugins: [
    new HtmlWebpackPlugin({
      template: "./index.html",
      filename: "./dist/index.html",
      chunks: ["index", "index_head"]
    }),
    new HtmlWebpackInjector()
  ]
}
```

And I get the output as:

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Archit's App</title>
    <script type="text/javascript" src="index_head.bundle.js"></script></head>
  </head>
  <body>
    <script src="index_bundle.js"></script> 
  </body>
</html>
```

jimmydief:
This PR essentially reverts the change from 37db0868efdbf334a1b60003fe5bd376cfd8ae01 that removed `chunks` from the asset parameters. In doing so, it should resolve #1158 and #1092. My hope is that using `toJson({all: false, chunkGroups: true})` with a filter for only the info we need is enough to mitigate the performance concerns that caused it to be removed previously. I benchmarked our build (6000 modules, 27 `HtmlWebpackPlugin` instances) and saw no change in duration with this change.

My use case for this is rendering `preload` tags, grouped by chunk, into the HTML template of our single page app. Each chunk's preload tags are inside of a conditional block that gets included in the final HTML rendered by the server only if that chunk is used by the initial route. Doing this allows us to start downloading route-specific chunks immediately before the main chunk has been dowloaded and parsed.

Our template that we feed into HtmlWebpackPlugin looks like this:
```handlebars
<!DOCTYPE html>
<html>
	<head>
		<meta charset="utf-8"/>
		<title>Test</title>

		<if isTrue="{! route='one'}">
			{{#each htmlWebpackPlugin.files.chunks.one.css as |file|}}
			<link rel="preload" href="{{file}}" as="style" />
			{{/each}}
			{{#each htmlWebpackPlugin.files.chunks.one.js as |file|}}
			<link rel="preload" href="{{file}}" as="script" />
			{{/each}}
		</if>
		<if isTrue="{! route='two'}">
			{{#each htmlWebpackPlugin.files.chunks.two.css as |file|}}
			<link rel="preload" href="{{file}}" as="style" />
			{{/each}}
			{{#each htmlWebpackPlugin.files.chunks.two.js as |file|}}
			<link rel="preload" href="{{file}}" as="script" />
			{{/each}}
		</if>

		{{#each htmlWebpackPlugin.files.css as |file|}}
		<link rel="stylesheet" href="{{file}}" type="text/css" />
		{{/each}}
	</head>
	<body>
		{{#each htmlWebpackPlugin.files.css as |file|}}
		<script src="{{file}}"></script>
		{{/each}}
	</body>
</html>

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,11 +263,11 @@ The following variables are available in the template:
         "js": [ "assets/head_bundle.js", "assets/main_bundle.js"],
         "chunks": {
           "head": {
-            "entry": "assets/head_bundle.js",
+            "js": ["assets/head_bundle.js"],
             "css": [ "main.css" ]
           },
           "main": {
-            "entry": "assets/main_bundle.js",
+            "js": ["assets/main_bundle.js"],
             "css": []
           },
         }
--- a/index.js
+++ b/index.js
@@ -612,8 +612,13 @@ class HtmlWebpackPlugin {
 
     const stats = compilation.getStats().toJson({all: false, chunkGroups: true});
 
-    Object.entries(stats.namedChunkGroups).forEach(([chunkGroupName, chunkGroupStats]) => {
-      assets.chunks[chunkGroupName] = this.processFiles(chunkGroupStats.assets, publicPath, compilationHash, new Set());
+    Object.keys(stats.namedChunkGroups).forEach((chunkGroupName) => {
+      assets.chunks[chunkGroupName] = this.processFiles(
+        stats.namedChunkGroups[chunkGroupName].assets,
+        publicPath,
+        compilationHash,
+        new Set()
+      );
     });
 
     return assets;
--- a/index.js
+++ b/index.js
@@ -27,6 +27,8 @@ const getHtmlWebpackPluginHooks = require('./lib/hooks.js').getHtmlWebpackPlugin
 const fsStatAsync = promisify(fs.stat);
 const fsReadFileAsync = promisify(fs.readFile);
 
+const EXTENSION_REG_EXP = /\.(css|js)(\?|$)/;
+
 class HtmlWebpackPlugin {
   /**
    * @param {HtmlWebpackOptions} [options]
@@ -565,6 +567,7 @@ class HtmlWebpackPlugin {
     /**
      * @type {{
         publicPath: string,
+        chunks: {[chunkGroupName: string]: {js: Array<string>, css: Array<string>}},
         js: Array<string>,
         css: Array<string>,
         manifest?: string,
@@ -574,6 +577,8 @@ class HtmlWebpackPlugin {
     const assets = {
       // The public path
       publicPath: publicPath,
+      // Will contain all js and css files by chunk.
+      chunks: {},
       // Will contain all js files
       js: [],
       // Will contain all css files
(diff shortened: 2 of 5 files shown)


jantimon:
Hey @jimmydief this pull request looks really awesome 👍 

The  idea of that new structure was that if there are plugins altering the included js/css files this would not have to be done twice.

If I understand your proposed changes correctly you would revert that idea.
Would using a custom `templateParameters` function also solve your problem?

jimmydief:
Hey @jantimon, thanks! Using a custom `templateParameters` function was the first thing I considered but the public path and hash-related logic in `processFiles` isn't straight-forward and I was hoping to avoid having to reimplement that in user-land. Making the public path and plugin instance available to the template generator function would make it possible to reuse the logic in `processFiles`. 
```js
function templateParametersGenerator (compilation, assets, assetTags, options, plugin, publicPath) {
	const stats = compilation.getStats().toJson({all: false, chunkGroups: true});

	const chunks = {};
	Object.keys(stats.namedChunkGroups).forEach((chunkGroupName) => {
		chunks[chunkGroupName] = plugin.processFiles(
			stats.namedChunkGroups[chunkGroupName].assets,
			publicPath,
			compilation.hash,
			new Set()
		);
	});

	const xhtml = options.xhtml;
	assetTags.headTags.toString = function () {
		return this.map((assetTagObject) => htmlTagObjectToString(assetTagObject, xhtml)).join('');
	};
	assetTags.bodyTags.toString = function () {
		return this.map((assetTagObject) => htmlTagObjectToString(assetTagObject, xhtml)).join('');
	};

	return {
		compilation: compilation,
		webpackConfig: compilation.options,
		htmlWebpackPlugin: {
			tags: assetTags,
			files: {...assets, chunks},
			options: options
		}
	};
}
```

Similarly, it seems like this could also be possible as a plugin that hooks into `beforeAssetTagGeneration` if the public path and compilation were made available to that hook. That could look something like this:
```js
HtmlWebpackPlugin.getHooks(compilation).beforeAssetTagGeneration.tapAsync('ChunkInfoPlugin', (data, cb) => {
	const stats = data.compilation.getStats().toJson({all: false, chunkGroups: true});

	data.assets.chunks = {};

	Object.keys(stats.namedChunkGroups).forEach((chunkGroupName) => {
		data.assets.chunks[chunkGroupName] = data.plugin.processFiles(
			stats.namedChunkGroups[chunkGroupName].assets,
			data.publicPath,
			data.compilation.hash,
			new Set()
		);
	});

	cb(null, data)
})
```

The latter seems preferable to me.

jantimon:
I see I have to look into it - hopefully we can find a solution here.

`data.compilation.getStats()` is very slow so unfortunately we can't use it here

jimmydief:
Wound up going the plugin route, worked fine.

PR #1203, opened 2019-04-28
Plugins: add html-webpack-partials-plugin

colbyfayock:
Adds [html-webpack-partials-plugin](https://github.com/colbyfayock/html-webpack-partials-plugin) to the plugins list. Compatible with `html-webpack-plugin` latest `4.0.0-beta.5`


jantimon:
Wouldn't it be better to implement this in form of a loader to get all benefits from caching and file watching?

colbyfayock:
hm yeah, I'm not versed enough with webpack to know the specifics to those details, would have to spend more time on it, but what you say makes sense. the goal for this was just to extend what you've made to avoid having to create and maintain a custom template every time for simple things and seems to work well for that purpose. does the same thing apply to a custom template with the core plugin though? IIRC changes to that didn't trigger rebuilds 🤔but it's been a bit

PR #1207, opened 2019-05-05
Make it works with .mjs files (Closes #1126)

### LINKED ISSUE #1126: Make it works with .mjs files
Great approach with compiling of mjs files for modern browsers is described in [this article](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/). I faced with a problem when I specify `output.filename` in a webpack configuration file as a file with `.mjs` extension:
```
output: {
  filename: 'main.mjs',
  path: path.resolve(__dirname, 'public'),
},
```
Webpack is sucessfully compile all files but html-webpack-plugin does not append the generated files to the template (`htmlWebpackPlugin.files.js` is an empty array). It would be great to make it works with `.mjs` output files by default or with some flag in the configuration.
Expected result: the string `<script type="module" src="main.mjs"></script>` is appended in the template.

artemirq:


jantimon:
Sorry for the delay - Can you please update also the code comments?

artemirq:
@jantimon Updated PR

Issue #1244, opened 2019-07-12
Property 'apply' is missing in type 'HtmlWebpackPlugin' but required in type 'Plugin'


artemirq:
## Expected behaviour

No TypeScript errors

## Current behaviour

After [this](https://github.com/jantimon/html-webpack-plugin/commit/d4eb1c749316af3964126606fe6c70a233c30fef) commit, when I try build [my project](https://github.com/artemirq/zero-scriptsjs/pull/389) I get this error:

``` error TS2741: Property 'apply' is missing in type 'HtmlWebpackPlugin' but required in type 'Plugin'.```

## Environment

Node.js v12.1.0
win32 10.0.15063

## Additional context

I think, if we revert this commit, problem should be fixed :)

jantimon:
Using @types/webpack caused multiple problems - maybe we could just add the apply method?

artemirq:
@jantimon Good idea

MrStLouis:
I just ran into this as well, following for fix :)

jantimon:
Released in 4.0.0-beta.8

PR #1245, opened 2019-07-13
fix(typings.d.ts): added apply method type to HtmlWwbpackPlugin class definition

MrStLouis:
quick fix copied from [@types/html-webpack-plugin/index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/html-webpack-plugin/index.d.ts)
for #1244

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -1,10 +1,13 @@
 import { AsyncSeriesWaterfallHook } from "tapable";
+import { Compiler } from 'webpack';
 import { Options as HtmlMinifierOptions } from "html-minifier";
 
 export = HtmlWebpackPlugin;
 
 declare class HtmlWebpackPlugin {
   constructor(options?: HtmlWebpackPlugin.Options);
+
+  apply(compiler: Compiler): void;
 }
 
 declare namespace HtmlWebpackPlugin {


typings.d.ts
```diff
@@ -1,10 +1,13 @@
 import { AsyncSeriesWaterfallHook } from "tapable";
+import { Compiler } from 'webpack';
 import { Options as HtmlMinifierOptions } from "html-minifier";
 
 export = HtmlWebpackPlugin;
 
 declare class HtmlWebpackPlugin {
   constructor(options?: HtmlWebpackPlugin.Options);
+
+  apply(compiler: Compiler): void;
```
k2snowman69: Possible alternative is to extend Plugin

```
import { Plugin } from 'webpack'; 
declare class HtmlWebpackPlugin extends Plugin {
  constructor(options?: HtmlWebpackPlugin.Options);
```

Haven't tested compilability, just a thought off the top of my head

jantimon:
Thanks @MrStLouis - if the restarted test succeeds I will merge this :)

typings.d.ts
jantimon: This caused some issues the last time we tried so I would rather go with the solution @MrStLouis proposed

PR #1248, opened 2019-07-17
update a readme link to webpack compilation docs

PhysRex:


jantimon:
Maybe we should rather link to https://webpack.js.org/contribute/writing-a-plugin/#compiler-and-compilation ?

README.md
```diff
@@ -290,7 +290,7 @@ The following variables are available in the template:
 - `webpackConfig`: the webpack configuration that was used for this compilation. This
   can be used, for example, to get the `publicPath` (`webpackConfig.output.publicPath`).
 
-- `compilation`: the webpack [compilation](https://webpack.js.org/api/compilation/) object.
+- `compilation`: the webpack [compilation](https://webpack.js.org/api/compilation-hooks/) object.
```
hg-pyun: image

The webpack docs link to `compilation-hooks` when clicked compilation button.

jantimon: Yes but during render you don't want to use hooks but rather output data from the compilation like file size or file names

PR #1257, opened 2019-08-07
Properly encode file names emitted as part of URLs.

### LINKED ISSUE #1254: Url of chunk filenames are not `urlencoded`
## Expected behaviour

URI of files in emitted HTML should be "urlencoded".

## Current behaviour

URI of files in emitted to HTML as is, with no visible encoding.

### Config

```js

module.exports = {
    context: __dirname,
    entry: {
        "app+foo": "./index.js",
    },
    plugins: [
        new HtmlWebpackPlugin()
    ]
}
```

Expected that emitted `<script>` tag contains `src="app%2bfoo.js"`.
But, we see `src="app+foo.js"`, which points at "app foo.js" (space inside).

## Relevant Links

- Repro: https://github.com/zbigg/hwp-urlencode

## Environment

```
html-webpack-plugin@3.2.0 
webpack@4.38.0 
```

zbigg:
Attempt to fix #1254

### DIFF
--- a/index.js
+++ b/index.js
@@ -604,7 +604,7 @@ class HtmlWebpackPlugin {
       // E.g. bundle.js -> /bundle.js?hash
       const entryPointPublicPaths = entryPointFiles
         .map(chunkFile => {
-          const entryPointPublicPath = publicPath + chunkFile;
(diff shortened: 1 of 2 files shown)


jantimon:
Wow that’s really well done! Thanks a lot

PR #1260, opened 2019-08-11
Feature/chunks enhance

charlyzeng:
When I want to split bundle in Webpack4, I usually use `optimization.splitChunks`. I use `optimization.splitChunks.{cachedGroup}.minSize` to limit every bundle size, and this will result like this:

image

As you can see, `chunkName` is auto generated, so *HtmlWebpackPlugin*'s `chunks` should has a way to match this. Then I enhance the `chunks` by RegExp and Function.

The same is `excludeChunks`.


charlyzeng:
Supplementary explanation: I checkout my new branch from tag v3.2.0.

jantimon:
We really need to such an improvement - however it should be based on html-webpack-plugin 4.x

charlyzeng:
> We really need to such an improvement - however it should be based on html-webpack-plugin 4.x

It's my pleasure for you saying this. 

I tried *html-webpack-plugin 4.x*, and found it became more intelligent. *html-webpack-plugin 4.x* seems to be able to extract all related splited chunks by entry name, even splited chunks' name is auto generated with hash.

Then, there maybe no need to make `options.chunks` support RegExp and Function.

Thanks for your work, that greatly improves my and my colleagues' work efficiency.

PR #1270, opened 2019-09-09
update lodash dependency for prototype polution vulnerability

lfbayer:
This pull requests updates lodash to address this vulnerability: https://npmjs.com/advisories/1065

commitizen also needed to be updated since its older version also depended on a vulnerable version of lodash.


jantimon:
Thanks 👍

PR #1281, opened 2019-09-27
fix: use fileDependencies to support webpack5

eyeccc:
It seems that compilationDependencies is not supported in webpack 5. Need to change it to `fileDependencies`.


eyeccc:
maybe should wait for webpack 5 becoming stable

PR #1286, opened 2019-10-10
Add static variable `globalMeta`

Fju:
If you want to keep meta tags similiar across multiple HTML files, you would probably define a constant in your `webpack.config.js` and pass it to every instance - like so:
```js
const GLOBAL_META = {
  'viewport': '...',
  'theme-color': '...',
  // etc
}
...
plugins: [
  new HtmlWebpackPlugin({
    filename: 'a.html',
    template: 'src/a.html',
    meta: GLOBAL_META
  }),
  new HtmlWebpackPlugin({
    filename: 'b.html',
    template: 'src/b.html',
    meta: GLOBAL_META
  })
]
```
It is even more inconvenient if you want to overwrite these global settings for single `HtmlWebpackPlugin` instances.

This is why I introduced a static variable in the `HtmlWebpackPlugin` class called `globalMeta`. This variable is the same for all instances of `HtmlWebpackPlugin` and is applied with the instance specific `meta` option during construction.

**Example:**
```js
const HtmlWebpackPlugin = require('html-webpack-plugin');
// set static variable
HtmlWebpackPlugin.globalMeta = {
  'viewport': '...'
}
```
```js
plugins: [
  new HtmlWebpackPlugin({
    'filename': 'a.html',
    'template': 'src/a.html'
    'meta': { 'theme-color': '...' }
    // Will generate 'theme-color' and the global 'viewport' meta tag
  }),
  new HtmlWebpackPlugin({
    'filename': 'b.html',
    'template': 'src/b.html',
    'meta': {} // can be omitted, of course
    // Will generate the global 'viewport' meta tag
  })
]
```

A section in the `README.md` and a corresponding test was added to `basic.spec.js` as well.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ For the default template the html-webpack-plugin will already provide a default 
 
 Please take a look at this well maintained list of almost all [possible meta tags](https://github.com/joshbuchea/HEAD#meta).
 
+
 #### name/content meta tags 
 
 Most meta tags are configured by setting a `name` and a `content` attribute.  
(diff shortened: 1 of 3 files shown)


jantimon:
What would be the benefit of this approach?

Fju:
You wouldn't have to specify the `meta` attribute at all, if you specified the static meta variable.
For example:
``` js
HtmlWebpackPlugin.globalMeta = {
  'viewport': '...',
  'theme-color': '...'
}
// ...
plugins: [
  new HtmlWebpackPlugin({
    filename: 'a.html',
    template: 'src/a.html'
  }),
  new HtmlWebpackPlugin({
    filename: 'b.html',
    template: 'src/b.html'
  })
]
```

Instead of:
``` js
const GLOBAL_META = {
  'viewport': '...',
  'theme-color': '...'
}
// ...
plugins: [
  new HtmlWebpackPlugin({
    filename: 'a.html',
    template: 'src/a.html',
    meta: GLOBAL_META
  }),
  new HtmlWebpackPlugin({
    filename: 'b.html',
    template: 'src/b.html',
    meta: GLOBAL_META
  })
]
```

I know that it is not a killer feature, but for my use-case it is kind of handy. Would like to hear more thoughts about it :v:

PR #1295, opened 2019-10-24
docs: Add templateContent param back to README.md

msmy:
This change has been made because I noticed that `templateContent` had disappeared from the docs! Looks like it was [accidentally removed here?](https://github.com/jantimon/html-webpack-plugin/commit/1b3adb1b5a9c58ac9d060068758999196fba22d8#diff-04c6e90faac2675aa89e2176d2eec7d8L126) 

This commit will add the templateContent variable back from this [previously merged PR](https://github.com/jantimon/html-webpack-plugin/pull/1024) 

Also noticed a small typo, so corrected that one too :)


jantimon:
done in 4.2.0

PR #1298, opened 2019-10-28
Update template-option.md

qidaneix:
In 22th line of this docs, it says html-webpack-pulgin kicks in lodash loader, not ejs loader

### DIFF
--- a/docs/template-option.md
+++ b/docs/template-option.md
@@ -63,7 +63,7 @@ new HtmlWebpackPlugin({
 ```
 
 However this also means that in the following example webpack will use the [html loader for your template](https://webpack.js.org/loaders/html-loader/).
-This will **cause html minification** and it will also **disable the ejs fallback** loader.
+This will **cause html minification** and it will also **disable the lodash fallback** loader.
 
 ```js
 {


jantimon:
The idea of saying `ejs` is that both templates languages are very similar

qidaneix:
but ejs support include syntax but lodash.template donot.

docs/template-option.md
```diff
@@ -63,7 +63,7 @@ new HtmlWebpackPlugin({
 ```
 
 However this also means that in the following example webpack will use the [html loader for your template](https://webpack.js.org/loaders/html-loader/).
-This will **cause html minification** and it will also **disable the ejs fallback** loader.
+This will **cause html minification** and it will also **disable the lodash fallback** loader.
```
djragsdale:
```suggestion
This will **cause html minification** and it will also **disable the [lodash](https://github.com/difelice/ejs-loader) fallback** loader.
```

jantimon:
I updated it in https://github.com/jantimon/html-webpack-plugin/commit/a195c3441a79f5ad12d524f09759d9baf764d1c7 please let me know if you think it still needs further improvements.

Issue #1110, opened 2018-11-19
The HtmlWebpackPlugin.getHooks is not defined


yibingxiong:
## Expected behaviour

I want to use `HtmlWebpackPlugin.getHooks` to use hook function to do something. But I found `HtmlWebpackPlugin.getHooks` is not defined.  Not same with the docs. The version is 3.2.0. 

## Current behaviour

`HtmlWebpackPlugin.getHooks` is not defined.  Not same with the docs.

## Environment

- node v10.8.0
- webpack v4.25.1
- html-webpack-plugin v3.2.0
- npm v6.2.0

### Config

Copy the minimal `webpack.config.js` to produce this issue:

```js
var path = require('path');
const MyExampleWebpackPlugin = require('./myplugin');
const HtmlWebpackPlugin = require('html-webpack-plugin');
module.exports = {
  mode: 'development',
  entry: './test_plugin/index.js',
  output: {
    path: path.resolve(__dirname, 'dist'),
    filename: 'bundle.js'
  },
  plugins: [
    new HtmlWebpackPlugin(),
    new MyExampleWebpackPlugin()
  ]
};
```

Copy your template file if it is part of this issue:

no template file, use default.

## Relevant Links

This is for test, no public links

## Additional contex

```js
const HtmlWebpackPlugin = require('html-webpack-plugin');

class MyExampleWebpackPlugin {
    apply(compiler) {

        compiler.hooks.compilation.tap('MyPlugin', (compilation) => {
          console.log(typeof HtmlWebpackPlugin.getHooks); // outputs undefined, expected 'function'
        })
    }
}

module.exports = MyExampleWebpackPlugin;
```

jantimon:
`getHooks` requires html-webpack-plugin 4

yibingxiong:
Thanks. I know.

annayo:
I'm having the same issue. Using:

- node v10.8.0
- webpack v4.26.1
- html-webpack-plugin v3.2.0
- npm v6.2.0

@yibingxiong were you able to resolve it?

Dalqin:
> 
> 
> `getHooks` requires html-webpack-plugin 4

What do you mean? There highest version of html-webpack-plugin is 3.2.0.

jantimon:
https://www.npmjs.com/package/html-webpack-plugin/v/4.0.0-beta.4

Dalqin:
Oh it's a beta! Thanks for clarifying.

[...]

PR #1305, opened 2019-11-04
Add hooks document for version 3

hg-pyun:
Hello. There was a confusing part about the hooks of `html-webpack-plugin`. 

The API called `getHooks` is not yet supported in v3, but It was written in the document. The current version is v3 and I think we should inform about it in the document.

## Reference
- https://github.com/jantimon/html-webpack-plugin/issues/1110

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,6 +531,27 @@ about which values are passed.
 Example implementation: [webpack-subresource-integrity](https://www.npmjs.com/package/webpack-subresource-integrity)
 
 **plugin.js**
+#### Version 3
+```js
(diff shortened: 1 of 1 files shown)


hg-pyun:
@jantimon Would you please review? :)

jantimon:
Sorry I tried to release 4.x earlier but I couldn't find time .. I am still working on it :)

However the beta version is very stable and I don't plan to add any further code changes

hg-pyun:
@jantimon Thank you for feedback. This is only guide for V3 and It seems to be added some information hooks. I think this will be very helpful for users who continue to use V3.

PR #1309, opened 2019-11-11
feat: Add support for minifying inline ES6 inside html templates

jantimon:
This pull request switches html-minifier to html-minfier-terser because of inactivity of https://github.com/kangax/html-minifier/issues/1040 for 3 month

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "webpack-recompilation-simulator": "3.0.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


rodrigograca31:
awesome to see this accepted!

a v4.0.0 of this plugin should be release. (nothing should break but since this is a big change I would bump to v4.0.0)

want me to send the changes for a v4 release? (code and git tags)

PR #1321, opened 2019-12-24
html-webpack-plugin-for-multihtml

loongMa:
html-webpack-plugin-for-multihtml can add support for webpack4 ?


jantimon:
The speed optimisations are already part of the 4.x branch

PR #1324, opened 2019-12-28
Bug Fix: cannot emit output file when use chunkhash with unique url code

ShuZhong:
output: {
  path: path.join(__dirname, 'dist/webpack-' + webpackMajorVersion),
  publicPath: '',
  filename: 'bundle.js?v=[chunkhash]'
}

encodeURLComponent break 'bundle.js?v=[chunkhash]'


jantimon:
You should not use a question mark for your output name.

Try filename: 'bundle.[chunkhash].js' instead

PR #1364, opened 2020-03-23
Update typedefs after switch to html-minifier-terser

ianberinger:
Just upgraded to 4.0.0. I had to patch the type definitions to make `tsc` happy. Seems like this got lost during the switch to `html-minifier-terser` in  c66766cdae3593091dee413b9c585359c24ef068.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     ]
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


jantimon:
Thanks 👍

jantimon:
Released as html-webpack-plugin 4.0.1

PR #1386, opened 2020-04-01
fix: Fix querystring encoding

### LINKED ISSUE #1355: js file is not injected, but css file is injected to the template with 4.0.0-beta.11 version.
```js
new HtmlWebpackPlugin({
        filename: `${filePath + chunkName}.html`,
        template: `${ item }/${ chunkName }.html`,
        chunks: ['common'],
        cache: true,
        hash:true,
        inject: true
   });
```

jantimon:
fix #1355

```
fo:demonstration-path/very fancy+name.js?path=/home?value=abc&value=def#zzz
   ^            ^    ^    ^     ^    ^  ^    ^^    ^     ^   ^     ^   ^
   |            |    |    |     |    |  |    ||    |     |   |     |   |
encoded         |    |    encoded    |  |    ||    |     |   |     |   |
               ignored              ignored  ignored     ignored   ignored              
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -939,10 +939,35 @@ class HtmlWebpackPlugin {
    * Encode each path component using `encodeURIComponent` as files can contain characters
    * which needs special encoding in URLs like `+ `.
    *
+   * Valid filesystem characters which need to be encoded for urls:
+   *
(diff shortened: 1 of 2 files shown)


jantimon:
Probably this implementation is wrong and we should always start from the first question mark.

PR #1390, opened 2020-04-02
webpack 5 compatibiltiy

jantimon:
Thanks to @PutziSan and @fivethreeo we have now a first working webpack 5 version!!


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,21 @@
+name: Node.js CI
```
jantimon, line 1: can we please keep this in travis and migrate from travis to github actions in a later pr?

.github/workflows/npm_release.yml
```diff
@@ -0,0 +1,47 @@
+name: Do release
```
jantimon, line 1: can we please discuss about a github action release feature in another pull request and remove it for now?

lib/compiler.js
```diff
@@ -288,63 +289,63 @@ function getFileDependencies (compiler) {
 }
 
 /**
- * @type {WeakMap<WebpackCompilation, WeakMap<HtmlWebpackChildCompiler, boolean>>}}
+ * @type {WeakMap<object, Snapshot>}
  */
-const hasOutdatedCompilationDependenciesMap = new WeakMap();
+const snapshotMap = new WeakMap();
+const fileDepependenciesKeyRef = {};
```
jantimon, line 295: This looks really dangerous as it is one single global variable used for (potentially) multiple compilers, compilations and html-webpack-plugin instances.

lib/compiler.js
```diff
@@ -288,63 +289,63 @@ function getFileDependencies (compiler) {
...
-  if (!hasOutdatedChildCompilerDependenciesMap) {
-    hasOutdatedChildCompilerDependenciesMap = new WeakMap();
-    hasOutdatedCompilationDependenciesMap.set(mainCompilation, hasOutdatedChildCompilerDependenciesMap);
+
+  const keyStr = JSON.stringify([...childCompiler.fileDependencies].sort());
+  // get reference for key-strings to use it with WeakMap:
+  let key = fileDepependenciesKeyRef[keyStr];
+  if (!key) {
+    key = {};
+    fileDepependenciesKeyRef[keyStr] = key;
```
jantimon, line 313: This is a memory leak as it will grow on every compilation if output name contains a hash.

lib/compiler.js
```diff
@@ -288,63 +289,63 @@ function getFileDependencies (compiler) {
...
-  /**
-   * @type {WeakMap<HtmlWebpackChildCompiler, boolean>|undefined}
-   */
-  let hasOutdatedChildCompilerDependenciesMap = hasOutdatedCompilationDependenciesMap.get(mainCompilation);
-  // Create map for childCompiler if none exist
-  if (!hasOutdatedChildCompilerDependenciesMap) {
-    hasOutdatedChildCompilerDependenciesMap = new WeakMap();
-    hasOutdatedCompilationDependenciesMap.set(mainCompilation, hasOutdatedChildCompilerDependenciesMap);
+
+  const keyStr = JSON.stringify([...childCompiler.fileDependencies].sort());
```
jantimon, line 308: This is probably way slower than comparing the file dependencies

spec/caching.spec.js
```diff
@@ -128,6 +139,7 @@ describe('HtmlWebpackPluginCaching', () => {
       .then(stats => {
         // Expect no errors:
         expectNoErrors(stats);
+        console.log(getCompiledModuless(stats.toJson()));
```
jantimon, line 142: was that added for debugging only?

spec/hot.spec.js
```diff
@@ -81,8 +85,12 @@ describe('HtmlWebpackPluginHMR', () => {
         const hotUpdateJsFileNames = Object.keys(stats.compilation.assets).filter((fileName) => /\.hot-update\.js$/.test(fileName));
         expect(hotUpdateJsFileNames).not.toEqual([]);
         expect(hotUpdateJsFileNames.length).toEqual(1);
-        const hotUpdateFileSource = stats.compilation.assets[hotUpdateJsFileNames[0]].source();
-        expect(hotUpdateFileSource).not.toEqual('');
+        if (webpackMajorVersion < 5) {
```
jantimon, line 88: Why do we need this check?

What about removing the old `hotUpdateFileSource` code as you added tests based on size anyway?

lib/compiler.js
```diff
@@ -288,63 +289,63 @@ function getFileDependencies (compiler) {
...
-}
 
-/**
- * Returns `true` if the file dependencies of the given childCompiler are outdated.
- *
- * @param {WebpackCompilation} mainCompilation
- * @param {HtmlWebpackChildCompiler} childCompiler
- * @returns {boolean}
- */
-function isChildCompilerCacheOutdated (mainCompilation, childCompiler) {
```
jantimon, line 333: why did you remove this part?

spec/caching.spec.js
fivethreeo: Yes :)

fivethreeo:
use the webpack5-dev branch it is without actions

spec/hot.spec.js
fivethreeo: True, I'll take it out :)

lib/compiler.js
fivethreeo: Comes from PutziSan's changes, I'll need to take a look.

fivethreeo: It is because the checking went from being sync to async

lib/compiler.js
fivethreeo: Use a weakmap? or a weakmap on compiler?

package.json
```diff
@@ -42,12 +43,12 @@
     "pug": "2.0.3",
     "pug-loader": "2.4.0",
     "rimraf": "2.6.3",
-    "semistandard": "13.0.1",
+    "semistandard": "^13.0.1",
```
jantimon, line 46: We should revert his part - otherwise unit tests might break randomly

fivethreeo: Oh, that snuck in. Will fix.

fivethreeo:
But I need a decision from you. 
Typescript fails because of missing /wrong types if the wrong webpack in installed in dev.

There are four options. 

1. Ignore types in compiler-webpack5 and compiler-webpack4.
2. Ignore types selectively by modifying tsconfig.json based on the webpack version found in pretest.
3. Release two versions of html-webpack-plugin, either by scoping it or doing a 5.x series, 4.x for webpack4. And drop webpack4 in 5.x.
4. Spend a long time creating custom types supporting both webpack versions or figure out how to do this some other way in typescript.

1 and 2 are not optimal solutions and will probably cause issues down the line. 4 needs a typescript expert which I’m not :(. But rolling releases I can do. :)

fivethreeo:
Close this and use #1392

lib/compiler.js
fivethreeo: Oh, I saw wrong. isChildCompilerCacheOutdated needs to go back in.

PR #1392, opened 2020-04-03
 webpack 5 compatibiltiy, without github actions

fivethreeo:


jantimon:
I prepared some changes in order to fix travis again can you please tick this checkbox?

image

jantimon:
If you take a look at the current implementation:

https://github.com/jantimon/html-webpack-plugin/blob/c5a58824b8d33c3f9b85d1f541a27844a62721e4/index.js#L136-L141

This code is executed multiple times:
 - ones for every html-webpack-plugin instance
 - ones for every webpack compilation

The following line is executed synchronously and needs to know if the cache is outdated or not:
```
childCompiler.addTemplateToCompiler(compiler, this.options.template);
```

However this information is only retrieved at a later time.

If multiple html-webpack-plugins are used it gets even worse - the cache is cleared multiple times and removes templates from the compiler.

fivethreeo:
Was already checked :) Yes, hm, need a way to make it sync it seems.

spec/hot.spec.js
```diff
@@ -11,6 +11,10 @@ const webpack = require('webpack');
 const rimraf = require('rimraf');
 const WebpackRecompilationSimulator = require('webpack-recompilation-simulator');
 const HtmlWebpackPlugin = require('../index.js');
+const webpackMajorVersion = Number(require('webpack/package.json').version.split('.')[0]);
+if (isNaN(webpackMajorVersion)) {
+  throw new Error('Cannot parse webpack major version');
+}
```
jantimon, line 17: You could use `assert` here https://nodejs.org/api/assert.html#assert_assert_value_message

```suggestion
asset (!isNaN(webpackMajorVersion), 'Cannot parse webpack major version');
```

lib/compiler-webpack5.js
```diff
@@ -0,0 +1,374 @@
...
+  const fileTimestamps = snapshot.fileTimestamps;
+  const isCacheOutOfDate = childCompiler.fileDependencies.some((fileDependency) => {
+    console.log(fileDependency);
+    const timestamp = fileTimestamps.get(fileDependency).timestamp;
+    // If the timestamp is not known the file is new
+    // If the timestamp is larger then the file has changed
+    // Otherwise the file is still the same
+    return !timestamp || timestamp > childCompiler.compilationStartedTimestamp;
+  });
+  return isCacheOutOfDate;
```
jantimon, line 365: Wow you really made it to be sync!

That's amazing :)

I'll have to look into this but this would solve it really nicely

fivethreeo: Hm, I'm not so sure, I think the compilation starts before this check is done.

timestamp > childCompiler.compilationStartedTimestamp is false :(

jantimon:
I talked to the core team and they told me it is not possible to get an uptodate timestamp with the snapshot api:

```
The process of these snapshotting is the following:
1. store startTime
2. make something with files on disk, keep track of each file read
3. create a snapshot of all files read + startTime
4. wait until next compilation
5. compare snapshot with files on disk, startTime is there to handle race conditions, e. g. change between 1. and 3.
6. if it matches, reused cache value.
7. if it doesn't match go to 1.
```

jantimon:
I tried to prepare a cache manager which is able to handle async compilations in a better way.

jantimon:
It is 80% done and would need to be integrated with your solution once the tests are green again.

jantimon:
I will try to get webpack 4 running with the async supported cache validation tonight (only 4 unit tests left) - once that is done we can try to merge both branches together

Issue #1383, opened 2020-03-31
4.0.3: TS2322: Type 'HtmlWebpackPlugin' is not assignable to type 'Plugin'


stilettk:
## Expected behaviour

Build should succeed like in v4.0.2.

## Current behaviour

```
TSError: ⨯ Unable to compile TypeScript:
webpack.config.ts:59:3 - error TS2322: Type 'HtmlWebpackPlugin' is not assignable to type 'Plugin'.
  Types of property 'apply' are incompatible.
    Type '(compiler: import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/html-webpack-plugin/node_modules/@types/webpack/index").Compiler) => void' is not assignable to type '(compiler: import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/@types/webpack/index").Compiler) => void'.
      Types of parameters 'compiler' and 'compiler' are incompatible.
        Type 'import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/@types/webpack/index").Compiler' is not assignable to type 'import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/html-webpack-plugin/node_modules/@types/webpack/index").Compiler'.
          The types of 'hooks.shouldEmit.call' are incompatible between these types.
            Type '(arg1?: import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/@types/webpack/index").compilation.Compilation | undefined, arg2?: any, arg3?: any, ...args: any[]) => any' is not assignable to type '(arg1?: import("C:/Users/n.batov/RiderProjects/restaurantcashier-frontend/node_modules/html-webpack-plugin/node_modules/@types/webpack/index").compilation.Compilation | undefined, arg2?: any, arg3?: any, ...args: any[]) => any'.
              Types of parameters 'arg1' and 'arg1' are incompatible.
[...]

PR #1402, opened 2020-04-08
fix: Allow to use outdated @types/webpack types

jantimon:
If an older version of @types/webpack is installed in the same project typescript reported "TS2322:
Type 'HtmlWebpackPlugin' is not assignable to type 'Plugin'"

#1383

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "@types/loader-utils": "1.1.3",
+    "@types/webpack": "~4.41.8",
     "@types/node": "11.13.9",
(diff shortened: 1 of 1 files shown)


kamilic:
hi. what time will this changes merged?

kamilic:
hi. what time will this changes merged?

jantimon:
Looks like this might cause even more problems: https://github.com/jantimon/html-webpack-plugin/issues/1383#issuecomment-616838203

jantimon:
The type support of `webpack@5` has improved a lot 👍 
For `html-webpack-plugin@5` we will drop `@types/webpack`

PR #1403, opened 2020-04-10
docs: fix typo on description of `templateContent` options

latipun7:
This patch fix a typo in README.md where the table of option not escaped properly in `templateContent` option.

templateContent


jantimon:
thanks 👍

PR #1418, opened 2020-04-28
fix: prevent scripts marked as hotModuleReplacement from being added to the html file

jantimon:
this fixes a bug for hot module replacement:

https://github.com/webpack/webpack/issues/10790#issuecomment-620647237

thanks @sokra


jordanbtucker:
Why was this implemented? This has broken my `npm run dev` script because it deletes my `index.html` file every time I make a change to a `.js` file.

PR #1419, opened 2020-04-28
Update templateParameters example to account for object merging

### LINKED ISSUE #1416: Since #1331 merged, templateParameters example is out-of-date
Since #1331 was merged, `templateParameters` will merge a given object with the default parameters. In the [example ](https://github.com/jantimon/html-webpack-plugin/tree/master/examples/template-parameters) for this option, [the comment about defaults not being merged](https://github.com/jantimon/html-webpack-plugin/blob/96e0ae607afa7295c40d0a5073c791189def6337/examples/template-parameters/webpack.config.js#L14) is no longer correct.

I'm not sure if the comment should be updated to reflect this new feature or the comment simply removed and the example rewritten as follows:

```javascript
plugins: [
  new HtmlWebpackPlugin({
    templateParameters: {
      'foo': 'bar'
    },
    template: 'index.ejs'
  })
]
```

genderquery:
PR #1331 introduces new behavior that merges a passed object with the defaults. Fixes #1416 where the templateParameters example is out-of-date.

### DIFF
--- a/examples/template-parameters/webpack.config.js
+++ b/examples/template-parameters/webpack.config.js
@@ -11,20 +11,24 @@ module.exports = {
   },
   plugins: [
     new HtmlWebpackPlugin({
-      // NOTE if you pass plain object it will be passed as is. no default values there, so be aware!
-      // for implementation detail, please see index.js and search for "userOptions" variable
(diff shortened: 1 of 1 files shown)


jantimon:
Thanks 👍

PR #1420, opened 2020-04-28
Update index.js

Vladi4-gh:
Icon tag rel attribute has been amended with modern value.


jantimon:
Cool ty I'll try to fix the tests tomorrow if you can't find out why they are broken

PR #1439, opened 2020-05-16
Add Open Graph tags support

mdesilva:


jantimon:
Isn't this already possible with meta tags?

jantimon:
```js
new HtmlWebpackPlugin({
  // <meta property="og:title" content="The Rock" />
  meta: { "og:title": { property: "og:title", content: "The Rock" } }
})
```

mdesilva:
Yes, it is possible with the meta tag, but the user must nest the tag keys within a redundant key, and must also be familiar with the meta tag keys (property and content). This is error prone and not very user friendly. With these new tags, one only needs to create a single key value pair for each tag they want to generate. Easy.

jantimon:
I understand your point but I would prefer not to to have two properties for meta tags in the options.

mdesilva:
Is there a way that we can put this to a vote/get more feedback on this from additional committers ?

jantimon:
Open Graph was part of the favicons-webpack-plugin but has been removed: https://github.com/itgalaxy/favicons/commit/7f59a2e20c1ed95605df07f496f484c4368de72e

A quick fix might be the following:

```js
const openGraphTags = (tags) => Object.fromEntries(Object.entries(tags).map(([k,v]) => ['og:' + k, {['og:' +k]: v}]));

new HtmlWebpackPlugin({
  // <meta property="og:title" content="The Rock" />
  meta: { 
    ...openGraphTags({ title: 'The Rock' })
  }
})
```

Or the full syntax: 

```
new HtmlWebpackPlugin({
  // <meta property="og:title" content="The Rock" />
  meta: { "og:title": { property: "og:title", content: "The Rock" } }
})
```

mdesilva:
Sorry, what does a different npm package have to do with this repo ?

Issue #1408, opened 2020-04-20
Deprecation Warning with MainTemplate after webpack 5 update


bartdominiak:
## Expected behaviour

Clean output without deprecation warning

## Current behaviour

After webpack update (from v4 to v5) I'm getting deprecation warnings in my console 

```
node --trace-deprecation node_modules/.bin/webpack-cli --progress --config ./config --env dev
```

Output:
```
10% building
webpack is watching the files…

(node:78415) [DEP_WEBPACK_MAIN_TEMPLATE_GET_ASSET_PATH] DeprecationWarning: MainTemplate.getAssetPath is deprecated (use Compilation.getAssetPath instead)
    at /Users/bartek/Sites/private/vue-starter/node_modules/html-webpack-plugin/lib/child-compiler.js:163:41
    at Array.map (<anonymous>)
    at extractHelperFilesFromCompilation (/Users/bartek/Sites/private/vue-starter/node_modules/html-webpack-plugin/lib/child-compiler.js:162:45)
    at /Users/bartek/Sites/private/vue-starter/node_modules/html-webpack-plugin/lib/child-compiler.js:113:13
    at /Users/bartek/Sites/private/vue-starter/node_modules/webpack/lib/Compiler.js:450:11
    at /Users/bartek/Sites/private/vue-starter/node_modules/webpack/lib/Compiler.js:922:16
    at Hook.eval [as callAsync] (eval at create (/Users/bartek/Sites/private/vue-starter/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:4:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (/Users/bartek/Sites/private/vue-starter/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
    at /Users/bartek/Sites/private/vue-starter/node_modules/webpack/lib/Compiler.js:918:32
    at Hook.eval [as callAsync] (eval at create (/Users/bartek/Sites/private/vue-starter/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:4:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (/Users/bartek/Sites/private/vue-starter/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
    at /Users/bartek/Sites/private/vue-starter/node_modules/webpack/lib/Compilation.js:1689:41
[...]

PR #1454, opened 2020-06-01
added v5 compilation support and deleted depreciation warnings

bartdominiak:
#1408

What I did:
- Deleted depreciation warnings in v5 (`mainTemplate.getAssetPath`, and `getPublicPath`)
- Added backward compatibility with v4

I also consider to put this `webpackMajorVersion`, as a helper somewhere - is quite repeated few times. Feedback will be needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,8 @@ const getHtmlWebpackPluginHooks = require('./lib/hooks.js').getHtmlWebpackPlugin
 const fsStatAsync = promisify(fs.stat);
 const fsReadFileAsync = promisify(fs.readFile);
 
+const webpackMajorVersion = Number(require('webpack/package.json').version.split('.')[0]);
+
 class HtmlWebpackPlugin {
   /**
    * @param {HtmlWebpackOptions} [options]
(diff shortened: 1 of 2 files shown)


bartdominiak:
It's looks like there is an error on `Compilation` array length, which is not related to this PR, and it occur on master branch also.

index.js
```diff
@@ -149,12 +151,16 @@ class HtmlWebpackPlugin {
...
-        const childCompilationOutputName = compilation.mainTemplate.getAssetPath(this.options.filename, 'compiledEntry' in templateResult ? {
+        const compiledEntries = 'compiledEntry' in templateResult ? {
           hash: templateResult.compiledEntry.hash,
           chunk: templateResult.compiledEntry.entry
         } : {
           hash: templateResult.mainCompilationHash
-        });
+        };
+
+        const childCompilationOutputName = webpackMajorVersion === 5
```
artola, line 161: I would invert the logic to support future versions too (and as it is done in other parts of the codebase)

```
= webpackMajorVersion === 4 ? ... : ...
```

bartdominiak: That was my first thought too, but we'll lose backward compatibility with v3 and lower, and I'm not 100% that we should support them also? And also this `Compilation` syntax is different only in webpack v5.

artola: > That was my first thought too, but we'll lose backward compatibility with v3 and lower, and I'm not 100% that we should support them also? And also this `Compilation` syntax is different only in webpack v5.

Compatibility is stated by:

```
  "peerDependencies": {
    "webpack": ">=4.0.0 < 6.0.0"
  },
```

https://github.com/jantimon/html-webpack-plugin/blob/115bd8a5f8641dc2bd58dd5efefe360e3cdc0102/package.json#L64

bartdominiak: I've one more idea to pick major `v5` and higher like:
```
webpackMajorVersion >= 5 ? ... : ...
```
What do you think?

artola: I was more looking to how it was written in `file-watcher-api.js`.

https://github.com/jantimon/html-webpack-plugin/blob/115bd8a5f8641dc2bd58dd5efefe360e3cdc0102/lib/file-watcher-api.js#L11

pi0:
Hi @jantimon can you please check this PR? :pray:

jantimon:
Released as 4.4.0 - sorry for the delay

PR #1491, opened 2020-08-04
Update index.js - added "/" to path inside generateStyleTags function

Pashkinz92:
I had problems with css styles. I use laravel mix html template and always when pages are generated I see wrong link to css. about js/images - all fine. For example: I expect to see "/static/<filename>.css" but see "static/<filename>.css" As for js I see "/static/<filename>.js"

"static" is juts my path. Problem only with one symbol

### DIFF
--- a/index.js
+++ b/index.js
@@ -705,7 +705,7 @@ class HtmlWebpackPlugin {
       tagName: 'link',
       voidTag: true,
       attributes: {
-        href: styleAsset,
+        href: "/" + styleAsset,
         rel: 'stylesheet'
       }
     }));


jantimon:
the html-webpack-plugin is using the webpack output public path setting:

https://webpack.js.org/configuration/output/#outputpublicpath

you will probably have to set it to `/static` or to `/`

Pashkinz92:
I do not agree. because with the same configuration I get the right route to js with the correct first symbol. But only for CSS route plugin misses the first symbol

PR #1503, opened 2020-08-30
fix: broken typings.d.ts in v4.4.0

AviVahl:
class declaration is outside the namespace, so it needs to use it as a prefix to access its types.


AviVahl:
heya @jantimon, any chance for a quick 4.4.1? this issue broke type checking in typescript projects which used `html-webpack-plugin` programmatic api.

jantimon:
Sure thanks for fixing it 👍

jantimon:
Released as 4.4.1

PR #1505, opened 2020-09-01
update webpack dependency range fix #1504

### LINKED ISSUE #1504: Webpack 5 gives peer dependency warning
## Expected behaviour

Plugin should allow webpack 5 pre-releases peer dependency without warning

## Current behaviour

> npm install

```
npm WARN html-webpack-plugin@4.4.1 requires a peer of webpack@>=4.0.0 < 6.0.0 but none is installed. You must install peer dependencies yourself.
```

## Environment

```
Node.js v12.18.2
win32 10.0.18362
npm 6.14.5
webpack@5.0.0-beta.29
html-webpack-plugin@4.4.1
```

### Notes

I guess 5.0.0-beta.29 is indeed less than 6.0.0 but I wonder if the embedded spaces in your `peerDependencies` [range specification](https://devhints.io/semver) is correct.  Maybe we want `webpack@>=4.0.0 <6.0.0`

GaryB432:
I used the same range specification as [css-loader](https://github.com/webpack-contrib/css-loader/blob/master/package.json#L43)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "util.promisify": "1.0.0"
   },
   "peerDependencies": {
-    "webpack": ">=4.0.0 < 6.0.0"
+    "webpack": "^4.27.0 || ^5.0.0"
   },
   "keywords": [
     "webpack",
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "util.promisify": "1.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


jantimon:
We can't support 4.0.0 as we rely on features which were created in 4.27.0

GaryB432:
Well I don't know why the peer dependency update broke your test 🤷‍♂️

PR #1513, opened 2020-09-20
Add preserveStrings option

ntilwalli:
This PR enables `html-webpack-plugin` to be used in projects that need to emit files in simple html-like templating languages like Elixir/Phoenix's EEx templating language.  When the `preserveStrings` option is enabled, the user can provide one or an array of `preserve` regular expressions which will trigger a pre-compilation substitution with a filler string.  Just before emission, those string a replaced with the original string, thus allowing the default processor to treat the template like a normal HTML template but enabling output in the desired templating language.  This PR was motivated by the EEx templating syntax which is almost exactly similar to the syntax for Lodash templates and was causing template substitution errors when applied with `html-webpack-plugin`'s default behavior. I use this functionality by providing an `eex` template in the `template` property and giving the desired `eex` directory/filename I'd like to be output in the `filename` property.  

Any suggestions on a better approach or guidance on improvements would be appreciated.


jantimon:
Couldn't you just use a different loader?

e.g. the webpack raw-loader?

```js
new HtmlWebpackPlugin({
  template: '!!raw-loader!./my-file.html'
})
```

ntilwalli:
Wow, yes!  I swear I tried that, and got some error.  Must have been something else.  Learning experience of sorts, I guess.  Thanks!

Issue #1514, opened 2020-09-21
[webpack 5] support `publicPath: "auto"`


sokra:
## Expected behaviour

When using the default of `publicPath: "auto"`, html-webpack-plugin should generate a script tag with an URL pointing to the script file.

## Current behaviour

``` js
<script src="auto/main.js"></script>
```

## Environment

webpack 5.0.0-rc.0
html-webpack-plugin: 4.4.1

### Config

Copy the minimal `webpack.config.js` to produce this issue:

```js
module.exports = {
  plugins: [
    new HtmlWebpackPlugin()
  ]
}
```

## Additional context

webpack has added `output.publicPath: "auto"`. This figures out the correct publicPath at runtime by reading it from the script tag.

html-webpack-plugin should understand this magic value and behave like `publicPath: ""`

jantimon:
Is there an API to get the public path?

sokra:
fixed by https://github.com/jantimon/html-webpack-plugin/pull/1516

FishOrBear:
@sokra 
When using the `DllReferencePlugin` plugin, there is still a problem of `auto`.

"auto/xxx.dll.js"

PR #1515, opened 2020-09-21
fix: update lodash

jantimon:


jantimon:
Unfortunately this is blocked by #1514 (which is breaking the test pipeline right now)

PR #1530, opened 2020-10-16
Update Hook Types

mbostwick:
This typing on the alpha branch is not correct. This PR fixes the types so they compile .


jantimon:
Thanks 👍 merged into next branch

PR #1536, opened 2020-10-21
docs: add html-webpack-inject-style-plugin to plugin list

niexias:
hi, can you take a look at this mr? this plugin can help users solve the problem of adapting RTL language.

webpack-rtl-plugin is plugin to use to create a second css bundle, processed to be rtl, rtl css bundle default filename is xxx.rtl.css. When html-webpack-plugin will generate an HTML5 file, they are all included with tags in the HTML head by default.

It is expected that the corresponding styles can be loaded according to the language type, instead of all styles being injected into index.html by default.This plugin is used to help us complete this work. It uses cookies to determine the language type, and then decides which style to load.


niexias:
hi,  can you take a look at this mr? this plugin can help users solve the problem of adapting RTL language.

niexias:
@

Issue #1527, opened 2020-10-13
webpack 5 <-> html-webpack-plugin 5


jantimon:
Finally [Webpack 5 has been released](https://webpack.js.org/blog/2020-10-10-webpack-5-release/)! 👍 

During the beta I tried to provide a html-webpack-plugin version which is compatible to webpack 4 **and** webpack 5.  
As there have been some API and typing changes in webpack 5 this approach is limited.

To provide the best possible webpack 5 support I have started on a dedicated webpack 5 version of the html-webpack-plugin.

## current progress:

- [x] drop webpack 4 support
- [x] drop node < 10 [like webpack](https://github.com/webpack/webpack/blob/da67529a213ccb0e687d5f95144faad7bdf26a53/package.json#L106-L108)
- [x] make use of Compiler.hook.initialize to get the latest options
- [x] update typings for webpack 5
- [x] fix scriptLoading defer for inject false
- [x] make scriptLoading defer the default
- [x] add assets in Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL (see https://github.com/webpack/webpack/issues/11425#issuecomment-686606318 )
- [x] add assets with new api `compilation.emitAsset` (#1522)
- [x] use `new webpack.sources.RawSource(html)` to create assets
- [x] add support for `[name]`
- [x] provide options and output name in all hooks
- [x] allow filename to be a function
- [x] emitAssets even if they have been generated in a previous compilation (to fix #1476 and #1432)
- [x] use [`compiler.webpack.LibraryTemplatePlugin`](https://github.com/webpack/webpack/issues/11909)
- [x] verify support for raw-loader (blocked by https://github.com/webpack/webpack/issues/11909)
- [x] remove `@types/tapable` as it is included in tapable 2
- [x] provide a publicPath to the child compiler to add support for `publicPath: 'auto'` & `'asset/resource'` (https://github.com/webpack/webpack/issues/11968#issuecomment-725117719)
- [x] use the correct stage to allow sri and service worker generation (blocked by https://github.com/webpack/webpack/issues/11822)
- [x] fix error reporting (which is currently breaking the travis build)
- [x] use the html-webpack-plugin `publicPath` option also for the template child compilation

- ~~remove html-minification (if a separate plugin which covers the html minification is available)~~ (postponed)

## current blockers (help needed by the webpack team):

- [x] [webpack/issues/11822](https://github.com/webpack/webpack/issues/11822)
[...]

PR #1537, opened 2020-10-24
do not convert favicon to string

akphi:
@jantimon As per mentioned in https://github.com/jantimon/html-webpack-plugin/issues/1527


jantimon:
Thanks i'll merge it tonight or tomorrow :)

PR #1538, opened 2020-10-25
Fix/empty public path

jantimon:
> The Webpack documentation for [Ouput PublicPath](https://webpack.js.org/configuration/output/#outputpublicpath) lists these examples
> 
> ```js
> module.exports = {
>   //...
>   output: {
>     // One of the below
>     publicPath: 'https://cdn.example.com/assets/', // CDN (always HTTPS)
>     publicPath: '//cdn.example.com/assets/', // CDN (same protocol)
>     publicPath: '/assets/', // server-relative
>     publicPath: 'assets/', // relative to HTML page
>     publicPath: '../assets/', // relative to HTML page
>     publicPath: '', // relative to HTML page (same directory)
>   }
> };
> ```

However right now `''` is not supported by the html-webpack-plugin

With Webpack 5 we can make use of the new default for publicPath (`'auto'`) and stop handling the empty string edge case


jantimon:
This branch was merged into 'next'

PR #1539, opened 2020-10-27
remove deprecated plugins from readme

numical:
Hi, I can no longer support two of the plugins listed in your README so they should be removed.
If someone chooses to fork and maintain these going forward, then they can ask for them to be restoted here.
Thanks for the platform!
Sorry I can no longer be part of it.


jantimon:
thanks for the heads up

Issue #1535, opened 2020-10-20
ERROR in TypeError: The 'compilation' argument must be an instance of Compilation


tw19920521:
ERROR in TypeError: The 'compilation' argument must be an instance of Compilation

  - JavascriptModulesPlugin.js:118 getCompilationHooks
    [webpack_demo]/[webpack]/lib/javascript/JavascriptModulesPlugin.js:118:10

  - CommonJsChunkFormatPlugin.js:30 compiler.hooks.thisCompilation.tap.compilation
    [webpack_demo]/[webpack]/lib/javascript/CommonJsChunkFormatPlugin.js:30:19

  - Hook.js:14 Hook.CALL_DELEGATE [as _call]
    [npm]/[webpack]/[tapable]/lib/Hook.js:14:14

  - Compiler.js:915 Compiler.newCompilation
    [npm]/[webpack]/lib/Compiler.js:915:30

  - Compiler.js:957 hooks.beforeCompile.callAsync.err
    [npm]/[webpack]/lib/Compiler.js:957:29

  - Hook.js:18 Hook.CALL_ASYNC_DELEGATE [as _callAsync]
    [npm]/[webpack]/[tapable]/lib/Hook.js:18:14

  - Compiler.js:952 Compiler.compile
    [npm]/[webpack]/lib/Compiler.js:952:28

  - Compiler.js:477 Compiler.runAsChild
    [npm]/[webpack]/lib/Compiler.js:477:8

webpack 5.0.0 compiled with 1 error in 1586 ms

tw19920521:
so? How to solve

tw19920521:
const path = require('path')
const HtmlWebpackPlugin = require('html-webpack-plugin')
plugins: [
	new HtmlWebpackPlugin({
		template: path.resolve(__dirname, 'index.html'),
	}),
],

csvan:
@tw19920521 this plugin is not fully compatible with Webpack 5 yet, see https://github.com/jantimon/html-webpack-plugin/issues/1527

ChrisAckerman:
See also this issue with discussion: https://github.com/jantimon/html-webpack-plugin/issues/1451

jantimon:
duplicate of #1451

jantimon:
Probably fixed in `html-webpack-plugin@5.0.0-alpha.8` please repoen this issue if it's not fixed.

PR #1544, opened 2020-10-31
feat: add an error if multiple incompatible webpack instance have been found

jantimon:
This pr tries to warn the user early about https://github.com/jantimon/html-webpack-plugin/issues/1535

Often this error is caused by mono repositories or by broken peer dependencies.

Some third party extensions for webpack like `cache-loader` or `webpack-dev-server` are locked to `webpack@4` peer dependency (see https://github.com/webpack/webpack-dev-server/issues/2807)


jantimon:
Thanks to a hint from @sokra we will use `compiler.webpack` instead of `require('webpack')` - this should fix all issues caused by multiple webpack installations

PR #1549, opened 2020-11-02
Add note that examples relies on ejs

jlsjonas:
This might not be obvious to people who haven't used ejs yet

### DIFF
--- a/examples/custom-insertion-position/readme.md
+++ b/examples/custom-insertion-position/readme.md
@@ -1,5 +1,7 @@
 # custom insertion example
 
+**Note**: this example uses [ejs](https://ejs.co/), please substitute `<%=` & `%>` for your engine's (non-escaping) tags
+
 This example shows how you can define the position where the scripts are injected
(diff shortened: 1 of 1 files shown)


jantimon:
I am really sorry - The hardest part in writing documentation is to know which parts are relevant...

I picked the ejs templates from underscore/lodash because it used to be the most common template language back then..

Unfortunately I don't understand your note 😄

The template option in the readme links to this explanation:

https://github.com/jantimon/html-webpack-plugin/blob/master/docs/template-option.md

jlsjonas:
We came there from 
> see the [inject:false example](https://github.com/jantimon/html-webpack-plugin/tree/master/examples/custom-insertion-position)

Not realizing the example used ejs syntax.
For some reason, we DID get a replacement in development, but then not when actually building for production, causing a lot of investigating before stumbling upon the simple fact that the wrong template language was being used ( _assisted_ by the fact that in-dev it did seem to parse it)

PR #1550, opened 2020-11-03
docs: add html-webpack-inject-externals-plugin to to plugin list

superwf:


jantimon:
Cool :)  
What's the idea behind injecting those tags?

superwf:
Thank you.
I like use as much externals as possible to reduce webpack output size. In my work, manually sync the version part in script src when upgrade npm dependencies is an offen forgotten job. So i write this to auto do that work.
And in the future, the `<script type="module">" import method with the platform `unpkg.com` will grow more useable.

PR #1553, opened 2020-11-06
Fix ProcessedOptions.filename type

### LINKED ISSUE #1552: Interface 'Options' incorrectly extends interface 'Partial<ProcessedOptions>'
Hi, with `5.0.0-alpha.10` and typescript 4.0.5 I'm getting this:

```
node_modules/html-webpack-plugin/typings.d.ts:36:13 - error TS2430: Interface 'Options' incorrectly extends interface 'Partial<ProcessedOptions>'.
  Types of property 'filename' are incompatible.
    Type 'string | ((entryName: string) => string)' is not assignable to type 'string'.
      Type '(entryName: string) => string' is not assignable to type 'string'.

36   interface Options extends Partial<ProcessedOptions> {
               ~~~~~~~

```

jscheid:
Closes #1552

### DIFF
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -74,7 +74,7 @@ declare namespace HtmlWebpackPlugin {
      *
      * @default 'index.html'
      */
-    filename: string
+    filename?: string | ((entryName: string) => string);
     /**
      * By default the public path is set to `auto` - that way the html-webpack-plugin will try
      * to set the publicPath according to the current filename and the webpack publicPath setting


jantimon:
After processing it is only a `string`

jantimon:
What about:

```
 type Options = Partial<ProcessedOptions> & {
    /**
     * The file to write the HTML to.
     * Supports subdirectories eg: `assets/admin.html`
     * [name] will be replaced by the entry name
     * Supports a function to generate the name
     *
     * @default 'index.html'
     */
    filename?: string | ((entryName: string) => string);
  }
```

jscheid:
> Maybe I should flip it the other way round (calculate ProcessedOptions out of Options)

That sounds like the way to go... I'll work off this branch for now and let you find a good long term solution.

jantimon:
Fixed in 5.0.0-alpha.12:

https://github.com/jantimon/html-webpack-plugin/blob/f15f8e5c695781f4f6a18a68b156da6a3dc356f3/typings.d.ts#L29-L162

jscheid:
Nice! 👍

PR #1575, opened 2020-12-17
Update Readme - no longer maintained plugin

donald-lendit:
## description

[html-webpack-inline-source-plugin](https://github.com/DustinJackson/html-webpack-inline-source-plugin) is outdated

스크린샷 2020-12-17 12 08 52


jantimon:
Thanks 👍

PR #1584, opened 2021-01-06
Align section Minification with the code regarding `keepClosingSlash`

### LINKED ISSUE #1583: Closing slash behavior doesn't match documentation
## Current behaviour 💣
The plugin minifies the production code while keeping closing slashes, which is different from the default behavior of `html-minifier-terser`.

## Expected behaviour ☀️
The closing slashes should be removed according to the documentation. While the documentation doesn't specifically mention `keepClosingSlash`, it mentions other minification overrides which makes the reader believe that this one is unaltered.

This is likely introduced by #1371, which modified the code without modifying the documentation. While I think that the current behavior is reasonable, it took some debugging to understand why it didn't match my expectations. Hence, perhaps the documentation should be updated?

sjoqvist:
Fixes #1583

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,7 @@ and the following options:
 ```js
 {
   collapseWhitespace: true,
+  keepClosingSlash: true,
   removeComments: true,
   removeRedundantAttributes: true,
   removeScriptTypeAttributes: true,


jantimon:
Thanks :)

PR #1587, opened 2021-01-12
Implement entries map between entries and their output dependencies

### LINKED ISSUE #1585: Expose asset dependencies to template
**Is your feature request related to a problem? Please describe.**

Maybe this is the wrong plugin but I have never found anything as reliable for generating asset manifests that can be loaded by WordPress/Drupal/Symfony projects. Usually just generating YML files that describe the assets that the project should be loading.

Over time I've started to split down a very large project's frontend code into many different bundles to load at relevant times, and split out common code into common bundles using code splitting. However, this creates a problem in that the Html-Webpack-Plugin just has `css` and `js` entries in `files` so everything is clumped together. There's no way to know that a specific `js` file is only used by a specific entry point. Ideally it would be great to generate the manifest to say that when code needs X entry point it needs to load files Y, Z, A and B (because Y is the main output, and the others dependencies containing code from code splitting). That way the project knows the exact set of files needed for each page and only needs to know which entry point corresponds to that page.

**Describe the solution you'd like**
It would be nice if we could also have a `entries` or something in the `files` variable that is essentially entry point as key and an array of required files as value. Then you could use a single Html-Webpack-Plugin to generate a YML output saying that here are my bundle outputs (based on entry point) and here are the list of files each bundle requires.

**Describe alternatives you've considered**
I've achieved a kind of result by using Html-Webpack-Plugin multiple times and filtering by each entry point chunk. Each generated file then describes each chunk and only it's dependencies so the project knows when it needs said bundle (chunk) it also needs the other commons entries. However it creates a lot of files and makes a mess of the web pack configuration, and effectively for our use case means another post-webpack step to combine these files because our projects expect a single manifest of assets.

**Additional context**
Using @next I can see this would only need a few extra lines for `htmlWebpackPluginAssets` like below. Would be great to hear if this is a complete mis-use of the plugin or if this is something that would be considered a useful feature.

```
      entryPointPublicPaths.forEach((entryPointPublicPath) => {
        const extMatch = extensionRegexp.exec(entryPointPublicPath);
        // Skip if the public path is not a .css, .mjs or .js file
        if (!extMatch) {
          return;
        }
        if (!assets.entries[entryName]) {
            assets.entries[entryName] = [];
        }
        assets.entries[entryName].push(entryPointPublicPath);
```

Would be great to hear if there are other tools too but I like html-webpack-plugin as it's pretty much 99% what we need and superbly maintained in the community <3. OK we're generating YML not HTML but it's all the same just generated code. We just use js-yaml in the template and output a JS object we build inside the template code. Adding this `entries` would give all information a template needs to generate all kinds of outputs - including directly generating a Drupal 8 libraries.yml or even a WordPress functions.php snippet that defines scripts and an exact set of dependencies for them. Happy to PR.

driskell:
Fixes #1585 

Allows better generation of JSON/YAML asset manifests, by exposing each entry and it's outputs (including split chunks).

I'm not 100% sure how it would impact hooks. I can see there's one hook - the `beforeAssetTagGeneration` one - that would process the `assets` object I added the `entries` to. I guess in most cases if the plugin either mutates or does a clone and mutation, things will be fine. I can see if a hook generated it's own it would then be missing - but I don't think this is a major issue as it could be said the set of plugins in use does not support the new feature yet.

Worth noting when I reviewed this area of code, the injection code refers to the outer scope's `assets` and therefore `assets.manifest` from that outer scope, instead of the `assets` object floating through the Promise that a plugin may have replaced if it treated it as immutable. I can raise another PR to tweak but I guess means carrying through the `assets` a few stages further in the promise waterfall, as it quickly drops off and only passes through `assetTags` prior to the injection promise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ The following variables are available in the template by default (you can extend
      </html>
      ```
   
-  - `htmlWebpackPlugin.files`: direct access to the files used during the compilation.
+  - `htmlWebpackPlugin.files`: direct access to the files used during the compilation. The `entries` property contains a mapping of entry names to the list of generated bundles required by that entry.
 
     ```typescript
     publicPath: string;
@@ -284,6 +284,7 @@ The following variables are available in the template by default (you can extend
     css: string[];
     manifest?: string;
     favicon?: string;
+    entries: { [entry: string]: Array<string> };
     ```
 
 
@@ -503,7 +504,8 @@ about which values are passed.
         js: Array<{string}>,
         css: Array<{string}>,
         favicon?: string | undefined,
-        manifest?: string | undefined
+        manifest?: string | undefined,
+        entries: { [entry: string]: Array<string> }
       },
       outputName: string,
       plugin: HtmlWebpackPlugin
--- a/index.js
+++ b/index.js
@@ -344,7 +344,8 @@ class HtmlWebpackPlugin {
       js: Array<string>,
       css: Array<string>,
       manifest?: string,
-      favicon?: string
+      favicon?: string,
+      entries: { [entry: string]: Array<string> }
     }} assets
    * @param {{
        headTags: HtmlTagObject[],
@@ -386,7 +387,8 @@ class HtmlWebpackPlugin {
       js: Array<string>,
       css: Array<string>,
       manifest?: string,
-      favicon?: string
+      favicon?: string,
+      entries: { [entry: string]: Array<string> }
     }} assets
    * @param {{
        headTags: HtmlTagObject[],
@@ -514,7 +516,8 @@ class HtmlWebpackPlugin {
       js: Array<string>,
       css: Array<string>,
       manifest?: string,
-      favicon?: string
+      favicon?: string,
+      entries: { [entry: string]: Array<string> }
     }} assets
    */
   isHotUpdateCompilation (assets) {
@@ -532,7 +535,8 @@ class HtmlWebpackPlugin {
       js: Array<string>,
       css: Array<string>,
       manifest?: string,
-      favicon?: string
+      favicon?: string,
+      entries: { [entry: string]: Array<string> }
     }}
    */
   htmlWebpackPluginAssets (compilation, childCompilationOutputName, entryNames, customPublicPath) {
@@ -574,7 +578,8 @@ class HtmlWebpackPlugin {
         js: Array<string>,
         css: Array<string>,
         manifest?: string,
-        favicon?: string
+        favicon?: string,
+        entries: { [entry: string]: Array<string> }
       }}
      */
     const assets = {
@@ -587,7 +592,9 @@ class HtmlWebpackPlugin {
       // Will contain the html5 appcache manifest files if it exists
       manifest: Object.keys(compilation.assets).find(assetFile => path.extname(assetFile) === '.appcache'),
       // Favicon
-      favicon: undefined
+      favicon: undefined,
+      // Will contain all entries along with their corresponding files
+      entries: {},
     };
 
     // Append a hash for cache busting
@@ -633,6 +640,10 @@ class HtmlWebpackPlugin {
         if (!extMatch) {
           return;
         }
+        if (!assets.entries[entryName]) {
+            assets.entries[entryName] = [];
+        }
+        assets.entries[entryName].push(entryPointPublicPath);
         // Skip if this file is already known
         // (e.g. because of common chunk optimizations)
         if (entryPointPublicPathMap[entryPointPublicPath]) {
@@ -1056,7 +1067,8 @@ class HtmlWebpackPlugin {
    js: Array<string>,
    css: Array<string>,
    manifest?: string,
-   favicon?: string
+   favicon?: string,
+   entries: { [entry: string]: Array<string> }
  }} assets
  * @param {{
      headTags: HtmlTagObject[],
--- a/typings.d.ts
+++ b/typings.d.ts
@@ -140,6 +140,7 @@ declare namespace HtmlWebpackPlugin {
             css: Array<string>;
             manifest?: string;
             favicon?: string;
+            entries: { [entry: string]: Array<string> };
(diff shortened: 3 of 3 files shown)


jantimon:
Okay I believe we should definitely try to get this into 5.x.

And I really like that your pr is quite simple and straight forward! 👍 

Plugin authors could still alter entries in the `beforeAssetTagGeneration` hook: https://github.com/jantimon/html-webpack-plugin/blob/c9ff9faa06e70609877f4e64626a59d0fc882dae/lib/hooks.js#L16-L27

However I also see one disadvantage if we go with your approach.

The current implementation works like this:

1. get the entry file result e.g. `bundle.js`
2. it calculates the public path e.g. `/my/page/bundle.js`
3. it converts it into a tag object e.g. `{ tagName: 'script', src: '/my/page/bundle.js', defer: true }`
4. it allows plugins to modify the object e.g. adding a [subresource integrity](https://github.com/waysact/webpack-subresource-integrity/) e.g.:  `{ tagName: 'script', src: '/my/page/bundle.js', defer: true, integrity:"sha384-dOTZf16X8p34q2/kYyEFm0jh89uTjikhnzjeLeF0FHsEaYKb1A1cv+Lyv4Hk8vHd",  crossorigin:"anonymous" }`
5. other plugins add additional tags (e.g. favicons or meta tags)
6. it converts it to html
6.1. with inject
6.2. with a custom template

If we would add you entries approach it would already improve the situation as users could access those values in `6.2` however they can't make use of anything which happens in step `2. - 5.`..

driskell:
@jantimon I think if someone opts to use the `entries` then in essence they're saying they don't want the tags at all - they're going to manage everything in the template. In my case those tags aren't much use because the framework will generate them, and I think the concept of dependencies doesn't really apply to any HTML generated template. So yes the 2-5 is somewhat ignored when you use `entries` in 6.2.

For the same reason above I don't think having `entries` available in the future hooks is useful either - and I sense that in doing so it would introduce further change with no justifiable use-case. Could be added in future though if someone reports a use-case.

It's why I wanted to raise an issue initially - in using this `entries` it does in some way state, "I don't want html-webpack-plugin to generate tags/html for me - I just want it to manage the gory aspects of webpack for me and give me a simple dependency tree so I can do the rest myself" which I know is exactly as you say - throwing 2-5 away - but I think that's fine and computing those 2-5 I'm happy to do since this entire plugin and its community saves a significant amount of time and maintenance work maintaining a custom plugin - which I have done, and it's not easy! 😄

driskell:
OK having thought things through a bit more - I do see there is an interesting element. Plugins are adding additional information to each bundle (in the tag area) to give things like sub integrity check. I can see now the HUGE benefit of making a link from the entries files lists to the tags that are generated, so framework could be even given those attributes like sub-integrity checks.

I will ponder on this a bit more as that is an interesting concept. It may be worth considering though it could be the next stage of `entries` to integrate that nicely and make it seamless with plugins. I'm a fan of an incremental approach!

jantimon:
Well there are also other quite promising use cases:

- preloading with `<link rel="preload" href="main.js" as="script">`
- service worker integration
- using `nomodule` for IE11 polyfilling
- WebApp manifest generation

What about the following flow:

1. add two new objects (not sure if we need both - probably `entryPublicPaths` is already enough):
```js
assets.entryFiles[entryName] = ['main.js'];
assets.entryPublicPaths[entryName] = ['../calculated/publicPath/main.js'];
```

2. send `entryFiles` and `entryPublicPaths` to the `beforeAssetTagGeneration` hook to allow modifications

3. in template parameters we add a getter for the parameters e.g.
```js

Object.defineProperty(templateParamaters.htmlWebpackPlugin.files, 'entryTags', {
  get() { 
    return {
       // just an example should be dynamic for all entry names:
       'main': {
          headTags: [/* ... */],
          bodyTags: [/* ... */],
       }
    }
  }
});

templateParamaters.htmlWebpackPlugin.files.entryFiles = /* same as in 1. */
templateParamaters.htmlWebpackPlugin.files.entryPublicPaths = /* same as in 1. */
```



So a user could then do sth like:

```js
new HtmlWebpackPlugin({
   inject: false,
   templateContent: (templateParamaters) => `
      <html>
      <head>
      ${templateParamaters.htmlWebpackPlugin.files.entryTags.main.headTags}
      </head>
      <body>
      ${templateParamaters.htmlWebpackPlugin.files.entryTags.main.bodyTags}
      </body>
      </html>
   `
})
```

or to generate json:

```js
new HtmlWebpackPlugin({
   inject: false,
   filename: 'tags.json',
   templateContent: (templateParamaters) => JSON.stringify({
      entryTags: templateParamaters.htmlWebpackPlugin.files.entryTags,
      entryFiles: templateParamaters.htmlWebpackPlugin.files.entryFiles
   })
})
```

or to generate yaml:

```js
new HtmlWebpackPlugin({
   inject: false,
   filename: 'tags.yaml',
   templateContent: (templateParamaters) => require('yaml').stringify({
      entryTags: templateParamaters.htmlWebpackPlugin.files.entryFiles,
      entryFiles: templateParamaters.htmlWebpackPlugin.files.entryFiles
   })
})
```

sigerello:
Hi @driskell, is there any progress on this? Thanks.

driskell:
@sigerello No progress sorry - I applied this PR across my projects via a patch in our build process and it's suited our purposes 100% since I did that. I haven't had any time or bandwidth to extend the scope - most specifically as testing the PR here was fairly easy as I had company time to test against our pipeline - the proposed extensions to the change here would require brand new code and testing to make sure it works.

IMHO I would say this PR could be merged as-is a phase 1. Then a phase 2 it can be made to be passed to the hooks as noted. That way we're not blocking 1 use case because of the requirements of another that hasn't been requested yet.

Of course, someone will need to fix the conflict. I might end up doing that at some point if we do a maintenance run on our node package versions as it'll need our patch updating but I can't say when that will be.

PR #1594, opened 2021-02-04
Update scriptLoading's default on README.md

kenrick95:
As per listed in CHANGELOG and in https://github.com/jantimon/html-webpack-plugin/blob/74fae99c1294c01b71c8ea082fe1557cb5375322/index.js#L54, the default value for `scriptLoading` has changed to `'defer'`


jantimon:
Cool thanks :)

Issue #1597, opened 2021-02-11
v5 not omitting null/undefined tag attribute 


themgoncalves:
## Current behaviour 💣

The issue is related to the output of html tag's attribute with null value, which was being suppressed in the previous versions. 

This behaviour is particularly not wanted, and at some extension, harmful as in my case.

My application is set to use and validate `Sub Resource Integrity` (aka `integrity` attribute), which by definition should always contain a value, otherwise, it must be suppressed from the output.

```html
<script defer="defer" src="https://www.googletagmanager.com/gtag/js?id=[GA_ID]" integrity="null" crossorigin="anonymous"></script>
```
## Expected behaviour ☀️

Suppression of _null/undefined_ tags, as on v4.

```html
<script defer="defer" src="https://www.googletagmanager.com/gtag/js?id=[GA_ID]" crossorigin="anonymous"></script>
```

### Reproduction Example 👾

This is the base setup:

```js
    new SriPlugin({
      hashFuncNames: ['sha256', 'sha384', 'sha512'],
      enabled: isProd,
    }),
    new HtmlWebpackPlugin({
      template: './src/index.html',
      title: pkg.title,
      inject: true,
      minify: {
        collapseWhitespace: true,
        preserveLineBreaks: false,
      },
    }),
    new HtmlWebpackTagsPlugin({
      append: false,
      usePublicPath: false,
      scripts: [
        {
          path: `https://www.googletagmanager.com/gtag/js?id=${GOOGLE_ANALYTICS_ID || 'foo'}`,
          attributes: { async: true, integrity: false },
        },
      ],
    }),
```

## Environment 🖥

- `Node.js`: v14.11.0
- `darwin`: 20.3.0
- `webpack`: v5.21.2
- `webpack-subresource-integrity`: v1.5.2
- `html-webpack-tags-plugin`: v3.0.0

And of course, `html-webpack-plugin`: v5.

jantimon:
the reason is this line: https://github.com/jantimon/html-webpack-plugin/blob/33d69f49e6e9787796402715d1b9cd59f80b628f/lib/html-tags.js#L33

themgoncalves:
Terrific! Thank you for the quick reply @jantimon.

I will work on the change and create the PR for it 💪

jantimon:
released as html-webpack-plugin@5.1.0 🎉

thank you so much for your help!

PR #1598, opened 2021-02-12
feat: omit html tag attribute with null/undefined/false value

themgoncalves:
### Sumary

This `pull request` update the _html tag parsing_ to also omit attributes that contains `null` and `undefined` value.

The motivation behind this change you can find on the [Issue #1597](https://github.com/jantimon/html-webpack-plugin/issues/1597).

> PS: not sure if this is a `feat` or ` fix` 🤷‍♂️, it depends on how you see it.

### DIFF
--- a/lib/html-tags.js
+++ b/lib/html-tags.js
@@ -25,12 +25,14 @@ const voidTags = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input', 'k
  *  A tag element according to the htmlWebpackPlugin object notation
  *
  * @param xhtml {boolean}
- *   Wether the generated html should add closing slashes to be xhtml compliant
+ *   Whether the generated html should add closing slashes to be xhtml compliant
  */
 function htmlTagObjectToString (tagDefinition, xhtml) {
   const attributes = Object.keys(tagDefinition.attributes || {})
     .filter(function (attributeName) {
-      return tagDefinition.attributes[attributeName] !== false;
+      return tagDefinition.attributes[attributeName] !== false &&
+        tagDefinition.attributes[attributeName] !== null &&
+        tagDefinition.attributes[attributeName] !== undefined;
     })
     .map(function (attributeName) {
       if (tagDefinition.attributes[attributeName] === true) {
@@ -49,13 +51,13 @@ function htmlTagObjectToString (tagDefinition, xhtml) {
  * @param {string} tagName
  * the name of the tag e.g. 'div'
  *
- * @param {{[attributeName: string]: string|boolean}} [attributes]
+ * @param {Object.<string, string | boolean>} [attributes]
  * tag attributes e.g. `{ 'class': 'example', disabled: true }`
  *
  * @param {string} [innerHTML]
  *
- * @param {{[attributeName: string]: string|boolean}} [meta]
- * meta information about the tag e.g. `{ 'pluhin': 'html-webpack-plugin' }`
+ * @param {Object.<string, string | boolean>} [meta]
+ * meta information about the tag e.g. `{ 'plugin': 'html-webpack-plugin' }`
  *
  * @returns {HtmlTagObject}
  */
--- a/lib/html-tags.js
+++ b/lib/html-tags.js
@@ -30,9 +30,7 @@ const voidTags = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input', 'k
 function htmlTagObjectToString (tagDefinition, xhtml) {
   const attributes = Object.keys(tagDefinition.attributes || {})
     .filter(function (attributeName) {
-      return tagDefinition.attributes[attributeName] !== false &&
-        tagDefinition.attributes[attributeName] !== null &&
-        tagDefinition.attributes[attributeName] !== undefined;
+      return tagDefinition.attributes[attributeName] === '' || tagDefinition.attributes[attributeName];
     })
     .map(function (attributeName) {
       if (tagDefinition.attributes[attributeName] === true) {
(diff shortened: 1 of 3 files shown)


lib/html-tags.js
```diff
@@ -25,12 +25,14 @@ const voidTags = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input', 'k
...
- *   Wether the generated html should add closing slashes to be xhtml compliant
+ *   Whether the generated html should add closing slashes to be xhtml compliant
  */
 function htmlTagObjectToString (tagDefinition, xhtml) {
   const attributes = Object.keys(tagDefinition.attributes || {})
     .filter(function (attributeName) {
-      return tagDefinition.attributes[attributeName] !== false;
+      return tagDefinition.attributes[attributeName] !== false &&
+        tagDefinition.attributes[attributeName] !== null &&
+        tagDefinition.attributes[attributeName] !== undefined;
```
themgoncalves, line 35: Ideally, this assertion could be rewritten as the following:

```js
[null, undefined, false].includes(tagDefinition.attributes[attributeName])
```

It has better readability, but due to lint constraints that was not possible.

lib/html-tags.js
```diff
@@ -49,13 +51,13 @@ function htmlTagObjectToString (tagDefinition, xhtml) {
...
  *
- * @param {{[attributeName: string]: string|boolean}} [attributes]
+ * @param {Object.<string, string | boolean>} [attributes]
  * tag attributes e.g. `{ 'class': 'example', disabled: true }`
  *
  * @param {string} [innerHTML]
  *
- * @param {{[attributeName: string]: string|boolean}} [meta]
- * meta information about the tag e.g. `{ 'pluhin': 'html-webpack-plugin' }`
+ * @param {Object.<string, string | boolean>} [meta]
```
themgoncalves, line 59: Also fixed the JSDoc definitions for `object` entry

jantimon: Can we please keep`attributeName` and add `undefined` & `null`?

themgoncalves: Of course, do you want to restore both changes on the JSDoc?

jantimon: Yes and please also for `typings.d.ts` :)

lib/html-tags.js
jantimon: What about `tagDefinition.attributes[attributeName] === '' || tagDefinition.attributes[attributeName]`?

lib/html-tags.js
themgoncalves: The change on `typingins.d.ts` are necessary to reflect the change

spec/basic.spec.js
```diff
@@ -1198,38 +1198,42 @@ describe('HtmlWebpackPlugin', () => {
     null, done, false, false);
   });
 
-  it('allows events to remove an attribute by setting it to false', done => {
```
jantimon, line 1201: please keep the old test and just add 2 more - 1 for null 1 for undefined..

that way we get better error messages if 1 test fails and to guarantee that the old tests are not modified and still work

lib/html-tags.js
jantimon: You are right :) sorry I didn't see `typings.d.ts` - here it should be the same as in `typings.d.ts`: 

```suggestion
 * @param {{[attributeName: string]: string | boolean | null | undefined}} [meta]
```

lib/html-tags.js
themgoncalves: This would introduce a bug in the plugin due to the nature of `truthy` values, which excludes `false`, `0`, `""`, `null`, `undefined` and `NaN` values.

As you can see, `0` as value might be needed.

jantimon: but in your typings you didn't allow numbers - should we add them too?

themgoncalves: that's actually a good point! I'm not sure about the broad use case of it, but if we're going to proceed in this way, then we can simply refactor it to:

```js
  const attributes = Object.keys(tagDefinition.attributes || {})
    .filter(function (attributeName) {
      return tagDefinition.attributes[attributeName];
    })
```

This would satisfy my use case, although I would appreciate your feedback regarding it, as you might have a better sense of the community usage.

jantimon: this would not allow an empty string like: `<input value="">` I am not sure if that's used but removing support for it would be a breaking change...

what about adding a small util function in this file so we can do sth like:

```js
  const attributes = Object.keys(tagDefinition.attributes || {})
    .filter(function (attributeName) {
      return hasValue(tagDefinition.attributes[attributeName]);
    })
```

jantimon: e.g.:

```js
/**
 * @param {undefined | null | string | number | boolean} value
 */
function hasValue(value) {
  if (value === undefined || value === null || value === false) {
    return false;
 }
 return true;
}
```

jantimon: Or we just keep your initial proposal :)

themgoncalves: The idea to have this `hasValue` is interesting, but if the usage is restricted to this single case, have the assessment inline provides better understand about the code, so I would go for that suggestion of yours:

```js
tagDefinition.attributes[attributeName] === '' || tagDefinition.attributes[attributeName]
```

This fits nicely

jantimon: Okay cool so we have two valid options I guess:

1. The one you already committed (which also allows numbers)
2. The `tagDefinition.attributes[attributeName] === '' || tagDefinition.attributes[attributeName]` which does not allow numbers

I am fine with both ways :)

themgoncalves: I have applied the suggestion you made, as you have a better understanding about its usage.

Thank you for the nice brainstorming

PR #1601, opened 2021-02-14
docs: remove `publicPath` from config options

jeffgore00:
Adding the `publicPath` key in the config options has no effect. What does have the desired effect is supplying this key in the main webpack config under `output`.


jantimon:
@jeffgore00 here is a working example with this option: https://codesandbox.io/s/html-webpack-plugin-5x-public-path-khkxp?file=/webpack.config.js

The idea is to overwrite the original publicPath if needed e.g. if you want to use a cdn or a special relative path for a generated html

jantimon:
@jeffgore00 here is a working example with this option: https://codesandbox.io/s/html-webpack-plugin-5x-public-path-khkxp?file=/webpack.config.js

The idea is to overwrite the original publicPath if needed e.g. if you want to use a cdn or a special relative path for a generated html

PR #1605, opened 2021-02-18
fix: issues#1603

913453448:
Webpack v5.22.0` breaks HtmlWebpackLoader

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,10 +133,13 @@ class HtmlWebpackPlugin {
     // Evaluate code and cast to string
     let newSource;
     try {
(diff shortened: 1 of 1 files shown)


jantimon:
hey @913453448 

thank you so much - but you are trying to fix html-webpack-plugin 5.1 which already works..

jantimon:
Thanks for the idea - I changed it slightly and released it as `html-webpack-plugin@4.5.2` - however I strongly recommend to use webpack 5 with html-webpack-plugin 5

PR #1610, opened 2021-02-22
hotfix: document v5.0 change to default scriptLoading

david-fong:
The default value of the `scriptLoading` option was changed to "defer" in 35b6b87.


jantimon:
thanks :)

PR #1637, opened 2021-04-17
feat: injection of custom root element for react/vue/angular

robertmain:
I was previously using [this plugin](https://github.com/LKay/html-webpack-root-element-plugin#readme) to inject a custom div root element into the generated HTML page.

However, that plugin has not been maintained for several years and no longer works with Webpack 5. I know many others have wished for root element support in the issues, so the time felt right to stick a PR in to implement that.

Reviews/comments/thoughts/suggestions welcome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,7 +62,8 @@ class HtmlWebpackPlugin {
         meta: {},
         base: false,
         title: 'Webpack App',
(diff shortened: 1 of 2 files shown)


robertmain:
Bump. @jantimon

PR #1650, opened 2021-06-07
Add support for `__dirname` in JavaScript bundle

### LINKED ISSUE #1649: Support `__dirname` in JavaScript bundle
It appears that there were some changes to [improve support for SSR](https://github.com/jantimon/html-webpack-plugin/commit/73d2a660b10b9ebf8a341f0ddb173bcaaf1e513c#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R138) but this didn't include the addition of `__dirname`. 

I've hit this myself in a slightly different context when messing around with Webpack + Electron which can have full nodejs in the browser. If your JavaScript bundle contains __dirname`, you get an error in `evaluateCompilationResult` saying that `__dirname is not defined`. I'm currently working around this with `dirname(__filename)`.

timfish:
Fixes #1649

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,8 +133,8 @@ class HtmlWebpackPlugin {
       require: require,
       htmlWebpackPluginPublicPath: publicPath,
       URL: require('url').URL,
-      __filename: templateWithoutLoaders
-      // __dirname: path.dirname(templateWithoutLoaders)
+      __filename: templateWithoutLoaders,
+      __dirname: path.dirname(templateWithoutLoaders)
     });
     const vmScript = new vm.Script(source, { filename: templateWithoutLoaders });
     // Evaluate code and cast to string
--- a/index.js
+++ b/index.js
@@ -133,8 +133,8 @@ class HtmlWebpackPlugin {
       require: require,
       htmlWebpackPluginPublicPath: publicPath,
       URL: require('url').URL,
-      __filename: templateWithoutLoaders,
-      __dirname: path.dirname(templateWithoutLoaders)
+      __filename: templateWithoutLoaders
+      // __dirname: path.dirname(templateWithoutLoaders)
     });
     const vmScript = new vm.Script(source, { filename: templateWithoutLoaders });
     // Evaluate code and cast to string
--- a/index.js
+++ b/index.js
@@ -131,9 +131,11 @@ class HtmlWebpackPlugin {
       ...global,
       HTML_WEBPACK_PLUGIN: true,
       require: require,
+      __non_webpack_require__: require,
       htmlWebpackPluginPublicPath: publicPath,
       URL: require('url').URL,
-      __filename: templateWithoutLoaders
+      __filename: templateWithoutLoaders,
+      __dirname: path.dirname(templateWithoutLoaders)
     });
     const vmScript = new vm.Script(source, { filename: templateWithoutLoaders });
     // Evaluate code and cast to string
--- a/index.js
+++ b/index.js
@@ -131,7 +131,6 @@ class HtmlWebpackPlugin {
       ...global,
       HTML_WEBPACK_PLUGIN: true,
       require: require,
-      __non_webpack_require__: require,
       htmlWebpackPluginPublicPath: publicPath,
       URL: require('url').URL,
       __filename: templateWithoutLoaders,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -131,9 +131,11 @@ class HtmlWebpackPlugin {
       ...global,
       HTML_WEBPACK_PLUGIN: true,
       require: require,
+      __non_webpack_require__: require,
```
jantimon, line 134: I guess we don't need this as webpack will convert `__non_webpack_require__` to `require` - or did you find any scenario where this might be useful?

index.js
```diff
@@ -131,9 +131,11 @@ class HtmlWebpackPlugin {
...
       htmlWebpackPluginPublicPath: publicPath,
       URL: require('url').URL,
-      __filename: templateWithoutLoaders
+      __filename: templateWithoutLoaders,
+      __dirname: path.dirname(templateWithoutLoaders)
```
jantimon, line 138: probably we should also add a unit test for `__filename` and `__dirname` - what do you think?

index.js
timfish: You mentioned it in the issue so I just blindly added it😬. You're right though, webpack already creates the mapping to `require`

index.js
timfish: Will add a test!

timfish:
I've tried adding a test that fails before this change but not had any luck. Everything passes! Maybe when run via jest it magically has the right context?

I need to create a minimal reproduction to prove that it is in fact broken before this change. Maybe it'll needs to be an example rather than a unit test 🤷‍♂️

timfish:
I can't trigger this issue with basic webpack configurations but finally managed to reproduce it with this.

It looks like `__dirname` is only an issue when it's been added via `requireExtensions`:
```js
const { join } = require("path");

module.exports = {
  mode: "production",
  target: "electron-renderer",
  entry: join(__dirname, "renderer.js"),
  output: {
    path: __dirname,
    filename: "renderer_bundle.js",
  },
  plugins: [
    {
      apply(compiler) {
        compiler.hooks.compilation.tap(
          "some-require-hook-plugin",
          (compilation) => {
            compilation.mainTemplate.hooks.requireExtensions.tap(
              "some-require-hook",
              (source, chunk) => {
                return `const something = __dirname + '/';`;
              }
            );
          }
        );
      },
    },
    new HtmlWebpackPlugin(),
  ],
};

```

timfish:
I added a commit with the change reverted to demonstrate the tests function in CI but then noticed the tests aren't run on commits in PRs 🤦‍♂️

spec/basic.spec.js
```diff
@@ -2799,4 +2799,33 @@ describe('HtmlWebpackPlugin', () => {
       done();
     });
   });
+
+  it('allows __dirname and __filename in requireExtensions', done => {
+    testHtmlPlugin({
+      mode: 'production',
+      target: 'electron-renderer',
```
jantimon, line 2806: I guess the target is not relevant for `__dirname`

spec/basic.spec.js
```diff
@@ -2799,4 +2799,33 @@ describe('HtmlWebpackPlugin', () => {
...
+                compilation.mainTemplate.hooks.requireExtensions.tap(
+                  'some-require-hook',
+                  (source, chunk) => {
+                    return `const something = __dirname + __filename;`;
+                  }
+                );
+              }
+            );
+          }
+        },
```
jantimon, line 2827: Please remove the hook - we can have a simple template like this:

dirname.ejs
```
<html>
<body>
  dirname: <%= __dirname %>
</body>
</html>
```

spec/basic.spec.js
```diff
@@ -2799,4 +2799,33 @@ describe('HtmlWebpackPlugin', () => {
...
+        new HtmlWebpackPlugin()]
+    }, [/<script defer="defer" src="index_bundle.js"><\/script>[\s]*<\/head>/], null, done);
```
jantimon, line 2829: here we can check what we expect to show up in the plugin - for our test maybe:

```
 ['dirname: ' + path.resolve('./fixtures/dirname.ejs'), null, done)
```

jantimon (changes requested):
Cool 👍 can you please simplify the test?

spec/basic.spec.js
timfish: I've just tried this and the test does not fail when I comment out the addition of `__dirname` to `vm.createContext` so it doesn't actually test the code path which causes the issue.

Issue #1687, opened 2021-09-29
Upgrade to Terser 5 + html-minifier-terser 6


slorber:
**Is your feature request related to a problem? Please describe.**

Docusaurus users using mono repo had problems due to conflict between Terser 4 / Terser 5

html-minifier-terser v6 has been released recently with Terser 5 support (API is now async), so this lib could also try to upgrade, ensuring we are now all using Terser 5, having less problems in monorepos, reduce install size/time

**Describe the solution you'd like**

Upgrade libs, keep same behavior.

Make `minifyHtml` async (The caller sites are already async)

https://github.com/jantimon/html-webpack-plugin/blob/0a6568d587a82d88fd3a0617234ca98d26a1e0a6/index.js#L1044

**Describe alternatives you've considered**

NA

**Additional context**

- https://github.com/terser/html-minifier-terser/issues/49
- https://github.com/facebook/docusaurus/issues/3515
- https://github.com/facebook/docusaurus/pull/5609

jantimon:
hey 👋 

thanks for reaching out - if it's only about sync vs async we should just add this.. do you have time to create a pr?


are there any other breaking changes? e.g. dropping nodejs support or anything which might cause issues for html-webpack-plugin users?

gabrielcsapo:
I can pick this up throwing a PR up now.

PR #1688, opened 2021-09-29
[chore] upgrades html-minifier-terser@5.0.0 -> 6.0.2

gabrielcsapo:
# Summary

fixes https://github.com/jantimon/html-webpack-plugin/issues/1687.

Updates the snapshots to reflect the terser upgrade.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1041,12 +1041,12 @@ function hookIntoCompiler (compiler, options, plugin) {
    *
    * @param {string} html
    */
-  function minifyHtml (html) {
+  async function minifyHtml (html) {
     if (typeof options.minify !== 'object') {
       return html;
     }
     try {
-      return require('html-minifier-terser').minify(html, options.minify);
+      return await require('html-minifier-terser').minify(html, options.minify);
     } catch (e) {
       const isParseError = String(e.message).indexOf('Parse Error') === 0;
       if (isParseError) {
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
     "webpack-recompilation-simulator": "3.2.0"
   },
   "dependencies": {
-    "@types/html-minifier-terser": "^5.0.0",
-    "html-minifier-terser": "^5.0.1",
+    "@types/html-minifier-terser": "^6.0.0",
+    "html-minifier-terser": "^6.0.2",
     "lodash": "^4.17.21",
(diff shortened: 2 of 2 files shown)


gabrielcsapo:
@jantimon I will have a follow up PR to make these fixtures snapshots as this would have been easier to run `jest -u` instead of having to update the fixtures manually if that okay with you?

jantimon:
You can rebuild all examples with `npm run rebuild-examples`

Snapshots are way harder to read

gabrielcsapo:
I will run that to make sure I got everything! @jantimon 

Is there anything else I should do before we can merge?

index.js
```diff
@@ -1041,12 +1041,12 @@ function hookIntoCompiler (compiler, options, plugin) {
    *
    * @param {string} html
    */
-  function minifyHtml (html) {
+  async function minifyHtml (html) {
```
jantimon, line 1044: please remove async

```suggestion
  function minifyHtml (html) {
```

index.js
```diff
@@ -1041,12 +1041,12 @@ function hookIntoCompiler (compiler, options, plugin) {
...
     if (typeof options.minify !== 'object') {
       return html;
     }
     try {
-      return require('html-minifier-terser').minify(html, options.minify);
+      return await require('html-minifier-terser').minify(html, options.minify);
```
jantimon, line 1049: please remove await
```suggestion
      return require('html-minifier-terser').minify(html, options.minify);
```

jantimon (commented):
We don't need async await to return the promise.  
Can you please revert those two changes?

jantimon:
@gabrielcsapo did you already have time to look into my feedback? :)

jantimon:
thanks @gabrielcsapo

jantimon:
@gabrielcsapo your changes break the build pipelines - could you please take a look what’s wrong?

gabrielcsapo:
Yes, I’ll check today.

gabrielcsapo:
@jantimon there seems to be a race condition with checking the contents of the files on disk and starting a new test. It might make sense to have each test output to a specific directory on disk to avoid collisions. I am not seeing this same issue locally, so it will be hard to debug to be for sure. 

Another option is using snapshots as they are in memory and won't have collision issues.

PR #1692, opened 2021-10-16
[chore] fixes CI to avoid race conditions

gabrielcsapo:
# Summary

Bumps dependencies related to testing and tries to fix race condition.

Tries to fix CI issue introduced in https://github.com/jantimon/html-webpack-plugin/pull/1688


gabrielcsapo:
@jantimon we should enable testing on PRs. My tests pass locally but it would be good to understand the impact to CI.

Issue #1387, opened 2020-04-02
An extra ‘viewport’ meta with default template 'src/index.ejs'


foyzhao:
I have an html template file **src/index.ejs** and the **viewport meta** has been defined.
```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,viewport-fit=cover">
  <title>title</title>
</head>
<body>
  <div id="root"></div>
</body>
</html>
```
When I don't configure the template option (the documentation says that **src/index.ejs** will be used by default)
```
plugins: [
  ...
  new HtmlWebpackPlugin(),
],
```
Another viewport meta is generated.
```
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,viewport-fit=cover">
  <title>title</title>
  <meta name="viewport" content="width=device-width,initial-scale=1">
  <link href="/main.css?d892d63a" rel="stylesheet">
</head>
<body>
  <div id="root"></div>
  <script src="/main.js?a6a01c58"></script>
</body>
</html>
```
But if I specify **src/index.ejs** as the template
```
plugins: [
  ...
  new HtmlWebpackPlugin(
    template: 'src/index.ejs',
  ),
],
```
The result is what I expected.
```
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no,viewport-fit=cover">
  <title>title</title>
  <link href="/main.css?d892d63a" rel="stylesheet">
</head>
<body>
  <div id="root"></div>
  <script src="/main.js?a6a01c58"></script>
</body>
</html>
```

jantimon:
Oh you are right what would be the best solution for that?

foyzhao:
First of all, I think the result should be exactly the same whether you explicitly specify template as src/index.ejs or implicitly use src/index.ejs as a template.

jantimon:
Yes that sounds just right

hectorgrebbell:
Bumping as this is still an issue. Agree that this should offer the same behaviour as an explicitly provided user template.

jantimon:
[...]

PR #1693, opened 2021-10-20
fix: do not modify index.ejs with default viewport meta tag

Den-dp:
Take into account `src/index.ejs` existence while deciding to add default meta-tags.

Closes: #1387


Den-dp:
I've rebased my PR just in case

Sonny40 (approved):

Sonny40 (approved):

Sonny40 (approved):

PR #1719, opened 2022-01-22
Add missing "none" value to chunksSortMode type

xemlock:
According to README (and ChunkSorter implementation) "none" value should be allowed as a `chunksSortMode` option, but TypeScript typings doesn't allow it.


alexander-akait:
Sorry for delay, fixed in master, release will be soon

Issue #1432, opened 2020-05-05
html was not generated in watch mode, after changing one of the files


gevorg-martir:
## Expected behaviour

I would expect that after every change in any other file rather than template file (example js, css, etc). I should get regenerated html file in watch mode

## Current behaviour

Now html-webpack-plugin  generates html file only first time and usually people use clean-webpack-plugin, and the result is missing html file after any change inside any file js,css, etc

## Environment

```
Node: 12.13.0
OS: darwin 19.4.0
npm: 6.12.0
webpack@4.43.0 
html-webpack-plugin@4.3.0 
```

### Config

```js
module.exports = {
  entry: 'app.js',
  output: {
    path: 'dist',
    filename: 'index_bundle.js'
  },
  module: {
    rules: [
      ...
    ]
  }
  plugins: [
    new HtmlWebpackPlugin(),
    ...
  ]
}
```

## Relevant Links

- https://github.com/gm-portfolio/pf-front

jantimon:
The reason is probably here:

https://github.com/jantimon/html-webpack-plugin/blob/683435db8d3a7b41b69adf482c44aefa00858671/index.js#L168-L170

Can you try the following:

```
new HtmlWebpackPlugin({
  cache: false
})
```

This should regenerate the template on every change (and will slow down your webpack build)

daniele-orlando:
I can confirm that

```
new HtmlWebpackPlugin({
  cache: false
})
```

fixes the problem with the `clean-webpack-plugin`.

FedeBev:
> The reason is probably here:
> 
> https://github.com/jantimon/html-webpack-plugin/blob/683435db8d3a7b41b69adf482c44aefa00858671/index.js#L168-L170
> 
> Can you try the following:
> 
> ```
> new HtmlWebpackPlugin({
>   cache: false
> })
> ```
> 
> This should regenerate the template on every change (and will slow down your webpack build)

I can confirm too, that fixes the problem.

jantimon:
fixed in html-webpack-plugin@5.0.0-alpha.10

jantimon:
this was also released official as html-webpack-plugin@5.0.0

PR #1748, opened 2022-05-14
fix: call emitAsset when template and assets are not changed

DBvc:
close issues #1432 for version 4.x


alexander-akait:
Sorry, we don't want to support v4, please update to v5, it is deprecated, anyway thank you for your PR

PR #1761, opened 2022-08-31
Update version of html-minifier-terser dependency

bradentg:
Current version of html-minifier-terser depends on outdated version of terser vulnerable to ReDOS.
html-minifier-terser@7.0.0 depends on terser ^5.14.2, which addresses the vulnerability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   },
   "dependencies": {
     "@types/html-minifier-terser": "^6.0.0",
-    "html-minifier-terser": "^6.0.2",
+    "html-minifier-terser": "^7.0.0",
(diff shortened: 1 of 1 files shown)


nksfrank:
can we get this security vulnerability fix released?

hawkril:
Would be great if @jantimon or @mastilver  Could take a look so this vulnerability fix can be merged and released. Thank you!

imki123:
I have solved this issue. There are some cached codes in a lock file. 
https://github.com/webpack/webpack/issues/16306#issuecomment-1290527482

DharanBro (commented):
Why there is no change in lock file?

PR #1763, opened 2022-09-08
fix: add support for __dirname in JavaScript bundle

VerteDinde:
This PR adds support for `__dirname` in the JS bundle. A previous PR implementing this was closed due to inactivity - this PR picks up that work to see if we can get it over the finish line.

Supersedes #1650
Fixes https://github.com/jantimon/html-webpack-plugin/issues/1649


alexander-akait:
Sorry for delay, fixed in master, release will be soon

PR #1772, opened 2022-11-18
Add "none" to chunksSortMode

kbrooks:
Missing type definition. Documentation says "none" is ok


alexander-akait:
Sorry for delay, fixed in master with some additional comment and thank you

Issue #1750, opened 2022-05-18
No documented replacement / upgrade path for `htmlWebpackPluginAlterChunks` removed in 4.0.0


OliverJEvans:
## Current behaviour 💣

In v4.0.0. removed a hook called `htmlWebpackPluginAlterChunks` but without any reasoning or documented replacement (https://github.com/jantimon/html-webpack-plugin/commit/342867e1edb7c2a8748b0aca396f160f0b13d42e).

I've picked up an old project and I'm rying to use this hook in a plugin pulled from a (now archived) Shopify Slate theme tool. However without any documented upgrade path I've hit a dead end.

I understand that its been missing for years now, but if theres any documentaiton or thoughts on how it can replaced, I'd be greatly appreciative.

## Expected behaviour ☀️

To have a documented replacement for `htmlWebpackPluginAlterChunks`

### Reproduction Example 👾

https://github.com/Shopify/slate/blob/master/packages/slate-tools/tools/webpack/html-webpack-include-chunks.js

```class HtmlWebpackIncludeLiquidStylesPlugin {
  constructor(options) {
    this.options = options;
    this.files = [];
  }

  apply(compiler) {
    compiler.hooks.compilation.tap(
      'htmlWebpackIncludeChunksPlugin',
      this.onCompilation.bind(this),
    );
  }

  onCompilation(compilation) {
    this.compilation = compilation;

    compilation.hooks.htmlWebpackPluginAlterChunks.tap(
      'htmlWebpackIncludeChunksPlugin',
      this.onAlterChunks.bind(this),
    );

    compilation.hooks.htmlWebpackPluginBeforeHtmlGeneration.tap(
      'htmlWebpackIncludeChunksPlugin',
      this.onBeforeHtmlGeneration.bind(this),
    );
  }

  onAlterChunks(chunks) {
    this.chunks = chunks;
  }

  onBeforeHtmlGeneration(htmlPluginData) {
    const assets = htmlPluginData.assets;
    const publicPath = assets.publicPath;

    this.chunks.forEach((chunk) => {
      const name = chunk.names[0];
      const chunkFiles = []
        .concat(chunk.files)
        .map((chunkFile) => publicPath + chunkFile);

      const css = chunkFiles
        .filter((chunkFile) => /.(css|scss)\.liquid($|\?)/.test(chunkFile))
        .map((chunkFile) => chunkFile.replace(/(\.css)?\.liquid$/, '.css'));

[...]

PR #1774, opened 2022-12-09
Process named chunks instead of entry points

roryabraham:
This PR refactors the plugin so that all named chunks are sorted, not just specific entry points. So far as I can tell everything else works the same. This addresses the limitation described [here](https://github.com/jantimon/html-webpack-plugin/issues/1750#issuecomment-1344724958)

### DIFF
--- a/index.js
+++ b/index.js
@@ -227,10 +227,10 @@ function hookIntoCompiler (compiler, options, plugin) {
          * @param {(err?: Error) => void} callback
          */
         (compilationAssets, callback) => {
-          // Get all entry point names for this html file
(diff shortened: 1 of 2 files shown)


roryabraham:
I'm not sure if this is a breaking change, but I don't think so. For me, the list of chunk names is essentially a superset of the list of entry point names.

tholewebgods:
Any update on this? This appears like a useful change to me.

ylavoie:
That would prove usefull here too

roryabraham:
For the sake of transparency, I don't have any current plans to continue with this PR since it's no longer a business need for us.

PR #1789, opened 2023-04-14
fix: defer loading of pretty-error to improve startup time

### LINKED ISSUE #1788: lazy-require pretty-error to improve startup time
## Current behaviour 💣

`require("pretty-error")` is run at startup since it is required in module scope. However, somehow that is rather heavy package to load, can take 200ms.

## Expected behaviour ☀️

Only do `require("pretty-error")` when actual error happens.

### Reproduction Example 👾

Anything that includes html-webpack-plugin, start node with `--inspect-brk`, attach debugger and profile. Or put console.logs in errors.js

## Environment 🖥

```
node 16, node 18
html-webpack-plugin 5.5.0
```

Knagis:
Fixes #1788


alexander-akait (approved):

alexander-akait:
@Knagis Can you fix linting?

Knagis:
fixed

TheLarkInn (approved):
Nice improvement! Thanks for the contribution 🙇

Knagis:
the failures are caused by timeouts in tests. locally the tests passed green in around a minute..

alexander-akait:
@Knagis Looks like we need update some snapshots, can you do it ehre?

alexander-akait:
You can see all of them here - https://github.com/jantimon/html-webpack-plugin/actions/runs/4702143575/jobs/8339169287?pr=1789

Knagis:
doesn't seem to be snapshots

```
  ● HtmlWebpackPlugin Examples › inline example

    thrown: "Exceeded timeout of 30000 ms for a test.
```

Knagis:
hmm.. you're right

```
    - /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = (_node_modules_css_loader_dist_cjs_js_main_css__WEBPACK_IMPORTED_MODULE_1__/* ["default"].locals */ .Z.locals || {});
    + /* harmony default export */ const __WEBPACK_DEFAULT_EXPORT__ = (_node_modules_css_loader_dist_cjs_js_main_css__WEBPACK_IMPORTED_MODULE_1__/* .default.locals */ .Z.locals || {});
```

but locally on node 16 all snapshots match

alexander-akait:
@Knagis Yeah, it is very strange, do I need to look why it happens (maybe updating them to solve the problem on CI, because we don't have node.js v16 here and it is not critical change)?

Knagis:
switched to node12, still everything passes..

```
 PASS  spec/example.spec.js
  HtmlWebpackPlugin Examples
    √ custom-template example (667 ms)
    √ default example (142 ms)
    √ favicon example (116 ms)
    √ html-loader example (157 ms)
    √ inline example (210 ms)
    √ pug-loader example (153 ms)
    √ javascript example (115 ms)
    √ javascript-advanced example (118 ms)
    √ sort manually example (129 ms)
    √ multi-page example (107 ms)
    √ template-parameters example (67 ms)
```

Knagis:
hmm. now running `npm install --ignore-scripts --force --legacy-peer-deps webpack@5` that the CI runs (previously installed deps via `yarn` since `npm i` complained about something). Will test with that

Knagis:
So running that modifies dependency:

```
-    "webpack": "5.24.3",
+    "webpack": "^5.79.0",
```

Why would CI force newest webpack instead of using the one in package.json?

Should i perhaps commit this updated package.json together with fresh snapshots?

Knagis:
ok, so ended up making more changes:
* fixed example.spec.js to correctly return errors and not get stuck with timeout
* updated webpack in package.json to newest - to match the version currently used by CI 
* updated typescript because webpack types failed with the previous version (using 4.9, not 5.0 because ts5 bumped node requirements)
* updated example snapshots (those are not jest snapshots, so left a helpful code in example.spec.js that allows easy updating)

alexander-akait:
Thank you

PR #1802, opened 2023-06-10
fix: don't add extra meta tag if it exists

alexander-akait:
fixes https://github.com/jantimon/html-webpack-plugin/issues/1387

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,7 +76,8 @@ class HtmlWebpackPlugin {
       // Default metaOptions if no template is provided
       if (!userOptions.template && options.templateContent === false && options.meta) {
         const defaultMeta = {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -76,7 +76,8 @@ class HtmlWebpackPlugin {
       // Default metaOptions if no template is provided
       if (!userOptions.template && options.templateContent === false && options.meta) {
         const defaultMeta = {
-        // From https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag
+          // TODO remove in the next major release
```
alexander-akait, line 79: I think we should avoid to add any meta tags in the next major release, we can provide a good default `HTML` template, it should be inside template or we should fully parse HTML document and check if we already have duplications

PR #1829, opened 2023-11-08
feat: add @rspack/core as an optional peer dependency

chenjiahan:
## Background

[Rspack](https://github.com/web-infra-dev/rspack) is a fast Rust-based web bundler and it is compatible with the architecture and ecosystem of webpack.

The `html-webpack-plugin` can be used with Rspack, but users will get a peer dependency warning:

```bash
 WARN  Issues with peer dependencies found
.
 └─┬ html-webpack-plugin 5.5.3
    └── ✕ missing peer webpack@^5.20.0
```

## Description

This PR adds `@rspack/core` as an optional peer dependency, and make webpack as an optional peer dependency too.

## Related issues

- https://github.com/web-infra-dev/rspack/issues/4537
- https://github.com/webpack/webpack-dev-server/issues/4776

PR #1830, opened 2023-11-09
feat: add @rspack/core as an optional peer dependency

chenjiahan:
Same as https://github.com/jantimon/html-webpack-plugin/pull/1829


chenjiahan:
This PR is a mistake, please ignore it.

Issue #1835, opened 2023-12-12
Memory leak in watch mode due to unique child compiler IDs


helloitsjoe:
## Current behaviour 💣

During watch mode, each new child compiler creates a copy of source strings, which are never removed from memory.

I think I've narrowed this down to this commit aa64b824606896ed01daf7e6da80429785ecd7bc where HtmlWebpackPlugin started using a unique ID for each child compiler instead of the template path. [This cache in MemoryCachePlugin](https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/cache/MemoryCachePlugin.js#L23) uses the unique ID when storing an entry, so the cache entries grow with each compilation.

I don't see any mention of the change in the [changelog](https://github.com/jantimon/html-webpack-plugin/blob/main/CHANGELOG.md) or [PR description](https://github.com/jantimon/html-webpack-plugin/pull/1528) (😄)  so I'm not sure why the change was made, but `id` doesn't seem to be used anywhere else in `HtmlWebpackPlugin`. Let me know if there's an explanation I'm missing somewhere!

## Potential fix 🛠️ 

The memory leak goes away when aa64b824606896ed01daf7e6da80429785ecd7bc is reverted to use `template`, and unit tests still pass. Is it possible to revert this change, or if it's solving a problem, to implement it in a different way? I'm happy to help.

<details open>
<summary>Screenshots</summary>

**Duplicated strings in memory**

The size of the leak is small in this example, but the reproduction repo below shows an example of a much larger leak.
<img width="1680" alt="Screenshot 2023-12-11 at 10 08 47 PM" src="https://github.com/jantimon/html-webpack-plugin/assets/8823810/e4fd718a-0460-4761-bc31-c5a5b7b9bbfe">

**MemoryCachePlugin cache**

Note the `HtmlWebpackPlugin_0-15` - this is the 15th iteration i.e. 15th child compiler.
<img width="1694" alt="Screenshot 2023-12-11 at 10 12 58 PM" src="https://github.com/jantimon/html-webpack-plugin/assets/8823810/034ec261-51b5-4cde-8a29-8d01f6532d16">

</details>

## Expected behaviour ☀️

Source strings are cleaned up (or able to be garbage collected) between recompilations in watch mode

### Reproduction Example 👾

The main branch of this repo includes a custom loader that makes the leak obvious: https://github.com/helloitsjoe/webpack-cache-memory-leak

[...]

PR #1836, opened 2023-12-12
Avoid using unique ID in child compiler template path

helloitsjoe:
## What problem does this solve?

Unique child compiler IDs cause a memory leak in watch mode

Resolves https://github.com/jantimon/html-webpack-plugin/issues/1835

## Details

This change reverts https://github.com/jantimon/html-webpack-plugin/commit/aa64b824606896ed01daf7e6da80429785ecd7bc, which introduced a change where each child compiler is given a unique ID. This causes duplicate source strings to be cached on every recompile, without removing the old sources.

More details and screenshots in the issue linked above. The main thing I'm unclear about is why unique IDs were used in the first place, but as far as I can tell the ID isn't referenced anywhere else. Let me know if I've missed something!

## Testing done

- [x] Unit tests pass
- [x] Tested locally, the memory leak is resolved when using `template` instead of the unique ID

### DIFF
--- a/lib/child-compiler.js
+++ b/lib/child-compiler.js
@@ -13,7 +13,6 @@
 /** @typedef {import("webpack").sources.Source} Source */
 /** @typedef {{hash: string, entry: Chunk, content: string, assets: {[name: string]: { source: Source, info: import("webpack").AssetInfo }}}} ChildCompilationTemplateResult */
 
-let instanceId = 0;
 /**
  * The HtmlWebpackChildCompiler is a helper to allow reusing one childCompiler
  * for multiple HtmlWebpackPlugin instances to improve the compilation performance.
@@ -24,8 +23,6 @@ class HtmlWebpackChildCompiler {
    * @param {string[]} templates
    */
   constructor (templates) {
-    /** Id for this ChildCompiler */
-    this.id = instanceId++;
     /**
      * @type {string[]} templateIds
      * The template array will allow us to keep track which input generated which output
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

alexander-akait:
Thank you

creage:
@alexander-akait @helloitsjoe this change leads to an issue - webpack-dev-server doesn't pick changes:
```
[webpack-dev-server] App updated. Recompiling...
[webpack-dev-server] Nothing changed
```
Rolling back to version html-webpack-plugin version 5.5.3 fixes the issue.

alexander-akait:
@creage Can you create reproducible test repo?

creage:
@alexander-akait here you go https://github.com/creage/html-webpack-plugin-nothing-changed

alexander-akait:
@creage Thank you, I will look soon

helloitsjoe:
@creage I cloned your reproduction repo and it looks like 5.5.4 also has the problem you describe. This change was released in 5.6.0, so it seems likely this is caused by something between 5.5.3 and 5.5.4, unless you're seeing something different.

alexander-akait:
@creage Intresting, I would say that we have fixed the bug, before you always had extra compilation and spent extra resources and that is why you have reloading, but when you watch `dll`, `dll-user` doesn't really know should be the application reloaded or not, you don't have HMR and you don't watch, and so you have:

```
[webpack-dev-server] App updated. Recompiling...
[webpack-dev-server] Nothing changed.
``` 

Because webpack doesn't know how to apply changes. How to fix it:
```
watchFiles: [path.join(__dirname, "../dist/manifest.json")],
```

And now dev server will reload the page when `manifest.json` was changed.

Also you can debug it using:

```
stats: {
  hash: true
},
```

And you will see that application wasn't changed, because nothing was changed in `dll-user`.

Feel free to feedback

creage:
@alexander-akait Well, since my app uses a dll, it becomes part of it, and changes to dll are changing my app, which should be reflected by reload. And, it used to work up to 5.5.3. I can watch for manifest, but that’s kinda hacky.

alexander-akait:
@creage The manifest is just a file and just a file dependency, changes in dll didn't change your application code (i.e. `dll-user`), as I said above you can see `hash` was not changed when you done a change in `dll`, for example just create `index.html` manually and add the `script` tags and you will see the same behaviour, `html-webpack-plugin` had a bug and changed a hash, but should not do it

Issue #1846, opened 2024-04-12
TypeError: childCompilation.errors.map is not a function


robert-irelan-tiktokusds:
## Current behaviour 💣

When I attempt to build my project, I find the following error:

```
TypeError: childCompilation.errors.map is not a function
at /cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/html-rspack-plugin@5.6.2_@rspack+core@0.6.1/node_modules/html-rspack-plugin/lib/child-compiler.js:166:56
at finalCallback (/cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/@rspack+core@0.6.1_@swc+helpers@0.5.3/node_modules/@rspack/core/dist/Compiler.js:223:17)
at /cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/@rspack+core@0.6.1_@swc+helpers@0.5.3/node_modules/@rspack/core/dist/Compiler.js:250:20
at /cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/@rspack+core@0.6.1_@swc+helpers@0.5.3/node_modules/@rspack/core/dist/Compiler.js:423:28
at Hook.eval [as callAsync] (eval at create (/cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/tapable@2.2.1/node_modules/tapable/lib/HookCodeFactory.js:33:10), :6:1)
at Hook.CALL_ASYNC_DELEGATE [as callAsync] (/cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/tapable@2.2.1/node_modules/tapable/lib/Hook.js:18:14)
at /cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/@rspack+core@0.6.1@swc+helpers@0.5.3/node_modules/@rspack/core/dist/Compiler.js:419:41
at /cloudide/workspace/fe_mono/common/temp/node_modules/.pnpm/@rspack+core@0.6.1_@swc+helpers@0.5.3/node_modules/@rspack/core/dist/Compiler.js:743:65
```

Looking at the [code](https://github.com/jantimon/html-webpack-plugin/blob/41f4a7b952e80eea4878d96638b4021f4a762750/lib/child-compiler.js#L195-L214), it appears that the code expects `childCompilation.errors` to be an array, but it is actually an iterator, which doesn't (yet) have a `map` method.

## Expected behaviour ☀️

[...]

PR #1847, opened 2024-04-12
Handle childCompilation.errors being an iterator rather than array

robert-irelan-tiktokusds:
Fix https://github.com/jantimon/html-webpack-plugin/issues/1846


alexander-akait (approved):

alexander-akait:
Can you fix test cases?

PR #1869, opened 2025-01-03
Store hooks as property of compilation

GeorgeTaveras1231:
This allows reading the hooks more reliably. 

Using a WeakMap in a closure causes issues if multiple versions of html-webpack-plugin are installed because it requires that plugins that tap into these hooks use the version of this package that is resolved by the webpack config. This may be a common issue in monorepos. Using a global symbol allows any version of `html-webpack-plugin` to resolve the hooks used in the compilation, allows plugins in monorepos to work more seamlessly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,10 +30,8 @@ const { AsyncSeriesWaterfallHook } = require("tapable");
 /** @typedef {Array<{ name: string, source: import('webpack').sources.Source, info?: import('webpack').AssetInfo }>} PreviousEmittedAssets */
 /** @typedef {{ publicPath: string, js: Array<string>, css: Array<string>, manifest?: string, favicon?: string }} AssetsInformationByGroups */
 /** @typedef {import("./typings").Hooks} HtmlWebpackPluginHooks */
-/**
- * @type {WeakMap<Compilation, HtmlWebpackPluginHooks>}}
- */
-const compilationHooksMap = new WeakMap();
+
+const compilationHooksSymbol = Symbol.for('html-webpack-plugin/compilation-hooks');
 
 class HtmlWebpackPlugin {
   // The following is the API definition for all available hooks
@@ -98,7 +96,7 @@ class HtmlWebpackPlugin {
    * @returns {HtmlWebpackPluginHooks}
    */
   static getCompilationHooks(compilation) {
-    let hooks = compilationHooksMap.get(compilation);
+    let hooks = compilation[compilationHooksSymbol];
 
     if (!hooks) {
       hooks = {
@@ -109,7 +107,9 @@ class HtmlWebpackPlugin {
         beforeEmit: new AsyncSeriesWaterfallHook(["pluginArgs"]),
         afterEmit: new AsyncSeriesWaterfallHook(["pluginArgs"]),
       };
-      compilationHooksMap.set(compilation, hooks);
+      Object.defineProperty(compilation, compilationHooksSymbol, {
+        value: hooks,
+      })
     }
 
     return hooks;


jantimon:
I don't think that the html-webpack-plugin should alter the webpack compilation object 

What about passing the html-webpack-plugin to the custom plugin?

```js
class MyHTMLPlugin {
  constructor(plugin) {
    this.plugin = plugin
  }
}
```

GeorgeTaveras1231:
@jantimon What is the concern with altering the object? I think this is a completely appropriate use of symbols and I imagine it's unlikely that Webpack will begin to freeze the compilation object (via `Object.freeze`).

I think requiring that the plugin instance be passed by reference is not ideal because it will make the plugin API much more verbose and move away from the existing API for implementing plugins for the `html-webpack-plugin`.

Another alternative that I would not oppose is to use a `WeakMap` but store the reference to the `WeakMap` globally (and obscure it via a global symbol) - making sure to only initialize the `WeakMap` once.

```ts
const weakMapSymbol = Symbol.for('html-webpack-plugin/hooks');
globalThis[weakMapSymbol] ??= new WeakMap();

// store in var for convenient access
const hooksMap = globalThis[weakMapSymbol];
```

The other thing to consider is that it's a bit unusual that the `html-webpack-plugin` exposes its own tappable hooks and uses the `compilation` object as the central integration point. This makes sense I suppose if the goal is to allow authoring plugins that tap into both the the underlying webpack compilation object as well as the html-webpack-plugin hooks. An alternative approach to designing the plugin APIs would be to have the `HTMLWebpaclPlugin` constructor accept child plugins that tap into its hooks. But supporting plugins that that act on the underlying compilation or work without the `html-webpack-plugin` (treating the `html-webpack-plugin` as a sort of "progressive enhancement") will require more thought/design - since this would be a significant API change.

Here's an example of what the API could look like using this approach:

```ts

module.exports = {
  plugins: [
    new HtmlWebpackPlugin({
        // Child plugins....
        plugins: [
           new SubResourceWebpackPlugin()
        ]
    })
  ]
}
```

Scutua (approved):

GeorgeTaveras1231:
@jantimon Any updates on this?

GeorgeTaveras1231:
@jantimon I was affected by this issue again so thought I'd revisit it.

I'll reiterate an old proposal and list another alternative, which can serve as a point of reference for others but to consider tradeoffs of not addressing this.

A solution I mentioned before is to store the `WeakMap` globally. I've done this in a few similar cases and works pretty well with a minor risk (I'll elaborate if needed but prob not worth getting in the weeds).

```ts
const hooksSymbol = Symbol.for('html-webpack-plugin/hooks');
globalThis[hooksSymbol] ??= new WeakMap();

const hooksMap = globalThis[hooksSymbol];
```

Another workaround I've applied in my plugins is to accept a `getHtmlWebpackPluginHooks` function,

```ts
const plugin = new MyCustomPlugin({
  getHtmlWebpackPluginHooks: (compilation) => {
    return LocalWebpackPlugin.getHooks(compilation)
  }
});
```

This works fine but feels like an unnecessary burden to place on plugin developers.

Let me know if you have any thoughts. If you oppose any change to this please just close the PR 🙏🏽

alexander-akait:
Can you run `npm ls html-webpack-plugin`?

GeorgeTaveras1231:
@alexander-akait I dont have access to this environment. But the issue is relatively straightforward. Using a traditional `node_modules` folder structure in a monorepo, where the root workspace has a version of 'html-webpack-plugin' that is not compatible with two child workspaces. This leads to multiple instances of the library loading in memory if the two child workspaces both import it.

Something like


```
V2
|_ V1
|_ V1
```

Hope this helps if you are seeking to understand the problem

alexander-akait:
@GeorgeTaveras1231 it means this should be resolved by using only one version, unfortunately there are some limitation with old plugins, especial v1, why you can't use the only one version?

GeorgeTaveras1231:
@alexander-akait in theory I agree with you.

You could force the usage of a single version by marking it as a `peerDependencies` in the package.json file of each workspace.

However, in practice, any package listed in `peerDependencies` also has to be installed as a `devDependencies` for testing and other things. And unfortunately, JS runtimes and bundlers that use the `node_modules` folder don't care wether s package is a dev only or production dependency when resolving import requests. 

You could place it in the root workspace as a `devDependencies` but this is also an anti-pattern - if for example, your tests are in a child workspace.

You could also create a whole new workspace just for testing - but I imagine for some projects will co-locate tests inside the workspace being tested - so this approach could be undesired in those cases.

I think this ultimately leaves us with limited options for solving this 

1. Solving this in the user space - force the use of a single version via dependency injection (analogous to my solution of accepting a function to get the hooks in your plugin)
2. (Solving it in this package) Making all instances of the package access the same hooks object (like my solution to use a global `Symbol` and/or global `WeakMap`)
3. Ignoring the problem - I imagine this affects a small subset of users. But the ones affected will be limited to either the dependency injection solution above (which burdens devs with identifying this bug and the additional API surface for consumers). OR doing something non-trivial to make sure the `node_modules` folder structure to always resembles a tree that prevents this issue (like I mentioned before - placing tests in their own workspace). OR doing something non-trivial to change how the `html-webpack-plugin` package is resolved by their target JS runtime or bundler before it is used.

After working in a monorepo for a couple years, I feel pretty confident in my assessment of this problem but let me know if you see any gaps in my thought process. Ultimately, I'm not blocked by this because I worked around the issue with the 'dependency injection' approach, and am just interested in giving back. So I'll accept whatever you feel is appropriate 😌

GeorgeTaveras1231:
I accidentally left out of the solutions available in the user space the original work around I added to the PR description which is to resolve `html-webpack-plugin` using the compilation context. This is maybe aligned with my suggestion to "do something non-trivial to make sure `html-webpack-plugin` is resolved correctly". Either way, calling attention to it again for the sake of thoroughness.

jantimon:
I can understand the pain and get that this really happens in large codebases.

However I believe there are better approaches than using global state and would prefer to keep the current isolation for stability.

Rather than working around multiple versions I would enforcing a single version.. That's possible for all mono repository managers. 

With pnpm you can use:

```json
{
  "pnpm": {
    "overrides": {
      "html-webpack-plugin": "^5.5.0"
    }
  }
}
```

Yarn has `resolutions` and npm has `overrides` which work similarly. Wouldn't that also solve your problem?

The main issue with your proposed global approach is that some plugins legitimately use html-webpack-plugin internally and apply their own instances:

```js

class CustomTemplatePlugin {
  apply(compiler) {
    // This plugin creates its own HtmlWebpackPlugin instance
    const htmlPlugin = new HtmlWebpackPlugin({
      template: 'custom-template.html',
      filename: 'special-output.html'
    });
    
    // Applies it to the compiler
    compiler.apply(htmlPlugin);
    
    // Later wants to hook into its own instance
    compiler.hooks.compilation.tap('CustomTemplatePlugin', (compilation) => {
      const hooks = HtmlWebpackPlugin.getCompilationHooks(compilation);
      hooks.beforeEmit.tap('CustomTemplatePlugin', (data) => {
        // Expects to work with its own instance's hooks
      });
    });
  }
}
```

With your proposed global approach, this plugin might accidentally get hooks from a different html-webpack-plugin version that the main webpack config is using, potentially causing runtime errors from incompatible hook signatures, data format mismatches between versions, or unexpected behavior when multiple html-webpack-plugin instances exist.

The current WeakMap approach ensures that each plugin gets exactly the hooks from the version it's designed to work with

If package manager overrides are not possible for you I believe using the html-webpack-plugin from root like you showed is also fine:

```js
const require = createRequire(path.resolve(compiler.options.context, 'webpack.config.js');
const HtmlWebpackPlugin = require('html-webpack-plugin');
```

I know that's not what you were looking for but I hope it helps a little bit to understand the intention of non global hooks

GeorgeTaveras1231:
@jantimon your comment totally makes sense and I understand the risk with globally storing the hooks.

I'll just say I think I spotted a contradiction in your recommendation. If the recommendation is to use a single version/instance of `html-webpack-plugin` - wouldn't the recommendation for plugin develops that use it internally (as you showed) also be to use a single a single version/instance of it? In which case then treating the hook store as a global singleton aligns with that guideline - if the guideline is "only a single instance of `html-webpack-plugin` should run at a time" 

I do see the risk of storing it globally following my recommendation leading to weird issues especially if the hook API changes significantly between major versions and a user loads more than one version in a single runtime. I think this can be managed by changing the global symbol every time there is a significant change to the hook API (required feature or breaking change) - yes this adds some maintenance overhead since you have to to keep some loose versioning scheme outside of the package.json but would allow multiple instances to communicate while mitigating some of the risk.

As far as the recommendation to use resolutions and overrides. You are right to suggest that will fix the issue but these workarounds should be reserved for special cases. And are easier to manage in smaller repos. In general, this is a pattern that limits the growth of monorepos because it requires all sub projects to be in sync. This limits the ability to do partial migration between major releases (for example, migrate a workspace at a time, merge/integrate, then iterate).

GeorgeTaveras1231:
I think I may have misinterpreted a but of your message and am now not sure if you meant to suggest that using a single instance is a general rule.

I think this is ok - but I suppose I'm still hoping something can be done. At least if possible, add guidelines for plugin developers that align with proper plugin implementations that do not lead to this issue. I think this applies to all plugin developers (wether they use `html-webpack-plugin` internally or externally).

In order to ensure that different plugins work well with each other and to avoid this error, they should either.

1. Set `html-webpack-plugin ` as a peerDependency - allowing the how to manage a single instance of it.
2. Or, allow passing (in some way) some method to get/access the hooks.

Without this, there is the risk that a minor version mismatch or accidentally incompatible version range request leads to two or more plugins unable to properly tap into the html hooks and build. I have a vague memory of running into this issue with some open source plugins that were unable to be used together for this reason - though I don't have the example readily available 😅.

Would these guidelines seem reasonable to you? And does changing this request to a documentation request resonate with you?

PR #1880, opened 2025-11-13
fix: localStorage exception in Node 25+

tophf:
The problem I'm fixing here is that Node 25+ throws on reading `localStorage` that occurs implicitly in `...global`:

```js
    const vmContext = vm.createContext({
      ...global,
```

My fix is to use a "meta" clone via Object.getOwnPropertyDescriptors + Object.defineProperties, which copies literally every global and is arguably what f269fc01fe619c5a98388017770040dae6263928 intended to achieve. Hopefully there are no secret/sensitive globals used by this plugin.

### DIFF
--- a/index.js
+++ b/index.js
@@ -632,62 +632,23 @@ class HtmlWebpackPlugin {
     const templateWithoutLoaders = templateFilename
       .replace(/^.+!/, "")
       .replace(/\?.+$/, "");
-    const vmContext = vm.createContext({
-      ...global,
-      HTML_WEBPACK_PLUGIN: true,
-      require: require,
-      htmlWebpackPluginPublicPath: publicPath,
-      __filename: templateWithoutLoaders,
-      __dirname: path.dirname(templateWithoutLoaders),
-      AbortController: global.AbortController,
-      AbortSignal: global.AbortSignal,
-      Blob: global.Blob,
-      Buffer: global.Buffer,
-      ByteLengthQueuingStrategy: global.ByteLengthQueuingStrategy,
-      BroadcastChannel: global.BroadcastChannel,
-      CompressionStream: global.CompressionStream,
-      CountQueuingStrategy: global.CountQueuingStrategy,
-      Crypto: global.Crypto,
-      CryptoKey: global.CryptoKey,
-      CustomEvent: global.CustomEvent,
-      DecompressionStream: global.DecompressionStream,
-      Event: global.Event,
-      EventTarget: global.EventTarget,
-      File: global.File,
-      FormData: global.FormData,
-      Headers: global.Headers,
-      MessageChannel: global.MessageChannel,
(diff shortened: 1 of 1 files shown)


slorber:
I tried this patch locally for Docusaurus, and it didn't work under Node 25.2:

```
×   Error: require is not defined
  │ 
  │   - dev.html.template.ejs:63
  │     /projects/docusaurus/packages/docusaurus/lib/webpack/templates/dev.html.template.ejs:63:1
  │ 
```

I tried adding back the "require" removed, leading to another error:

```
×   Error: __webpack_require__ is not defined
  │ 
  │ 
  │   - eval
  │ 
  │   - dev.html.template.ejs:10 data:text/javascript,__webpack_public_path__ = __webpack_base_uri__ = htmlWebpackPluginPublicPath;
  │     /projects/docusaurus/packages/docusaurus/lib/webpack/templates/dev.html.template.ejs:10:1
```

---

Need to double check, but I also think it fails on Node 24:

```
×   TypeError: Cannot set property crypto of #<Object> which has only a getter
  │ 
  │   - Object.assign
```

---

Repro?:

```bash
git clone git@github.com:facebook/docusaurus.git
cd docusaurus
yarn install
yarn start:website
```

alexander-akait:
@slorber Yeah, I think we will need to wait Node.js answer here, we need to pass globals here, technically we can filter as a workaround, but I still think it is a Node.js regression, let's wait an answer from them

tophf:
This change in node is spec-compliant, so it's unlikely they'll revert it.

I've re-worked the fix to handle one problem that turned out to be caused by `eval`, however I couldn't ensure the test completes locally because eslint kept finding dozens of [objectively present] violations in the existing code that I didn't change and some of `test:only` suites claimed unexpected values while they looked exactly like the expected one.

alexander-akait:
@tophf I still think it is a bug and should be reverted, any way I am fine with these changes too, do can fix eslint problem?

tophf:
> fix eslint problem

In this PR? But that code is completely unrelated, so it'd be weird to change it here. What if I just exclude the file from eslint checks altogether and someone will deal with those issues separately?

alexander-akait:
@tophf I mean fix prettier - https://github.com/jantimon/html-webpack-plugin/actions/runs/19344385065/job/55400556942?pr=1880

tophf:
Yeah, I saw that. Thing is, it doesn't say what's wrong. If the project is using prettier I would expect it to be applied automatically without me doing anything. Should I run prettier explicitly myself?

tophf:
Okay, I found and ran `fix:prettier` which fixed `lint:prettier` test, but I still can't commit the change due to `lint:js` which finds problems in this file that aren't mine, so I've just disabled the pre-commit hook and force-pushed the changes anyway.

alexander-akait:
@tophf Thanks for this PR

PR #1890, opened 2026-05-08
feat: validate chunk-related options for the plugin

n0th1ng-else:
The plugin accepts `excludeChunks` and `chunks` options that tell which JS modules to inject into the HTML. In fact, it accepts arbitrary strings as the values.
This makes it confusing while refactoring the bundler configuration, as old and non-existing chunk names may be specified, and I have to understand if it's something with the build, or the plugins configuration was not adjusted during the past update.

I am suggesting adding the plugin options validation, so the plugin will throw an error if the mentioned chunk does not exist in the build

### Examples:

- Webpack

image

- RSpack

image


alexander-akait (commented):
Good improvement, but I am afraid it will be a breaking change, because other can have the same, I think logging warning is a better solution with `TODO throw error in the next major release`

n0th1ng-else:
@alexander-akait super, appreciate your time and the suggestion.

Yes I can definitely make it a warning in the scope of this PR 🌀

I will also prepare the follow-up PR to turn it into an error, so once you plan the major release, it will be ready 👍🏿

n0th1ng-else:
@alexander-akait I have updated the PR. let me know if I can do something else 👍🏿

n0th1ng-else:
@alexander-akait I wonder if you had a chance to take another look at it? not burning, just checking if it is good anough

n0th1ng-else:
@alexander-akait just checking if I can do something more here 🙏🏿  thank you

PR #1893, opened 2026-06-16
feat: extend @rspack/core peer dependency range

Cherry:
This extends the version range for `@rspack/core` added in https://github.com/jantimon/html-webpack-plugin/pull/1829 to support the new 2.x version.


alexander-akait (approved):

karpiuMG:
when will it be released?

karpiuMG:
@jantimon @mastilver @alexander-akait  when will it be released?

PR #1903, opened 2026-09-04
[LOW] ci: limit workflow token permissions

OskarEichler:
## Summary

Declare read-only `contents` and `pull-requests` permissions for the CI workflow.

## Security impact

Without an explicit `permissions` block, the workflow token inherits repository or organization defaults, which may grant broader write access than CI needs. The workflow only checks out source, reads pull-request commits for commit linting, and uploads coverage with a separate Codecov token.

## Verification

- workflow YAML parses successfully
- verified the declared permissions are `contents: read` and `pull-requests: read`
- `git diff --check` passes

This PR addresses only workflow token permissions.


thomassavery92-beep:
jantimon/html-webpack-plugin:main

